PR #13, opened 2022-09-26
Add postgres integration

sandromello:


gateway/transport/agent.go
```diff
@@ -85,39 +86,34 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
 		}
 
 		// receive data from stream
-		reqPacket, err := stream.Recv()
+		pkt, err := stream.Recv()
 		if err != nil {
+			if err == io.EOF {
+				return
```
danielbonilha, line 93: Just marking because I need to understand this.
- Why EOF does not remove the agent from the memory mapping?

gateway/transport/agent.go
```diff
@@ -85,39 +86,34 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
...
+			log.Printf("client connection not found for gateway connection id [%s]", gwID)
+			continue
 		}
+		log.Printf("received agent msg type [%s] and gateway connection id [%s]", pkt.Type, gwID)
+		s.processAgentPacket(pkt, clientStream)
 	}
 }
 
+func (s *Server) processAgentPacket(pkt *pb.Packet, clientStream pb.Transport_ConnectServer) {
+	_ = clientStream.Send(pkt)
```
danielbonilha, line 114: Just marking because I need to understand this.
- why remove the goroutine?
- why remove the case condition?

gateway/transport/agent.go
```diff
@@ -85,39 +86,34 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
...
-
-	case pb.PacketKeepAliveType:
-		return
-
-	case pb.PacketDataStreamType:
-		cliStream := getClientStream(clientId)
-		if cliStream == nil {
-			log.Println("Client stream not available, dropping msg...")
-			return
+		if pb.PacketType(pkt.Type) == pb.PacketKeepAliveType {
```
danielbonilha, line 98: whats the difference between 
`pb.PacketType(pkt.Type) == pb.PacketKeepAliveType` 
and 
`pkt.Type == pb.PacketKeepAliveType` ?

gateway/transport/client.go
```diff
@@ -94,73 +89,81 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
+		pkt.Spec[pb.SpecGatewayConnectionID] = []byte(c.Id)
 
+		// TODO: process router connect
 		agStream := getAgentStream(conn.AgentId)
 		if agStream == nil {
-			log.Printf("agent not found for client_id [%s]", c.Id)
-			s.disconnectClient(c)
-			return
+			log.Printf("agent connection not found for gateway connection id [%s]", c.Id)
+			s.disconnectClient(c) // could we send a disconnect with an error?
```
danielbonilha, line 133: Lets discuss this

gateway/transport/client.go
```diff
@@ -94,73 +89,81 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
+			// TODO: send error back to client
+			return nil
 		}
-
-		go s.processClientMsg(reqPacket, agStream)
+		log.Printf("receive client packet type [%s] and gateway connection id [%s]",
+			pkt.Type, c.Id)
+		s.processClientPacket(pkt, c.Id, conn, agStream)
```
danielbonilha, line 139: Just marking because I need to understand this.

why remove the goroutine?

gateway/transport/client.go
```diff
@@ -94,73 +89,81 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
+			}
 			log.Printf("received error from client: %v", err)
-			s.disconnectClient(c)
-			return
+			return err
 		}
-
-		if reqPacket.Spec == nil {
-			reqPacket.Spec = make(map[string][]byte)
+		if pb.PacketType(pkt.Type) == pb.PacketKeepAliveType {
```
danielbonilha, line 120: Just marking because I need to understand this.

why put the keep alive case outside the case condition of the `processClientPacket`?

gateway/transport/server.go
```diff
@@ -70,35 +70,9 @@ func (s *Server) Connect(stream pb.Transport_ConnectServer) error {
...
-		err := s.subscribeClient(stream, token)
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
-func (s *Server) startKeepAlive(stream pb.Transport_ConnectServer) {
```
danielbonilha, line 87: Just marking because I need to understand this.

No need for the keep alive ping originated on the gateway?

gateway/transport/client.go
```diff
@@ -94,73 +89,81 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
-			return
+			return nil
 		default:
 		}
 
 		// receive data from stream
-		reqPacket, err := stream.Recv()
+		pkt, err := stream.Recv()
 		if err != nil {
+			defer s.disconnectClient(c)
```
danielbonilha, line 113: My IDE marked this line with a warn (I am not sure though):
```
Possible resource leak, 'defer' is called in the 'for' loop
```

proto/const.go
```diff
@@ -0,0 +1,26 @@
+package proto
+
+import "time"
+
+const (
+	PacketKeepAliveType       PacketType = "KeepAlive"
+	PacketDataStreamType      PacketType = "DataStream"
```
danielbonilha: I think this one is not used. I created just for the initial dummy test of the data being transferred back and forth. 

If you are not using, lets remove it.

proto/go.mod
```diff
@@ -1,16 +1,18 @@
 module github.com/runopsio/hoop/proto
 
-go 1.18
+go 1.19
```
danielbonilha, line 3: We should use the same go version accross the whole project, what you think? Other modules uses 1.18.
Maybe we update them all to use 1.19...

gateway/transport/agent.go
sandromello: PacketType is just a explicit type to make things more legible, using type we could create a function receiving a explicit constant where you know where to look for its accepted types, examples:

```go
# myFunc receives a PacketType, if I check the source of this type
# I could see which types are "accepted"
func myFunc(packetType pb.PacketType) error {
  // func body
}
```

We could make things more implicit, this way is hard to figure out what are acceptable types for the input of the function:

```go
func myFunc(packetType string) error {
  // func body
}
```

proto/go.mod
```diff
@@ -1,16 +1,18 @@
...
 
 require (
+	github.com/google/uuid v1.3.0
+	github.com/lib/pq v1.10.7
```
danielbonilha, line 7: just for the sake of curiosity, is this a postgres client?

gateway/transport/agent.go
sandromello: 1. Packets need to be processed in order, for now we just block everything which is very bad, but it's ok for the stage of the stage of users that we have. Processing packets in order in a non-blocking way needs more effort to accomplish it. 
2. We're already checking the keep alive packet above (line 98), we don't want to check the client stream for keep alive packets. For now we just need to route packets to agent, no need further logic, thus I removed the switch case.

gateway/transport/client.go
sandromello: Yeah, good point!
But it's not a problem because we're returning from the function right after.

gateway/transport/client.go
sandromello: Because we just ignore keep alive requests. We don't need to check the agent stream for these kind of packets. Maybe there's a better way to organize this.

gateway/transport/client.go
sandromello: Packets need to be processed in order, for now we just block everything which is very bad, but it's ok for the stage of the stage of users that we have. Processing packets in order in a non-blocking way needs more effort to accomplish it.

gateway/transport/server.go
sandromello: Yes. I've tested and there's no need for it. Though in clojure this is necessary

gateway/transport/agent.go
sandromello: You're right, it's a bug

client/cmd/flags.go
```diff
@@ -0,0 +1,6 @@
+package cmd
+
+var (
+	proxyPort      string
+	connectionName string
```
danielbonilha, line 5: I think this is not being used... is it?

proto/const.go
sandromello: Removed, keep alive still being used

proto/go.mod
sandromello: Yes, we just use a function to deal with scram authentication (we could copy/paste the whole code to our codebase)

client/cmd/flags.go
sandromello: It's being used in postgres.go, it's like a global for setting the flags of postgres (proxyPort, connectionName). After we add mysql it will reuse the same variables

proto/go.mod
sandromello: Done!

PR #16, opened 2022-09-29
Feat/login

danielbonilha:
- Added token validation (when not DEV mode)

### DIFF
--- a/gateway/api/middleware.go
+++ b/gateway/api/middleware.go
@@ -41,7 +41,7 @@ func (api *Api) Authenticate(c *gin.Context) {
 }
 
 func parseEmail(c *gin.Context) string {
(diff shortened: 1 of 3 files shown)


gateway/api/middleware.go
```diff
@@ -23,19 +25,46 @@ func CORSMiddleware() gin.HandlerFunc {
...
-		return
+		c.AbortWithStatus(http.StatusUnauthorized)
 	}
 
 	c.Set("context", ctx)
 	c.Next()
 }
+
+func parseEmail(c *gin.Context) string {
+	if MODE == pb.DevProfile {
```
sandromello: This should be named as PROFILE right?

PR #17, opened 2022-09-30
Add terminal integration

sandromello:
- Add run script mode and tty mode (terminal session)
- Read parameters from connection (secret, command, client args)
- Allow defining filesystem types of env allowing creating files with temporary content

Example:

```json
{
  "name": "kubectl",
  "agent_id": "test-agent",
  "type": "notused",
  "command": [
    "kubectl"
  ],
  "secret": {
		"filesystem:KUBECONFIG": "bXlrdWJlY29uZmln",
                 "envvar:MYENV": "dmFs"
	}
}
```


agent/exec.go
```diff
@@ -0,0 +1,114 @@
...
+	"log"
+	"strconv"
+	"syscall"
+
+	pb "github.com/runopsio/hoop/proto"
+	pbexec "github.com/runopsio/hoop/proto/exec"
+	"github.com/runopsio/hoop/proto/runtime"
+)
+
+func (a *Agent) processExec(pkt *pb.Packet) {
```
danielbonilha, line 14: Here in this method, can we put all the logic for each CASE in separated methods? This is for cleanness only. 
For instance, we could have a method like:
```
func (a *Agent) processExec(pkt *pb.Packet) {
	switch pb.PacketType(pkt.Type) {
	case pb.PacketExecRunProcType:
             doExecRunProc()
	case pb.PacketExecWriteAgentStdinType:
             doExecWriteAgentStdin()            
	case pb.PacketExecCloseTermType:
             doExecCloseTerm()             
```

Any thoughts?

gateway/transport/client.go
```diff
@@ -157,11 +164,38 @@ func (s *Server) processClientPacket(
...
 			Spec: map[string][]byte{
-				pb.SpecGatewayConnectionID: []byte(gwID),
-				pb.SpecAgentEnvVars:        encEnvVars,
+				pb.SpecGatewayConnectionID: []byte(client.Id),
+				// TODO: refactor to use agent connection params!
+				pb.SpecAgentEnvVarsKey: encEnvVars,
 			},
 		})
 	default:
+		if startup && client.Protocol == string(pb.ProtocoTerminalType) {
```
danielbonilha: Can we add this whole block in a separated method? This way we keep a clean case condition with only one line below each CASE (the same suggestion as the agent case above)

proto/exec/server-exec.go
```diff
@@ -0,0 +1,240 @@
...
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/exec"
```
danielbonilha: Can we add an alias here? My IDE is complaining that the exec.* does not exists. I didnt try to compile, but we can avoid confusion if we differentiate the os/exec methods from the hoop.exec ones.

proto/exec/const.go
```diff
@@ -0,0 +1,10 @@
+package exec
+
+import "syscall"
+
+const SIGWINCH = syscall.Signal(28)
```
danielbonilha: My IDE do not recognize the syscall.Signal method. It seems to not exist. What could be wrong?

(I also could not find the method in the docs: https://pkg.go.dev/syscall)

PR #18, opened 2022-10-03
Add start-cmd option

sandromello:
After this PR it's possible to 

```sh
# start hoop demo
hoop start
# start hoop gateway component
hoop start gateway
# start hoop agent component
hoop start agent
```

This will remove the three binaries and having one for starting all process:`hoop start ...`

## Changes

- Modified the default port of API / static assets to 8009
- Modified the default port of gRPC to 8010
- Make postgres proxy bind to 127.0.0.1 instead of 0.0.0.0

## Pending

- Release a image in the default hoop dockerhub org (hoophq)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,12 @@
 *.out
 
 # IDE related
-*/.idea/
-
+agent/.idea/*
+client/.idea/*
+gateway/.idea/*
+plugins/.idea/*
+proto/.idea/*
+.idea/*
 # Dependency directories (remove the comment below to include it)
 # vendor/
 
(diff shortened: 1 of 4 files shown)


gateway/main.go
```diff
@@ -1,4 +1,4 @@
-package main
+package gateway
```
danielbonilha, line 1: why did u change the package name?

sandromello: Because we call the main in the client: `hoop start gateway`

sandromello: @danielbonilha

PR #21, opened 2022-10-06
Feat/plugins

danielbonilha:
Added endpoints for plugin management:
- GET /api/plugins
- GET /api/plugins/{name}
- POST /api/plugin

Fixed minor bugs related to API error responses
Renamed auth method to be more clear on what they do
Moved xtdb structs to storage files (scope purposes)

### DIFF
--- a/gateway/api/middleware.go
+++ b/gateway/api/middleware.go
@@ -46,12 +46,7 @@ func validateClaims(c *gin.Context) (string, error) {
 	if len(tokenParts) != 2 || tokenParts[0] != "Bearer" || tokenParts[1] == "" {
 		return "", invalidAuthErr
 	}
-	email, err := parseClaims(tokenParts[1])
-	if err != nil {
-		return "", err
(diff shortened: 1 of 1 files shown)


gateway/api/middleware.go
```diff
@@ -50,25 +36,53 @@ func (api *Api) Authenticate(c *gin.Context) {
...
-		return "", fmt.Errorf("failed parsing token from header")
+		return "", invalidAuthErr
 	}
-	tokenValue := tokenParts[1]
+	email, err := parseClaims(tokenParts[1])
+	if err != nil {
+		return "", err
+	}
+
+	return email, nil
```
sandromello: TIP: No need for all this because you're not returning nothing specific, just:
```
return parseClaims(tokenParts[1])
```

It will have the same affect

danielbonilha: Thats right. Tks.

PR #28, opened 2022-10-18
Feat/add audit

sandromello:
- Add Write Ahead Log for commiting audit logs safely
- Add plugin hooks on gateway (connect, receive, disconnect, etc)
- Add generic plugin execution
- Add postgres query auditing
- Add session endpoint `/api/plugins/audit/sessions (GET,list)`
- Add session endpoint `/api/plugins/audit/sessions/:sesison_id (GET)`
- Allow filtering sessions by start_date, end_date, type, connection, user_id, limit, offset

### DIFF
--- a/gateway/api/server.go
+++ b/gateway/api/server.go
@@ -75,8 +75,8 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 	route.GET("/plugins", api.Authenticate, api.PluginHandler.FindAll)
 	route.GET("/plugins/:name", api.Authenticate, api.PluginHandler.FindOne)
 
-	route.GET("/sessions/:session_id", api.Authenticate, api.SessionHandler.FindOne)
-	route.GET("/sessions", api.Authenticate, api.SessionHandler.FindAll)
+	route.GET("/plugins/audit/sessions/:name", api.Authenticate, api.PluginHandler.FindOne)
+	route.GET("/plugins/audit/sessions", api.Authenticate, api.SessionHandler.FindAll)
 }
 
 func (api *Api) CreateTrialEntities() error {


gateway/transport/agent.go
```diff
@@ -68,6 +69,10 @@ func (s *Server) subscribeAgent(stream pb.Transport_ConnectServer, token string)
 	bindAgent(ag.Id, stream)
 
 	log.Printf("successful connection hostname: [%s], machineId [%s], kernelVersion [%s]", hostname, machineId, kernelVersion)
+	if err := s.pluginOnDisconnectPhase(pluginscore.ParamsData{}); err != nil {
+		log.Printf("ua=agent - failed processing plugin on-disconnect phase, err=%v", err)
+	}
+	s.disconnectAgent(ag)
```
danielbonilha, line 76: Looks like this is a bug. I tried to run the agent locally, and the code runs fine until this point, where agent is disconnected for no reason.

gateway/session/storage.go
```diff
@@ -0,0 +1,230 @@
...
+	"strings"
+	"time"
+
+	pluginscore "github.com/runopsio/hoop/common/plugins/core"
+	st "github.com/runopsio/hoop/gateway/storage"
+	"github.com/runopsio/hoop/gateway/user"
+	"olympos.io/encoding/edn"
+)
+
+const (
```
danielbonilha: Do you think this consts and vars definitions belong to the service layer? The service, in all domains, contains the busines logic (the filtering thing is kind of a business logic...)

sandromello: Yeah, you're right, however I'm using this both in storage and in the api, I compose the arguments in xtdb based in the available options. I could move them all options stuff to api.go, but it still going to be used in the storage.go file.

sandromello: Done @danielbonilha

gateway/transport/agent.go
```diff
@@ -77,7 +82,6 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
 	for {
 		select {
 		case <-ctx.Done():
-			s.disconnectAgent(ag)
```
danielbonilha, line 80: we can't remove this and all other disconnectAgent. This is placed within a block where we detect that the agent closed the connection.

gateway/transport/agent.go
```diff
@@ -86,7 +90,6 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
 		pkt, err := stream.Recv()
 		if err != nil {
 			if err == io.EOF {
-				s.disconnectAgent(ag)
```
danielbonilha, line 89: we can't remove this and all other disconnectAgent. This is placed within a block where we detect that the agent closed the connection.

gateway/transport/agent.go
```diff
@@ -95,20 +98,28 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
 				return nil
 			}
 			log.Printf("received error from agent, err=%v", err)
-			s.disconnectAgent(ag)
```
danielbonilha, line 98: we can't remove this and all other disconnectAgent. This is placed within a block where we detect that the agent closed the connection.

gateway/transport/agent.go
```diff
@@ -95,20 +98,28 @@ func (s *Server) listenAgentMessages(ag *agent.Agent, stream pb.Transport_Connec
...
+		}
+		clientStream := getClientStream(sessionID)
 		if clientStream == nil {
 			// TODO: warn!
-			log.Printf("client connection not found for gateway connection id [%s]", gwID)
+			log.Printf("client connection not found for session id [%s]", sessionID)
 			continue
 		}
-		log.Printf("received agent msg type [%s] and gateway connection id [%s]", pkt.Type, gwID)
+		// DEBUG
```
danielbonilha: remove this

gateway/transport/agent.go
sandromello: Yeah, thx for noticing.
Fixed it

gateway/transport/agent.go
sandromello: We "disconnect it" when this function returns

gateway/transport/agent.go
sandromello: We "disconnect it" when this function returns

gateway/transport/agent.go
sandromello: done!

gateway/transport/agent.go
sandromello: We "disconnect it" when this function returns

common/plugins/core/audit/audit.go
```diff
@@ -0,0 +1,169 @@
+package audit
```
danielbonilha: I am wondering if this whole plugins/core package should belong to the gateway module. The gateway is the only who uses it, right?

common/plugins/core/core.go
```diff
@@ -0,0 +1,68 @@
...
+	PacketData interface {
+		Packet() []byte
+		Type() string
+	}
+	Plugin interface {
+		Name() string
+		OnStartup(PluginConfig) error
+		OnConnect(ParamsData) error
+		OnDisconnect(ParamsData) error
+		OnReceive(sessionID, pktype string, payload []byte) error
```
danielbonilha: Could we reuse our structs, and pass the PacketData to OnReceive instead?
```
OnReceive(sessionID, pcktData PacketData)
```

gateway/storage/xtdb.go
```diff
@@ -33,6 +43,58 @@ func (s *Storage) Connect() error {
...
+			return "", err
+		}
+		trxVector = append(trxVector, fmt.Sprintf(`[:xtdb.api/put %v]`, string(txEdn)))
+	}
+	return fmt.Sprintf(`{:tx-ops [%v]}`, strings.Join(trxVector, "")), nil
+}
+
+// SubmitPutTx sends put transactions to the xtdb API
+// https://docs.xtdb.com/clients/1.22.0/http/#submit-tx
+func (s *Storage) SubmitPutTx(trxs ...TxEdnStruct) (*TxResponse, error) {
```
danielbonilha, line 61: I didn't understando why you created this new method to put things at xtdb. Any particular reason for not using the generic `PersistEntities(payloads []map[string]interface{}) (int64, error)` ?

sandromello: The `PersistEntities` was not putting the right type in the xtdb. I've needed that a date was encoded properly as a `#inst <datetime>` in Java in order to compare date in queries. With this method I was not able to do that because it encodes time.Time as string

common/plugins/core/audit/audit.go
sandromello: Yeah, you're right. I will refactor it.

common/plugins/core/core.go
sandromello: Yeah, good point

common/plugins/core/audit/audit.go
danielbonilha: This is a core feature, very coupled to the gateway. It depends on the gateway to have a storage provider installed, and this "plugin" cannot work alone. The decoupled plugin thing is for the users eyes only (enable or not enable). 

This core functionality will be used only by the gateway, so the business logic belongs to it. This code seems out of of context in the common package....

common/plugins/core/core.go
sandromello: Done it

common/plugins/core/audit/audit.go
sandromello: Yeah, you're right! Done it

gateway/storage/xtdb.go
danielbonilha: Maybe we should change the behaviour of persistEntities then, to handle all dates the same way (using #inst). 

I think there is no date yet being used anywhere, so it is safe to adjust it, and also affects all other entities of the project.

It seems confusing to me to have two method that does the same thing, with a little difference on the date format. Maybe we could get rid of this and try to use a single generic persistence mechanics through persistEntities? It is just a matter of adding new logic to `func EntityToMap(obj interface{}) map[string]interface{}`

gateway/plugins/core/audit/audit.go
```diff
@@ -0,0 +1,169 @@
...
+
+const (
+	Name             string = "audit"
+	defaultAuditPath string = "/opt/hoop/auditdb"
+	defaultFlushSec  int    = 30
+)
+
+var (
+	pluginAuditPath          string
+	pluginAuditFlushDuration time.Duration
```
danielbonilha: I think this is not being used for now, we could remove it

gateway/plugins/core/audit/audit.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+func init() {
```
danielbonilha: is this being called? I could not find anywhere

sandromello: This is a init function
https://golangdocs.com/init-function-in-golang

gateway/plugins/core/audit/audit.go
sandromello: Done!

gateway/plugins/core/audit/audit.go
danielbonilha: Nice. I didnt know about it. Tks.

gateway/api/server.go
```diff
@@ -71,6 +74,9 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 	route.PUT("/plugins/:name", api.Authenticate, api.PluginHandler.Put)
 	route.GET("/plugins", api.Authenticate, api.PluginHandler.FindAll)
 	route.GET("/plugins/:name", api.Authenticate, api.PluginHandler.FindOne)
+
+	route.GET("/sessions/:session_id", api.Authenticate, api.SessionHandler.FindOne)
```
danielbonilha: According to the docs we wrote in the past, the correct URL would 
- /api/plugins/audit/sessions
- /api/plugins/audit/sessions/{sessionId}

sandromello: Done

PR #29, opened 2022-10-19
fixed PROFILE handling at gateway domains

danielbonilha:
- [bugfix] changed from POST /api/login?email= to GET /api/login?email=
- [enhancement] to run production mode (with authentication) it is required to pass a PROFILE variable (this means that the app defaults to "dev" mode if no PROFILE set)
- [bugfix] defaultProviderDomain fixed URL typo
- [bugfix] edn injection fix
- [bugfix] hide the "provider" field from connections endpoint
- [bugfix] remove the "secret/" prefix from all secret items in the GET /api/connections/{name}

### DIFF
--- a/gateway/security/idp/authenticator.go
+++ b/gateway/security/idp/authenticator.go
@@ -19,6 +19,7 @@ import (
 const (
 	defaultProviderDomain   = "hoophq.us.auth0.com"
 	defaultProviderClientID = "DatIOCxntNv8AZrQLVnLb3tr1Y3oVwGW"
+	defaultProviderAudience = "https://" + defaultProviderDomain + "/api/v2/"
 )
 
 type Provider struct {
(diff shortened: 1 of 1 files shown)


gateway/connection/storage.go
```diff
@@ -77,8 +79,10 @@ func (s *Storage) FindAll(context *user.Context) ([]BaseConnection, error) {
 func (s *Storage) FindOne(context *user.Context, name string) (*Connection, error) {
 	var payload = `{:query {
 		:find [(pull ?connection [*])] 
-		:where [[?connection :connection/name "` + name + `"]
-                [?connection :connection/org "` + context.Org.Id + `"]]}}`
+		:in [name org]
+		:where [[?connection :connection/name name]
+                [?connection :connection/org org]]}
+		:in-args ["` + name + `" "` + context.Org.Id + `"]}`
```
sandromello, line 85: TIP: This becomes harder to read, using fmt.Sprintf here could be better. If you need to maintain the double quotes, use `%q`. Example:

```
fmt.Sprintf(`{:query {
		:find [(pull ?connection [*])] 
		:in [name org]
		:where [[?connection :connection/name name]
                [?connection :connection/org org]]}
		:in-args [%q %q]}`, name, context.Org.Id)

danielbonilha: Tks. Will do this next PR. Does that work for raw bytes?

PR #31, opened 2022-10-20
CLI login

danielbonilha:
- added "hoop login" command
- accept -u {email} flag, but if not passed, will listen to stdin to capture typed email
- return a login_url, and tries to open the browser
- if browser fails to open, than the link + instructions will be printed
- the callback, if success, will persist the token at $HOME/.hoop/config. If error, will print to user.
- When sending a grpc connection, will try to get the token from the $HOME/.hoop/config. If file or token not found, will run in DEV mode, with a colorful msg to user

### DIFF
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"os"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/briandowns/spinner"
@@ -59,13 +58,8 @@ func runConnect(args []string) {
 		if err := doLogin(nil); err != nil {
 			panic(err)
 		}
+		config = loadConfig()
 	}
-
-	config = loadConfig()
-
-	serverAddr := config.ServerAddress
-	serverAddr = strings.Replace(serverAddr, "https://", "", -1)
-	serverAddr = strings.Replace(serverAddr, "http://", "", -1)
 
 	loader := spinner.New(spinner.CharSets[78], 70*time.Millisecond)
 	loader.Color("green")
@@ -80,7 +74,7 @@ func runConnect(args []string) {
 	}
 
 	client, err := grpc.Connect(
-		serverAddr+":8010",
+		config.Host+":"+config.Port,
 		config.Token,
 		grpc.WithOption(grpc.OptionConnectionName, args[0]),
 		grpc.WithOption("origin", pb.ConnectionOriginClient))
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -21,6 +21,10 @@ type (
 	}
 )
 
+const (
+	httpsProtocol = "https://"
+)
+
 var loginCmd = &cobra.Command{
 	Use:   "login",
 	Short: "Authenticate at Hoop",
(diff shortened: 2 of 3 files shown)


client/cmd/connect.go
```diff
@@ -237,3 +240,32 @@ func (c *connect) printErrorAndExit(format string, v ...any) {
...
 	os.Exit(1)
 }
+
+func parseToken() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	f, err := os.Open(fmt.Sprintf("%s/.hoop/config", home))
```
sandromello: TIP: with `os.ReadFile()` it's more easy to do this

client/cmd/login.go
```diff
@@ -0,0 +1,192 @@
...
+	if resp.StatusCode == http.StatusOK {
+		b, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return "", err
+		}
+
+		var l login
+		if err := json.Unmarshal(b, &l); err != nil {
+			return "", nil
+		}
```
sandromello, line 148: TIP: You don't need to create a type to unmarshal things, just use map[string]string for that in such cases.

client/cmd/login.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if err != "" {
+		browserMsg = fmt.Sprintf("Login failed: %s", err)
+		userMsg = fmt.Sprintf("Login failed: %s\n", err)
+	}
+
+	if token != "" {
+		browserMsg = fmt.Sprintf("Login succeeded. You can close this tab now.")
+		userMsg = "Login successful\n"
+	}
```
sandromello: TIP: This is more simple to read and has less code:
```
browserMsg := fmt.Sprintf("Login succeeded. You can close this tab now.")
userMsg := "Login successful\n"
if err != ""  {
  browserMsg = fmt.Sprintf("Login failed: %s", err)
  userMsg = fmt.Sprintf("Login failed: %s\n", err)
}
(...)

File: client/cmd/login.go [hunk 4]
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	defer f.Close()
+
+	_, err = f.WriteString(token)
+	if err != nil {
+		panic(err)
+		return
```
sandromello: no need if you're using panic. Just write to stdout and exit, no need for panic

gateway/transport/server.go
```diff
@@ -58,20 +58,38 @@ func (s *Server) Connect(stream pb.Transport_ConnectServer) error {
...
-	tokenValue := t[0]
-	tokenParts := strings.Split(tokenValue, " ")
-	if len(tokenParts) != 2 || tokenParts[0] != "Bearer" || tokenParts[1] == "" {
-		return status.Errorf(codes.Unauthenticated, "invalid authentication")
+	origin := o[0]
+
+	if s.Profile == pb.DevProfile {
+		if origin == pb.ConnectionOriginAgent {
+			token = "x-agt-test-token"
+		} else {
```
sandromello: Try to avoid using else when it's not necessary, it's harder to read. Instead:
```
token = "x-hooper-test-token"
if origin == pb.ConnectionOriginAgent {
  token = "x-agt-test-token"
}

File: common/proto/const.go [hunk 6]
```diff
@@ -3,6 +3,8 @@ package proto
 import "time"
 
 const (
+	DefaultHost string = "app.hoop.dev"
```
sandromello: We shouldn't rely on defaults for our hosted solution since the customers will be running this on-premisse. They choose to run this way because of compliance and security. If we started putting defaults, the state of the app could eventually hit errors that are related to our hosted solution instead of pointing to an internal misconfiguration. 
My suggestion is that it's better to abort the program or return an error describing that a required configuration is missing.

client/cmd/login.go
```diff
@@ -0,0 +1,192 @@
...
+	apiUrl := os.Getenv("API_URL")
+	if apiUrl == "" {
+		apiUrl = pb.DefaultHost
+	}
+
+	if strings.HasPrefix(apiUrl, "localhost") {
+		apiUrl = "http://" + apiUrl
+	} else {
+		apiUrl = "https://" + apiUrl
+	}
```
sandromello: This is somehow confuse, we should expect that the user pass an `API_URL` which is an URL not a host. Why not let this comes from the env?

client/cmd/connect.go
```diff
@@ -54,6 +53,20 @@ type connect struct {
...
+
+	if config.Token == "" {
+		if err := doLogin(nil); err != nil {
+			panic(err)
+		}
+	}
+
+	config = loadConfig()
+
+	serverAddr := config.Host
```
sandromello: So, this is a server address not a host. I think it's better to change the toml to contain an attribute with `server_address` or `serveraddr` instead of host

client/cmd/connect.go
```diff
@@ -65,10 +78,12 @@ func runConnect(args []string) {
 		connectionName: args[0],
 		loader:         loader,
 	}
+
 	client, err := grpc.Connect(
-		connectFlags.serverAddress,
-		connectFlags.token,
-		grpc.WithOption(grpc.OptionConnectionName, args[0]))
+		serverAddr+":8010",
```
sandromello: The SaaS is deployed with the port 8443. The user must known which port is configured the gateway. This hardcoded port will break our setup

client/cmd/login.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+var done chan bool
+
+func doLogin(args []string) error {
+	config := loadConfig()
+
+	if config.Host == "" {
+		reader := bufio.NewReader(os.Stdin)
+		fmt.Print("No server address configured.\n")
```
sandromello: Maybe put something like: "press enter to leave the default address"

PR #35, opened 2022-10-24
Add connect type cmd

sandromello:
- Add connection params to command-line type
- Add mutex to all Send calls to gateway
- Use a incremental counter to count connections instead of uuid
- Refactor Name: change gateway id to session
- Port fixes from https://github.com/runopsio/runops-proxy/pull/26

> This PR disables one-off commands for now due to the refactor of connect and to speed things for now. Not needed for our use yet.

### DIFF
--- a/common/grpc/client.go
+++ b/common/grpc/client.go
@@ -1,0 +1,117 @@
+package grpc
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"sync"
+	"time"
+
+	pb "github.com/runopsio/hoop/common/proto"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+type (
+	OptionKey     string
+	ClientOptions struct {
+		optionKey OptionKey
+		optionVal string
+	}
+	mutexClient struct {
+		grpcClient  *grpc.ClientConn
+		stream      pb.Transport_ConnectClient
+		ctx         context.Context
+		mutex       sync.RWMutex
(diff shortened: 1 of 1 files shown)


gateway/transport/client.go
```diff
@@ -227,12 +197,19 @@ func (s *Server) addConnectionParams(startup *bool, c *client.Client, conn *conn
 			ClientArgs: clientArgs,
 		})
 		if err != nil {
-			return fmt.Errorf("failed encoding command exec params err=%v", err)
+			return fmt.Errorf("failed encoding connection params err=%v", err)
```
danielbonilha: can we put this block under a separated method just to keep the switch/case more readable?

client/grpc/client.go
```diff
@@ -12,69 +14,86 @@ import (
...
+		stream      pb.Transport_ConnectClient
+		ctx         context.Context
+		mutex       sync.RWMutex
+		closeSignal chan struct{}
 	}
 )
 
-func ConnectGrpc(connectionName string, protocol pb.ProtocolType) (*Client, error) {
-	// dial server
+func Connect(token string, kv ...string) (pb.ClientTransport, error) {
```
danielbonilha: What if we replace this generic kv by a struct with the specific fields?
```
clientMeta struct {
  hostname string
  kernelId string
 ...
}
```

danielbonilha: We could also add the token here instead of a separated param, since they all go to the metadata headers...

PR #36, opened 2022-10-26
Add connection flags to clients: token, server address

sandromello:
- Add runtime info to gRPC connection
- Removed protocol name which was not used anymore

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"fmt"
 	"log"
+	"os"
 
 	"github.com/runopsio/hoop/common/grpc"
 	"github.com/runopsio/hoop/common/version"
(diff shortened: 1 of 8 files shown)


agent/main.go
```diff
@@ -3,14 +3,19 @@ package agent
...
 )
 
 func Run() {
 	fmt.Println(string(version.JSON()))
-	client, err := grpc.Connect("x-agt-test-token")
+	token := os.Getenv("TOKEN")
+	if token == "" {
+		token = "x-agt-test-token"
+	}
+	client, err := grpc.Connect(os.Getenv("SERVER_ADDRESS"), token)
```
danielbonilha, line 18: Should we default to 127.0.0.1:8010 the same way we do for the client?

PR #38, opened 2022-10-27
Connect to gRPC server with TLS

sandromello:


common/grpc/client.go
```diff
@@ -38,27 +45,36 @@ func WithOption(optKey OptionKey, val string) *ClientOptions {
 
 func Connect(serverAddress, token string, opts ...*ClientOptions) (pb.ClientTransport, error) {
 	if serverAddress == "" {
-		serverAddress = "127.0.0.1:8010"
+		serverAddress = defaultAddress
 	}
-	conn, err := grpc.Dial(serverAddress, grpc.WithInsecure())
+	rpcCred := oauth.NewOauthAccess(&oauth2.Token{AccessToken: token})
```
danielbonilha, line 50: I think this token is not being validated on the gateway side using this approach, is it? How does this work? We can leave this here, but I am not sure about removing the authorization header from the contextOptions. Maybe this is how to token is parsed on gateway side..

PR #45, opened 2022-11-01
added agent registration url

danielbonilha:


agent/main.go
```diff
@@ -12,9 +12,18 @@ import (
...
+		log.Fatal("missing required SERVER_ADDRESS variable")
+	}
+
 	token := os.Getenv("TOKEN")
+	if token == "" {
+		log.Fatal("missing required TOKEN variable")
+	}
 
-	client, err := grpc.Connect(os.Getenv("SERVER_ADDRESS"), token, grpc.WithOption("origin", pb.ConnectionOriginAgent))
+	client, err := grpc.Connect(svrAddr, token, grpc.WithOption("origin", pb.ConnectionOriginAgent))
```
sandromello: This will break the D0 flow, since we rely on connecting with localhost

sandromello: The checks aren't necessary because the customer will add the TOKEN and SERVER_ADDRESS

danielbonilha: Well, it will fail in the grpc.Connect() anyway... will remove them...

agent/main.go
```diff
@@ -12,9 +12,11 @@ import (
 
 func Run() {
 	fmt.Println(string(version.JSON()))
-	token := os.Getenv("TOKEN")
 
-	client, err := grpc.Connect(os.Getenv("SERVER_ADDRESS"), token, grpc.WithOption("origin", pb.ConnectionOriginAgent))
+	svrAddr := os.Getenv("SERVER_ADDRESS")
```
sandromello, line 16: Forget to tell that the server address will have the port of the gRPC not the API, you'll need to parse this address to obtain the host.
Ref: https://docs.runops.io/docs/self-host/eks#4-agents

PR #47, opened 2022-11-02
added agent.toml config

danielbonilha:


agent/main.go
```diff
@@ -2,21 +2,38 @@ package agent
...
-	"github.com/runopsio/hoop/common/grpc"
-	"github.com/runopsio/hoop/common/version"
 )
 
 func Run() {
 	fmt.Println(string(version.JSON()))
 
-	svrAddr := os.Getenv("SERVER_ADDRESS")
-	token := os.Getenv("TOKEN")
+	defaultServerAddress := "127.0.0.1:8010"
```
sandromello, line 17: You don't need this, because it's already the default server on grpc.Connect func

agent/main.go
```diff
@@ -2,21 +2,38 @@ package agent
...
+		if conf.Token == "" {
+			conf.Token = "x-agt-" + uuid.NewString()
+		}
+	}
+
+	if conf.ServerAddress == "" {
+		conf.ServerAddress = os.Getenv("SERVER_ADDRESS")
+		if conf.ServerAddress == "" {
+			conf.ServerAddress = defaultServerAddress
+		}
```
sandromello, line 31: Don't need this line. This is already being validated in grpc.Connect func

agent/main.go
```diff
@@ -26,10 +43,70 @@ func Run() {
...
+	if _, err := os.Stat(filepath); os.IsNotExist(err) {
+		f, err := os.Create(filepath)
+		if err != nil {
+			log.Fatal(err)
+		}
+		defer f.Close()
+	}
+
+	return filepath
+}
```
sandromello, line 112: I think it's best for refactor this to use a single config file in the same path. The good thing of using toml is that we can have a single definition of configuration with distinct sections. Example `$HOME/.hoop/config`:

```toml
[client]
token = "jwt-token"
host = "127.0.0.1"
port = "8010"
email = "user@domain.tld"

[agent]
server_address = "127.0.0.1:8010"
token = "api-key-token"
```

The same code is already here: https://github.com/hoophq/hoop/blob/main/client/cmd/config.go

PR #50, opened 2022-11-04
user management API

danielbonilha:
Added endpoints:
- GET /users
- GET /users/:id
- PUT /users/:id

https://3.basecamp.com/5385186/buckets/28171754/todos/5506971066
https://3.basecamp.com/5385186/buckets/28171754/todos/5506972912
https://3.basecamp.com/5385186/buckets/28171754/todos/5507046756

### DIFF
--- a/gateway/user/api.go
+++ b/gateway/user/api.go
@@ -73,7 +73,7 @@ func (a *Handler) Put(c *gin.Context) {
 		return
 	}
 
-	if !inStatus(user.Status) {
+	if !isInStatus(user.Status) {
 		c.JSON(http.StatusBadRequest, gin.H{"message": "invalid status"})
 		return
 	}
@@ -81,7 +81,7 @@ func (a *Handler) Put(c *gin.Context) {
 	// for admins to not auto-exclude themselves from admin by mistake
 	if existingUser.Id == context.User.Id &&
 		existingUser.isAdmin() &&
-		!inList(GroupAdmin, user.Groups) {
+		!isInList(GroupAdmin, user.Groups) {
 		user.Groups = append(user.Groups, GroupAdmin)
 	}
 
--- a/gateway/user/service.go
+++ b/gateway/user/service.go
@@ -125,10 +125,10 @@ func isPublicDomain(domain string) bool {
 }
 
 func (user *User) isAdmin() bool {
-	return inList(GroupAdmin, user.Groups)
+	return isInList(GroupAdmin, user.Groups)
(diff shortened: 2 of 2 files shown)


gateway/user/api.go
```diff
@@ -8,5 +13,86 @@ type (
...
+		c.AbortWithStatus(http.StatusInternalServerError)
+		return
+	}
+
+	c.PureJSON(http.StatusOK, users)
+}
+
+func (a *Handler) FindOne(c *gin.Context) {
+	ctx, _ := c.Get("context")
+	context := ctx.(*Context)
```
sandromello, line 37: This line will panic.
The below will not:
```go
context, _ := ctx.(*Context)
```

However it's important to check if in the downstream it won't panic as receiving a nil context

gateway/user/service.go
```diff
@@ -96,3 +123,25 @@ func isPublicDomain(domain string) bool {
...
+}
+
+func inStatus(status StatusType) bool {
+	for _, s := range statuses {
+		if s == status {
+			return true
+		}
+	}
+	return false
+}
```
sandromello: I think `isInStatus` and `isInList` self describe that these are are bool functions just by reading it. Like you did with `isAdmin`

gateway/user/storage.go
```diff
@@ -39,17 +39,24 @@ func (s *Storage) FindById(identifier string) (*Context, error) {
...
+		return nil, err
 	}
 
-	return txId, nil
+	var users []User
+	if err := edn.Unmarshal(b, &users); err != nil {
+		return nil, err
+	}
+
+	return users, nil
```
sandromello, line 59: TIP: Since you're not checking anything, this also works:
```go
return users, edn.Unmarshal(b, &users)
```

gateway/user/api.go
danielbonilha: The context is never nil because we create and return it here https://github.com/hoophq/hoop/blob/main/gateway/user/storage.go#L15 

and then we set it to the context here https://github.com/hoophq/hoop/blob/main/gateway/api/middleware.go#L29.

If it is nil at somepoint, than there's a bug in the middleware =)

gateway/user/service.go
danielbonilha: Done.

gateway/user/storage.go
danielbonilha: Tks.

PR #51, opened 2022-11-07
Add non-obvious callback address for hoop client

sandromello:
- Change default callback port from 3333 to 3587 to prevent conflict with known apps


scripts/run-with-idp-provider.sh
```diff
@@ -0,0 +1,29 @@
...
+until curl -s -f -o /dev/null "http://127.0.0.1:3001/_xtdb/status"
+do
+  echo -n "."
+  sleep 1
+done
+echo " done!"
+echo "--> STARTING GATEWAY ..."
+
+export IDP_CLIENT_ID=hAgCLsLJeuEQLhNLpjrddd7NmSDHBIKF
+export IDP_CLIENT_SECRET=A9xw9TOXTr_YvGoA_UA4jPoqYS2GQVF7RfUptvCYwiHIKkHx11ZATRX2mQ0_1tm8
```
danielbonilha: can we avoid this secret commit?

sandromello: This is for development only, I can put some comments there

sandromello: I added a comment to this script, don't see this a security issue because it's a development application on auth0. Do you still think that we should remove it?

danielbonilha: Did you create a new app at runops auth0?

danielbonilha: It is not a good practice, anyhow. Github will warn you that there is a secret comitted.

client/cmd/login.go
```diff
@@ -12,6 +12,7 @@ import (
 	"runtime"
 	"strings"
 
+	pb "github.com/runopsio/hoop/common/proto"
```
sandroollem, line 15: Test notification

sandroollem: More

sandroollem: Test more

scripts/run-with-idp-provider.sh
sandroollem: I will remove it and add for my local development settings

PR #52, opened 2022-11-09
Add create user endpoint

mtmr0x:
> Although it needs some testing, the code itself could be reviewed already, if anyone wants. There's something wrong in the flow with Auth0 and I'm on it to finish testing

### This PR implements
- InvitedUser method for pre-creating a user;
- verification on signup if the user trying to signup was invited by someone and applying the information necessary to them;
- the endpoint `POST /users` as an HTTP interface to create users.

### DIFF
--- a/gateway/user/api.go
+++ b/gateway/user/api.go
@@ -109,7 +109,7 @@ func (a *Handler) Post(c *gin.Context) {
 		return
 	}
 
-	if len(newUser.Email) == 0 {
+	if newUser.Email == "" {
 		c.JSON(http.StatusBadRequest, gin.H{"message": "email can't be empty"})
 		return
 	}


gateway/user/api.go
```diff
@@ -96,3 +98,30 @@ func (a *Handler) Put(c *gin.Context) {
...
+	ctx, _ := c.Get("context")
+	context := ctx.(*Context)
+
+	var newUser InvitedUser
+	if err := c.ShouldBindJSON(&newUser); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"message": err.Error()})
+		return
+	}
+
+	if len(newUser.Email) == 0 {
```
danielbonilha: TIP: The email will become an empty string if not provided, and thus can be validated using a simpler notation:
`if newUser.Email == ""`

mtmr0x: nice, thanks! I knew that, but maybe I'm programming Clojure for way too much time hahaha. I will change, it's definitely simpler

PR #53, opened 2022-11-10
refac/plugin_with_configs

danielbonilha:
This PR addresses the need to manage plugins with `per connection` configurations.
- Memory map created to hold a list of registered plugins with its `config` for each `sessionID`
- Moved the global plugins registration to the main, where all concrete classes are instanciated
- Converted the `plugins.ParamsData` to a typed structure `plugin.Config`
- Created dummy review plugin

### DIFF
--- a/gateway/main.go
+++ b/gateway/main.go
@@ -2,8 +2,6 @@ package gateway
 
 import (
 	"fmt"
-	pluginsaudit "github.com/runopsio/hoop/gateway/transport/plugins/audit"
-	pluginsreview "github.com/runopsio/hoop/gateway/transport/plugins/review"
 	"os"
 
 	"github.com/runopsio/hoop/gateway/plugin"
@@ -29,10 +27,7 @@ func Run() {
 		panic(err)
 	}
 
-	transport.AllPlugins = []transport.EnabledPlugin{
-		pluginsaudit.New(),
-		pluginsreview.New(),
-	}
+	transport.LoadPlugins()
 
 	profile := os.Getenv("PROFILE")
 	idProvider := idp.NewProvider(profile)
--- a/gateway/transport/client.go
+++ b/gateway/transport/client.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"github.com/runopsio/hoop/gateway/plugin"
 	pluginsaudit "github.com/runopsio/hoop/gateway/transport/plugins/audit"
+	pluginsreview "github.com/runopsio/hoop/gateway/transport/plugins/review"
 	"github.com/runopsio/hoop/gateway/user"
 	"io"
 	"log"
@@ -30,11 +31,11 @@ type (
 	}
 
 	pluginConfig struct {
-		EnabledPlugin
+		Plugin
 		config []string
 	}
 
-	EnabledPlugin interface {
+	Plugin interface {
 		Name() string
 		OnStartup(config plugin.Config) error
 		OnConnect(p plugin.Config) error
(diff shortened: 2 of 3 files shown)


gateway/transport/plugins/audit/audit.go
```diff
@@ -35,33 +34,40 @@ func init() {
...
+	}
+
+	StorageWriter interface {
+		Write(config plugin.Config) error
+	}
+)
 
-func New() pluginscore.Plugin {
-	return &auditPlugin{walSessionStore: memory.New()}
+func New() *auditPlugin {
```
sandromello, line 51: The goal of having the interface when calling `New` is to point for the code reader which kind of methods are required to create a plugin. Imagine that tomorrow we need to add a new method for this plugin, looking at this file we can know which interface to change. What do you think, it makes sense?

gateway/transport/client.go
```diff
@@ -23,22 +25,44 @@ type (
...
+		plugins     map[string][]pluginConfig
 		mu          sync.Mutex
 	}
+
+	pluginConfig struct {
+		EnabledPlugin
+		config []string
+	}
+
+	EnabledPlugin interface {
```
sandromello: What do you think in calling this `Plugin` or `PluginProcesser`? I think it makes more sense

gateway/main.go
```diff
@@ -27,6 +29,11 @@ func Run() {
 		panic(err)
 	}
 
+	transport.AllPlugins = []transport.EnabledPlugin{
+		pluginsaudit.New(),
+		pluginsreview.New(),
+	}
```
sandromello: What do you think of moving this into `gateway/transport/client.go`? Since we're just an initializing stuff that makes sense for the transport package, it will be more clear when you see the use of the global `AllPlugins` in that file.

gateway/transport/client.go
```diff
@@ -23,22 +25,44 @@ type (
...
+		Name() string
+		OnStartup(config plugin.Config) error
+		OnConnect(p plugin.Config) error
+		OnReceive(sessionID string, config []string, packet *pb.Packet) error
+		OnDisconnect(p plugin.Config) error
+	}
 )
 
+var AllPlugins []EnabledPlugin
```
sandromello: I think this name should be call `RegisteredPlugins`, because it's what we're doing in practice.
- We register plugin when we write a plugin in the codebase and instantiate it.
- We enable a plugin by toggling the state in the database

gateway/transport/plugins/audit/audit.go
```diff
@@ -35,33 +34,40 @@ func init() {
...
-	enabled         bool
-	started         bool
-	mu              sync.RWMutex
-}
+type (
+	auditPlugin struct {
+		name            string
+		storageWriter   StorageWriter
+		walSessionStore memory.Store
+		enabled         bool
```
sandromello: Need to remove this, it's not used anymore right?

gateway/transport/plugins/audit/audit.go
danielbonilha: Generally the interface is declared where it is used, and not directly referenced where it is implemented.

The `EnabledPlugin` have a set of methods in there. In Go the interfaces are implicitly implemented, what means that the implementer (in this case the audit file) does not need to know about the interface itself, neithere reference it, it just implements its methods. 

If we change this for the interface, we are directly referencing the interface declared somewhere else.

gateway/transport/plugins/audit/audit.go
danielbonilha: Correct. I will remove.

gateway/transport/client.go
danielbonilha: They are actually all Plugins that we currently support. They are created in the main and survive all along the life of the application.

gateway/transport/client.go
danielbonilha: Yes, Plugin sounds better!!

gateway/main.go
danielbonilha: Yes, make sense. I moved into there.

PR #54, opened 2022-11-11
Redact content by default to interactive sessions

sandromello:
This feature will redact all content of interactive sessions by default when the env `GOOGLE_APPLICATION_CREDENTIALS_JSON` is set in the gateway. In case of failure, the content will be returned as is.

Any content available in the tty with more than 30 bytes in length will be redacted, anything below will not.

Discussion: https://3.basecamp.com/5385186/buckets/28171754/messages/5528611674

## Out of Scope

- Manage DLP plugin per connection
- Postgres redacting

### DIFF
--- a/agent/const.go
+++ b/agent/const.go
@@ -2,7 +2,6 @@ package agent
 
 const (
 	connectionStoreParamsKey string = "params:%s"
-	procStoreKey             string = "proc:%s"
 	cmdStoreKey              string = "cmd:%s"
 	EnvVarDBHostKey          string = "HOST"
 	EnvVarDBUserKey          string = "USER"
--- a/agent/dlp/gcp.go
+++ b/agent/dlp/gcp.go
@@ -33,7 +33,7 @@ func NewDLPClient(ctx context.Context, credentialsJSON []byte) (*Client, error) 
 func NewDLPStreamWriter(client pb.ClientTransport,
 	dlpClient *Client,
 	packetType pb.PacketType,
-	spec map[string][]byte) io.WriteCloser {
+	spec map[string][]byte) *streamWriter {
 	return &streamWriter{
 		client:     client,
 		dlpClient:  dlpClient,
@@ -50,9 +50,9 @@ func (s *streamWriter) Write(data []byte) (int, error) {
 	p.Type = s.packetType.String()
 	p.Spec = s.packetSpec
 	if s.dlpClient != nil && len(data) > 30 {
-		chunksBuffer := BreakPayloadIntoChunks(bytes.NewBuffer(data))
-		redactedChunks := RedactChunks(s.dlpClient, s.dlpClient.GetProjectID(), chunksBuffer)
-		dataBuffer, tsList, err := JoinChunks(redactedChunks)
+		chunksBuffer := breakPayloadIntoChunks(bytes.NewBuffer(data))
+		redactedChunks := redactChunks(s.dlpClient, s.dlpClient.GetProjectID(), chunksBuffer)
+		dataBuffer, tsList, err := joinChunks(redactedChunks)
 		if err != nil {
 			return 0, fmt.Errorf("failed joining chunks, err=%v", err)
 		}
@@ -135,10 +135,10 @@ func deidentifyContent(ctx context.Context, client *Client, conf *DeidentifyConf
 	return chunk
 }
 
-// RedactChunks process chunks in parallel reordering after the end of each execution.
+// redactChunks process chunks in parallel reordering after the end of each execution.
 // A default timeout is applied for each chunk. If a requests timeout or returns an error the chunk is returned
 // without redacting its content.
-func RedactChunks(client *Client, projectID string, chunksBuffer []*bytes.Buffer) []*Chunk {
+func redactChunks(client *Client, projectID string, chunksBuffer []*bytes.Buffer) []*Chunk {
 	conf := &DeidentifyConfig{
 		MaskingCharacter: defaultMaskingCharacter,
 		NumberToMask:     defaultNumberToMask,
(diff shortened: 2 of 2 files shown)


common/dlp/gcp.go
```diff
@@ -0,0 +1,160 @@
+package dlp
```
danielbonilha: The entire DLP thing is used by the gateway only (as a plugin). Any particular reason for this package to stay in the common?

agent/agent.go
```diff
@@ -163,6 +158,23 @@ func (a *Agent) processAgentConnect(pkt *pb.Packet) {
 	}
 	sessionID := pkt.Spec[pb.SpecGatewaySessionID]
 	log.Printf("session=%v - received connect request", string(sessionID))
+	if gcpRawCred, ok := pkt.Spec[pb.SpecAgentGCPRawCredentialsKey]; ok {
```
danielbonilha, line 161: DLP is plugin. Why is it not being handled as a plugin?

I mean, we should handle the same way we do for audit/review: create a plugin file at "gateway/plugins/dlp", that implements OnConnect and so on, and use that mechanic to handle the messages and redact them.

Please let me know if there is any particular reason for using this approach.

sandromello: I stated in the PR that it's out of scope for now:
> Manage DLP plugin per connection

common/dlp/gcp.go
sandromello: In this case is being used by the agent, but are just common functions that could be used by the gateway in the future. If you feel this is still a problem I can move this code to the agent.

danielbonilha: Why is it being used by the agent? The flow, as I understood is: 

1) client send a package -> package goes to gateway -> gateway sends to agent ->
agent process and send back to gateway -> gateway apply all the plugins (including DLP) -> gateway send back to client.

I do not see a usage for DLP neither at client or agent. It is a gateway thing.

agent/agent.go
danielbonilha: I am confused on this implementation haha. The DLP is all about plugins. What was implemented that is DLP but is not a plugin? The DLP should be conditioned to a connection plugin registration stating that it wants to use DLP. This means that DLP will only be applicable by connections eligible to it, and thus, must be correlated to plugin registration.

sandromello: I will do this later. This is to help to keep the PR more concise and facilitate the review instead of delivering a big PR with a lot of changes

danielbonilha: So, if the plugin thing is being implemented later, what is being implemented on this PR? I think it is not clear to me.

common/proto/types.go
```diff
@@ -68,13 +71,38 @@ func NewStreamWriter(client ClientTransport, packetType PacketType, spec map[str
 	return &streamWriter{client: client, packetType: packetType, packetSpec: spec}
 }
 
+func NewDLPStreamWriter(client ClientTransport,
+	dlpClient *dlp.Client,
+	packetType PacketType,
+	spec map[string][]byte) io.WriteCloser {
```
danielbonilha: Here we should declare the concrete struct as returning obj:
```
(client pb.ClientTransport,
	dlpClient *Client,
	packetType pb.PacketType,
	spec map[string][]byte) *streamWritter {...}
```

The struct that implements the interface does it in a implicit way, and should not reference the interface that it is implementing. This struct could implement more than one interface. We return the object created, and on the other side, it will be treated as a io.WriteCloser.

common/dlp/gcp.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, c := range chunks {
+		if _, err := res.Write(c.data.Bytes()); err != nil {
+			return nil, nil, fmt.Errorf("[dlp] failed writing chunk (%v) to buffer, err=%v", c.Index, err)
+		}
+		tsList = append(tsList, c.TransformationSummary)
+	}
+	return res, tsList, nil
+}
+
+func BreakPayloadIntoChunks(payload *bytes.Buffer) []*bytes.Buffer {
```
danielbonilha: This method belongs to DLP business logic, I think it could be an non-exported variable `breakPayloadIntoChunks` just for internal use. This method makes no sense outside the scope of its package.

common/dlp/gcp.go
```diff
@@ -0,0 +1,160 @@
...
+			fmt.Sprintf("%v", s.TransformedBytes)}
+	}
+	chunk.data = bytes.NewBufferString(r.Item.GetValue())
+	return chunk
+}
+
+// RedactChunks process chunks in parallel reordering after the end of each execution.
+// A default timeout is applied for each chunk. If a requests timeout or returns an error the chunk is returned
+// without redacting its content.
+func RedactChunks(client *Client, projectID string, chunksBuffer []*bytes.Buffer) []*Chunk {
```
danielbonilha: This method belongs to DLP business logic, I think it could be an non-exported variable `redactChunks` just for internal use. This method makes no sense outside the scope of its package.

agent/dlp/gcp.go
```diff
@@ -0,0 +1,202 @@
...
+			break
+		}
+		idx++
+	}
+	return redactedChunks
+}
+
+// JoinChunks will recompose the chunks into a unique buffer along with a list of
+// Transformations Summaries
+func JoinChunks(chunks []*Chunk) (*bytes.Buffer, []*TransformationSummary, error) {
```
danielbonilha: This method belongs to DLP business logic, I think it could be an non-exported variable `joinChunks` just for internal use. This method makes no sense outside the scope of its package.

agent/const.go
```diff
@@ -0,0 +1,12 @@
+package agent
+
+const (
+	connectionStoreParamsKey string = "params:%s"
+	procStoreKey             string = "proc:%s"
```
danielbonilha: This variable is being used anywhere? If not lets remove it.

common/proto/types.go
sandromello: Makes sense

common/dlp/gcp.go
sandromello: Done!

PR #55, opened 2022-11-11
Feat/review

danielbonilha:
1) Added REVIEW plugin.

If a connection has the review plugin registered, then the command will be persisted in the DB as a review item
Currently there is a limitation in the usage due to the connection mode send char by char to the agent

** It requires a EXEC type (one-off) to work properly, and this PR does not address that for now **

** Since it requires the EXEC verb, no blocks were introduced, and the message will flow all along the way, despite the review required. It will be persisted to the DB, but won't block the command from the user

2) Added 3 new endpoints:
- GET /reviews
- GET /reviews/:id
- PUT /reviews/:id

The put is to approve or reject a command. Many rules applies:
- Only PENDING reviews can be reviewed.
- Only user with a allowed group can review
- One rejection in one group rejects all the review
- ALL approves is required to approve the whole review

### DIFF
--- a/gateway/review/storage.go
+++ b/gateway/review/storage.go
@@ -60,7 +60,13 @@ func (s *Storage) FindAll(context *user.Context) ([]Review, error) {
 
 func (s *Storage) FindById(context *user.Context, id string) (*Review, error) {
 	var payload = `{:query {
-		:find [(pull ?review [* {:review/connection [:xt/id :connection/name]}
+		:find [(pull ?review [:xt/id
+                              :review/status
+							  :review/command
+							  :review/session
+							  :review/connection
+							  :review/created-by
+                                 {:review/connection [:xt/id :connection/name]}
                                  {:review/review-groups [*
                                      {:review-group/reviewed-by [:xt/id :user/name :user/email]}]}
                                  {:review/created-by [:xt/id :user/name :user/email]}])]


gateway/review/service.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (s *Service) FindAll(context *user.Context) ([]Review, error) {
+	return s.Storage.FindAll(context)
+}
+
+func (s *Service) Persist(context *user.Context, review *Review) error {
+	if review.Id == "" {
+		review.Id = uuid.NewString()
+	}
```
sandromello, line 70: In this case you're allowing the user to provide a unique identifier for this, am I right?
What's the reason for that? I think it's more safe for us to control this since this ID is the document store of xtdb.

gateway/review/storage.go
```diff
@@ -0,0 +1,134 @@
...
+		*st.Storage
+	}
+
+	XtdbReview struct {
+		Id           string   `edn:"xt/id"`
+		OrgId        string   `edn:"review/org"`
+		SessionId    string   `edn:"review/session"`
+		ConnectionId string   `edn:"review/connection"`
+		CreatedBy    string   `edn:"review/created-by"`
+		Command      string   `edn:"review/command"`
```
sandromello, line 20: We really shouldn't be using the input as a top key item, otherwise we'll be indexing the content to memory on xtdb and we'll have problems very soon, see:
Ref: https://docs.xtdb.com/extensions/full-text-search/#_indexing

A solution is using an inner key to store it. This will make the key not indexed by default.
We already have these kind of problems recently in runops. I think we must address this as soon as possible. 

I did this for session document: https://github.com/hoophq/hoop/blob/main/gateway/session/service.go#L27

I don't see problems in merging this as it is, because it will require refactoring queries. However it should be marked as technical debt and fixed in our next sprint.

I also think that we shouldn't be calling this as `Command`, this is an input to a command (stdinInput), or a script (follows same concept in runops since this will be applied only to one-off scripts).

gateway/transport/plugins/review/review.go
```diff
@@ -4,49 +4,90 @@ import (
...
-		return fmt.Errorf("failed processing audit plugin, missing org_id and session_id params")
+		return fmt.Errorf("failed processing review plugin, missing org_id and session_id params")
 	}
 
 	return nil
 }
 
-func (r *reviewPlugin) OnReceive(sessionID string, config []string, pkt *pb.Packet) error {
-	log.Printf("[%s] Review OnReceive plugin with config %v and pkt %v", sessionID, config, pkt)
+func (r *reviewPlugin) OnReceive(pluginConfig plugin.Config, config []string, pkt *pb.Packet) error {
```
sandromello, line 61: You should be applying the review in the `OnConnect` function, in this phase you be able to get the input of the user. The `OnReceive` phase will be applied for every packet flowing which doesn't make sense to apply a review.

gateway/review/storage.go
```diff
@@ -0,0 +1,134 @@
...
+		Status       Status   `edn:"review/status"`
+		ReviewGroups []string `edn:"review/review-groups"`
+	}
+
+	XtdbGroup struct {
```
sandromello, line 25: We're calling this as `XtdbGroup`, is there any reason for not calling only `Group`? The other packages doesn't use `Xtdb` prefixes

gateway/review/service.go
danielbonilha: This piece of code is because we reuse the Persist method to both save and update. In the case of update, there will already have an ID, so we do not create a new one. 

But I agree that this allow users to create their own ID, which is bad. All entities for now are being persisted using this notation. Maybe we can create a technical debt to make sure there is no ID being created by the user in any entities.

gateway/review/storage.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/runopsio/hoop/gateway/user"
+	"olympos.io/encoding/edn"
+)
+
+type (
+	Storage struct {
+		*st.Storage
+	}
+
+	XtdbReview struct {
```
sandromello, line 14: We're calling this as `XtdbReview`, is there any reason for not calling only `Review`? The other packages doesn't use `Xtdb` prefixes

gateway/review/storage.go
danielbonilha: Yes, the reason is that we already have a Group and a Review in the service layer. Those in the service layer include joins, it is user friendly, while the xtdb series is just for modeling the persistence layer with IDs instead of real objects

gateway/review/storage.go
danielbonilha: Yes, the reason is that we already have a Group and a Review in the service layer. Those in the service layer include joins, it is user friendly, while the xtdb series is just for modeling the persistence layer with IDs instead of real objects

gateway/transport/plugins/review/review.go
danielbonilha: I agree. But only when we move towards the EXEC command. The way it is now, we woldnt be able to test it during connect, since no command is being sent. Maybe we postpone this move to when we have the EXEC in place?

gateway/review/storage.go
danielbonilha: Good point. I will add a technical debt for that.

gateway/transport/plugins/review/review.go
sandromello: Add a review in connect with empty or fake input, then you still be able to test it.

PR #56, opened 2022-11-16
Add DLP info type configuration

sandromello:
Apply DLP configuration if a dlp plugin is enabled and set for a connection.
If the plugin has a wrong info type it will fail to redact.

The configuration below applies redact to `admin` groups:

```json
{
  "name": "dlp",
  "connections": [
    {
      "id": "{{CONNECTION_ID}}",
      "name": "psql",
      "config": [],
      "groups": {
        "admin": ["EMAIL_ADDRESS", "URL"]
      }
    }
  ]
}
```

### DIFF
--- a/gateway/transport/plugins/dlp/dlp.go
+++ b/gateway/transport/plugins/dlp/dlp.go
@@ -1,0 +1,41 @@
+package dlp
+
+import (
+	"github.com/runopsio/hoop/gateway/plugin"
+
+	pb "github.com/runopsio/hoop/common/proto"
+)
+
+const Name string = "dlp"
+
+type (
+	dlpPlugin struct {
+		name string
+	}
+)
+
+func New() *dlpPlugin {
+	return &dlpPlugin{name: Name}
+}
+
+func (p *dlpPlugin) Name() string {
+	return p.name
+}
+
+func (p *dlpPlugin) OnStartup(config plugin.Config) error {
+	return nil
+}
+
+func (p *dlpPlugin) OnConnect(config plugin.Config) error {
+	return nil
+}
+
+func (p *dlpPlugin) OnReceive(pluginConfig plugin.Config, config []string, pkt *pb.Packet) error {
+	return nil
+}
+
+func (p *dlpPlugin) OnDisconnect(config plugin.Config) error {
+	return nil
+}
+
+func (p *dlpPlugin) OnShutdown() {}


gateway/transport/plugins/dlp/dlp.go
```diff
@@ -0,0 +1,41 @@
...
+func (p *dlpPlugin) OnStartup(config plugin.Config) error {
+	return nil
+}
+
+func (p *dlpPlugin) OnConnect(config plugin.Config) error {
+	return nil
+}
+
+func (p *dlpPlugin) OnReceive(pluginConfig plugin.Config, config []string, pkt *pb.Packet) error {
+	return nil
```
danielbonilha, line 34: I was thinking that here would be the place to set the config data to the spec, to be used later in the agent. We could avoid a bunch of dlp fields in the structs:
```
func (p *dlpPlugin) OnReceive(pluginConfig plugin.Config, config []string, pkt *pb.Packet) error {
	pkt.Spec["dlpConfig"] = config
        return nil
}
```

Then in the agent we check for presence of the "dlpCOnfig" in the spec. Does it make any sense?

gateway/transport/client.go
```diff
@@ -247,10 +250,18 @@ func (s *Server) processPacketGatewayConnect(pkt *pb.Packet,
...
+		}
+	}
 	encConnectionParams, err := pb.GobEncode(&pb.AgentConnectionParams{
-		EnvVars:    conn.Secret,
-		CmdList:    conn.Command,
-		ClientArgs: clientArgs,
+		EnvVars:      conn.Secret,
+		CmdList:      conn.Command,
+		ClientArgs:   clientArgs,
+		DLPInfoTypes: infoTypes,
```
danielbonilha, line 264: I am not sure about adding dlpFields here. It is a plugin, and it does not make much sense for all scenarios. I would use the dlpFields setting in the pkt.Spec, during OnReceive of the plugin. (https://github.com/hoophq/hoop/pull/56/files#r1024480011). 

Setting it on the OnReceive we could delete the whole for loop in lines 254-259.

common/proto/types.go
```diff
@@ -30,9 +30,10 @@ type (
 		packetSpec map[string][]byte
 	}
 	AgentConnectionParams struct {
-		EnvVars    map[string]any
-		CmdList    []string
-		ClientArgs []string
+		EnvVars      map[string]any
+		CmdList      []string
+		ClientArgs   []string
+		DLPInfoTypes []string
```
danielbonilha, line 36: I am not sure about adding dlpFields here. It is a plugin, and it does not make much sense for all scenarios. I would use the dlpFields setting in the pkt.Spec, during OnReceive of the plugin. (https://github.com/hoophq/hoop/pull/56/files#r1024480011).

gateway/transport/client.go
sandromello: Because of the introduction of `hoop connect`, we have a step where we first configure all dependencies for a session in the agent and afterwards we start to exchange packets. That's why I send the info types in this step.

Passing the DLP Fields in the `OnReceive` phase will require more refactoring in the agent and introduce unnecessary data flowing between the gateway and the agent.

gateway/transport/plugins/dlp/dlp.go
sandromello: See my comment in: https://github.com/hoophq/hoop/pull/56/files/a69f3f67de3dff69d11add1c99199187ed31341d#r1024580044

common/proto/types.go
sandromello: See my comment in https://github.com/hoophq/hoop/pull/56/files/a69f3f67de3dff69d11add1c99199187ed31341d#r1024580044

gateway/transport/client.go
sandromello: One thing that could make sense is introducing a phase for this step (GatewayConnect). This phase could return a packet spec containing the info types. Future plugins could implement this phase if they need to pass something to the agent

danielbonilha: What would be difference between the OnConnect and GatewayConnect? I expressed the OnReceive, but indeed the OnConnect makes more sense for this case.

sandromello: The OnConnect happens when a gRPC connection is happening, the GatewayConnect happens when the client is trying to establishes a connection with the agent. The naming seems weird, but we can change that.

sandromello: Think on GatewayConnect as a step to initializing the session in the agent. It store everything it needs for that session, performs validation in the remote infra-structure and returns an error to the client in case of a misconfiguration or if some resource is not available.

danielbonilha: I didnt get it. The OnConnect is the client connecting to the gateway, but this also implies that gateway will ping the agent to verify the connection, uh? Not sure why we need another GatewayConnect and cannot use the OnConnect for both things (client to gateway and gateway checking agent)

sandromello: Indeed we could put this in the same flow of grpcConnect, but we need further refactoring to do that.
I've created a technical debt for that:
https://3.basecamp.com/5385186/buckets/28171754/todos/5553242309

PR #57, opened 2022-11-17
Renaming refactor

danielbonilha:
- Changed the name of the "Exec" to "Terminal" in all places, to not confuse with the next EXEC implementation
- Renamed few PacketTypes to better understanding of the role of each packet
- Moved few things around based on the responsability

### DIFF
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -75,7 +75,6 @@ func runConnect(args []string) {
 }
 
 func (c *connect) processPacket(pkt *pb.Packet) {
-	// client agent first connect
 	switch pb.PacketType(pkt.Type) {
 	case pb.PacketClientAgentConnectOKType:
 		sessionID, ok := pkt.Spec[pb.SpecGatewaySessionID]
--- a/agent/pg/examples/standalone.go
+++ b/agent/pg/examples/standalone.go
@@ -7,6 +7,7 @@ package main
 import (
 	"context"
 	"fmt"
+	proxypg "github.com/runopsio/hoop/common/pg"
 	"log"
 	"net"
 	"os"
@@ -14,7 +15,6 @@ import (
 
 	"github.com/runopsio/hoop/agent/pg"
 	"github.com/runopsio/hoop/agent/pg/middlewares"
-	proxypg "github.com/runopsio/hoop/common/proxy"
 )
 
 const (
--- a/agent/pg/middlewares/default.go
+++ b/agent/pg/middlewares/default.go
@@ -6,12 +6,12 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
+	pgtypes "github.com/runopsio/hoop/common/pg"
 	"log"
 	"strings"
 
 	"github.com/lib/pq/scram"
 	"github.com/runopsio/hoop/agent/pg"
-	pgtypes "github.com/runopsio/hoop/common/proxy"
 )
 
 type middleware struct {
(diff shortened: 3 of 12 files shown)


agent/agent.go
```diff
@@ -130,10 +130,26 @@ func (a *Agent) Run(svrAddr, token string) {
 			close(a.closeSignal)
 			return
 		}
-		a.processAgentConnect(pkt)
-		a.processPGProtocol(pkt)
-		a.processTCPCloseConnection(pkt)
-		a.processExec(pkt)
+
+		switch pb.PacketType(pkt.Type) {
```
danielbonilha, line 134: This change is because when I first read the code, looked like that the agent was performing ALL those actions, instead of selecting one. So I flattened the switch case.

common/proxy/pg.go
```diff
@@ -1,12 +1,24 @@
...
+type (
+	PacketType     byte
+	AuthPacketType int
+)
 
 func (t PacketType) Byte() byte {
 	return byte(t)
 }
 
-// http://www.postgresql.org/docs/9.4/static/protocol-message-formats.html
```
sandromello: Don't remove comments that show where these types are derived from.

common/proxy/pg.go
```diff
@@ -1,12 +1,24 @@
-package types
+package proxy
```
sandromello: It's important to have at least types in it's respective packages (terminal, postgres, etc). Later on we'll be implementing MySQL and it will be difficult to make a distinction of types by names.

client/cmd/connect.go
```diff
@@ -120,47 +75,9 @@ func runConnect(args []string) {
...
-
-func (c *connect) printHeader(sessionID string) {
-	termenv.NewOutput(os.Stdout).ClearScreen()
-	s := termenv.String("connection: %s | session: %s").Faint()
-	fmt.Printf(s.String(), c.connectionName, string(sessionID))
-	fmt.Println()
-}
-
-func (c *connect) processAgentConnectPhase(pkt *pb.Packet) {
+	// client agent first connect
```
sandromello: What's this comment? process packet doesn't mean that it's the "client agent first connect"

client/cmd/exec.go
```diff
@@ -0,0 +1,49 @@
...
+	loader.Suffix = "exec'ing command..."
+
+	c, spec := connectionClient(config, loader, args)
+
+	if err := c.client.Send(&pb.Packet{
+		Type: pb.PacketClientGatewayExecType.String(),
+		Spec: spec,
+	}); err != nil {
+		_, _ = c.client.Close()
+		c.printErrorAndExit("failed connecting to gateway, err=%v", err)
```
sandromello, line 39: You'll need a session id to send packets to the agent. Sending a gateway connect packet is necessary for that.

common/proxy/pg.go
danielbonilha: Ok, I will put it back. Tks.

common/proxy/pg.go
danielbonilha: I will put them in separated folders

client/cmd/connect.go
danielbonilha: I removed it.

client/cmd/exec.go
danielbonilha: I kinda copied from here: https://github.com/hoophq/hoop/blob/main/client/cmd/connect.go#L107

sandromello: So you put the wrong type here, it's `pb.PacketGatewayConnectType.String()`. But if this it's only a sketch, you can work it out in the next Pull Request, there's no problem

PR #58, opened 2022-11-18
Add DLP to postgres protocol

sandromello:
This Pull Request adds DLP to postgres protocol

- In case of failing to redact, the original content will be returned
- A max of 30 info types is allowed only, an error is returned if this limit is reached.
- Enable DLP by default on new connections
- If the configuration has empty info types the redact will be a noop.
- Add default info types

## Other Changes

- Fixed audit bug of output messages not being processed in the agent


gateway/connection/service.go
```diff
@@ -98,3 +99,31 @@ func (s *Service) bindAuditPlugin(context *user.Context, conn *Connection) {
 		return
 	}
 }
+
+func (s *Service) bindDLPPlugin(context *user.Context, conn *Connection) {
+	p, err := s.PluginService.FindOne(context, "dlp")
```
danielbonilha: Lets use `dlp.Name` here.

gateway/connection/service.go
```diff
@@ -98,3 +99,31 @@ func (s *Service) bindAuditPlugin(context *user.Context, conn *Connection) {
...
+			if c.ConnectionId == conn.Id {
+				registered = true
+			}
+		}
+		if !registered {
+			p.Connections = append(p.Connections, plugin.Connection{ConnectionId: conn.Id})
+		}
+	} else {
+		p = &plugin.Plugin{
+			Name:        "dlp",
```
danielbonilha: Lets use `dlp.Name` here.

agent/dlp/types.go
```diff
@@ -14,42 +15,60 @@ const (
...
+	defaultMaxChunkSize = 62500
+	maxFindings         = 3000
+	maxInfoTypes        = 30
 )
 
 type (
-	TransformationSummary struct {
-		Index int
-		Err   error
+	Client interface {
```
danielbonilha, line 25: It is ok to leave it here, but I do not understand the need of an interface here. Could be the simple dlp.Client being returned as is. Normally the interface is declared where it is used, and not where it is implemented.

gateway/connection/service.go
danielbonilha: Umm... I see the problem... it will import a package that does not make sense. Actually I think both methods are in the wrong place (bindAuditPlugin, bindDLPPlugin).

Forget about this comment for now, lets refactor this later.

gateway/connection/service.go
danielbonilha: Umm... I see the problem... it will import a package that does not make sense. Actually I think both methods are in the wrong place (bindAuditPlugin, bindDLPPlugin).

Forget about this comment for now, lets refactor this later.

agent/dlp/types.go
sandromello: I've created the dlp.Client interface to mock things in tests, see `gcp_test.go#fakeClient`.
There's other packages that follows this patterns, for example the go standard library:
- io.Writer
- http.Handler

These interfaces are being used where they're implemented, but they could also be used where they are not implemented.

When you mean "normally" what are you referring to?

PR #63, opened 2022-11-23
Feat/exec command

danielbonilha:
- Added "exec" verb to hoop client.
- The exec does NOT keep a persistent connection, it starts, runs and dies.
- The connection used, if has review configured, will block the exec until there is an approval in place
- After approval, user will be prompted to run the script
- If task rejected then user is disconnected
- If user drops the exec before being approved, then need to start over
- If user tries to connect to a connection that requires review, an error will be returned
- DLP is not implemented in this PR for exec

### DIFF
--- a/gateway/transport/client.go
+++ b/gateway/transport/client.go
@@ -236,7 +236,12 @@ func (s *Server) processClientPacket(pkt *pb.Packet, client *client.Client, conn
 	case pb.PacketClientGatewayExecType:
 		return s.processClientExec(pkt, client, conn)
 	default:
-		//_ = agentStream.Send(pkt)
+		agentStream := getAgentStream(conn.AgentId)
+		if agentStream == nil {
+			log.Printf("agent not found for connection %s", conn.Name)
+			return status.Errorf(codes.FailedPrecondition, fmt.Sprintf("agent not found for connection %s", conn.Name))
+		}
+		_ = agentStream.Send(pkt)
 	}
 	return nil
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -165,13 +165,13 @@ func (a *Agent) Run(svrAddr, token string) {
 	}
 }
 
-func (a *Agent) buildConnectionParams(pkt *pb.Packet, packetErrType pb.PacketType) (*pb.AgentConnectionParams, error) {
+func (a *Agent) buildConnectionParams(pkt *pb.Packet, packetErrType pb.PacketType) (*pb.AgentConnectionParams, *connEnv, error) {
 	sessionID := pkt.Spec[pb.SpecGatewaySessionID]
 	sessionIDKey := string(sessionID)
 
 	connParams := a.decodeConnectionParams(sessionID, pkt, packetErrType)
 	if connParams == nil {
-		return nil, fmt.Errorf("session %s failed to decode connection params", sessionIDKey)
+		return nil, nil, fmt.Errorf("session %s failed to decode connection params", sessionIDKey)
 	}
 	log.Printf("session=%s - connection params decoded with success, dlp-info-types=%d",
 		sessionIDKey, len(connParams.DLPInfoTypes))
@@ -179,19 +179,17 @@ func (a *Agent) buildConnectionParams(pkt *pb.Packet, packetErrType pb.PacketTyp
 	connType := string(pkt.Spec[pb.SpecConnectionType])
 	connEnvVars, err := parseConnectionEnvVars(connParams.EnvVars, connType)
 	if err != nil {
-		return nil, fmt.Errorf("session %s failed to parse env vars", sessionIDKey)
+		return nil, nil, fmt.Errorf("session %s failed to parse env vars", sessionIDKey)
 	}
 	if connType == pb.ConnectionTypePostgres || connType == pb.ConnectionTypeTCP {
 		if err := isPortActive(connEnvVars.host, connEnvVars.port); err != nil {
 			msg := fmt.Sprintf("session=%s - failed connecting to host=%q, port=%q, err=%v",
 				sessionIDKey, connEnvVars.host, connEnvVars.port, err)
 			log.Printf(msg)
-			return nil, fmt.Errorf("%s", msg)
+			return nil, nil, fmt.Errorf("%s", msg)
 		}
 	}
-
-	//connParams.EnvVars[connEnvKey] = connEnvVars
-	return connParams, nil
+	return connParams, connEnvVars, nil
 }
 
 func (a *Agent) decodeConnectionParams(sessionID []byte, pkt *pb.Packet, packetType pb.PacketType) *pb.AgentConnectionParams {
@@ -236,7 +234,7 @@ func (a *Agent) processClientConnect(pkt *pb.Packet) {
 	packetErrType := pb.PacketClientAgentConnectErrType
 	log.Printf("session=%s - received connect request", sessionIDKey)
 
-	connParams, err := a.buildConnectionParams(pkt, packetErrType)
+	connParams, connEnvVars, err := a.buildConnectionParams(pkt, packetErrType)
 	if err != nil {
 		_ = a.client.Send(&pb.Packet{
 			Type:    packetErrType.String(),
@@ -246,6 +244,15 @@ func (a *Agent) processClientConnect(pkt *pb.Packet) {
 	}
 
 	dlpClient := a.decodeDLPCredentials(sessionID, pkt, packetErrType)
+
+	connType := string(pkt.Spec[pb.SpecConnectionType])
+	if connType == pb.ConnectionTypePostgres || connType == pb.ConnectionTypeTCP {
+		connParams.EnvVars[connEnvKey] = connEnvVars
+	}
+
+	if connType == pb.ConnectionTypeCommandLine {
+		sessionIDKey = fmt.Sprintf(connectionStoreParamsKey, string(sessionID))
+	}
 
 	a.connStore.Set(dlpClientKey, dlpClient)
 	a.connStore.Set(sessionIDKey, connParams)
@@ -281,7 +288,7 @@ func (a *Agent) doExec(pkt *pb.Packet) {
 	packetErrType := pb.PacketClientAgentExecErrType
 	log.Printf("session=%v - received execution request", string(sessionID))
 
-	connParams, err := a.buildConnectionParams(pkt, packetErrType)
+	connParams, _, err := a.buildConnectionParams(pkt, packetErrType)
 	if err != nil {
 		_ = a.client.Send(&pb.Packet{
 			Type:    packetErrType.String(),
--- a/client/proxy/terminal.go
+++ b/client/proxy/terminal.go
@@ -75,7 +75,6 @@ func (t *Terminal) ConnectWithTTY() error {
 		_ = ptty.Close()
 		_ = tty.Close()
 	}()
-	log.Println("FINISHED CONNECT_WITH_TTY")
 	return nil
 }
 


client/cmd/connect.go
```diff
@@ -139,6 +145,58 @@ func (c *connect) processPacket(pkt *pb.Packet) {
...
+		if result == "y" {
+			pkt.Type = string(pb.PacketClientGatewayExecType)
+			_ = c.client.Send(pkt)
+			return
+		}
+		c.processGracefulExit(errors.New("user aborted"))
+	case pb.PacketClientGatewayExecWaitType:
+		c.waitingReview = pkt
+		fmt.Println("This command requires review. We will notify you right here when it is approved")
+		return
```
sandromello: This return is not needed here

client/cmd/connect.go
```diff
@@ -139,6 +145,58 @@ func (c *connect) processPacket(pkt *pb.Packet) {
...
+			c.waitingReview.Type = string(pb.PacketClientGatewayExecType)
+			_ = c.client.Send(c.waitingReview)
+			c.waitingReview = nil
+			return
+		}
+		c.processGracefulExit(errors.New("user cancelled the action"))
+	case pb.PacketClientGatewayExecRejectType:
+		c.processGracefulExit(errors.New("task rejected. Sorry"))
+	case pb.PacketClientAgentExecOKType:
+		c.printOutputAndExit(pkt.Payload)
```
sandromello, line 200: The packet will have the exitcode which can be used to propagate that to the client. Then you could:
```sh
os.Exit(agentExitCode)
```

See: https://github.com/hoophq/hoop/blob/1e5121446418e05a7fa46d6ea4b1fb38e5de4034/agent/terminal.go#L41

client/cmd/exec.go
```diff
@@ -17,28 +26,78 @@ var execCmd = &cobra.Command{
 }
 
 func init() {
-	execCmd.Flags().StringVarP(&connectFlags.proxyPort, "file", "f", "", "The path of the file containing the command")
+	execCmd.Flags().StringVarP(&commandPath, "file", "f", "", "The path of the file containing the command")
+	execCmd.Flags().StringVarP(&command, "command", "c", "", "The command to run remotely")
```
sandromello: - The "command" should be calling "input" or maybe "script" (same as runops), but command seems strange because connections already defines "commands".
- These options must not be required, because the user can execute a command from a connection without passing any input.

### TL;DR

These commands accepts the input as stdin, so it's not a command it's an input and I think we should expose it as `input` instead of `command`.

The user already has the ability to run it as a command without passing any input:

```sh
# connection bash
hoop exec mybash -- ls -l
# connection kubectl
hoop exec k8s -- get pods
# connection aws-ecs
hoop exec awsecs -- --cluster ecs-cluster-name --command "mycommand"
# connection psql
hoop exec psql -- -P pager=off -c "SELECT NOW()"
```

So having a command with empty `command` and `file` it's allowed on exec, because the user could compose it using args.

The "command" and "file" makes sense when you're passing an input to stdin only, a few examples:

```sh
# you can still pass args
hoop exec psql -f /tmp/myquery.sql -- -P pager=off
# this is the same
psql <<EOF
<contents of /tmp/myquery.sql>
EOF
```

```sh
hoop exec bash -f /tmp/myscript.sh
# this is the same as
bash <<EOF
<contents of /tmp/script.sh>
EOF
```

- The connection allows passing arguments to a command and letting you execute commands
- The input only makes sense if the connection command accepts the content from stdin

client/cmd/exec.go
```diff
@@ -17,28 +26,78 @@ var execCmd = &cobra.Command{
...
+				pkt.Payload = append(pkt.Payload, input)
+			}
+			fmt.Println(string(pkt.Payload))
+			pkt.Payload = []byte(strings.Trim(string(pkt.Payload), " \n"))
+		}
+	}
+
+	if pkt.Payload == nil {
+		c.printErrorAndExit("missing command, please run 'hoop exec help'")
+	}
```
sandromello: This should not be required, see my comment above

client/cmd/exec.go
```diff
@@ -17,28 +26,78 @@ var execCmd = &cobra.Command{
...
+
+	pkt.Payload = []byte(strings.Trim(string(pkt.Payload), " \n"))
```
sandromello: Why do you need to do this?

common/proto/const.go
```diff
@@ -6,19 +6,25 @@ const (
 	PacketKeepAliveType          PacketType = "KeepAlive"
 	PacketCloseTCPConnectionType PacketType = "CloseTCPConnection"
 
-	// client starting new connection
-	PacketClientGatewayExecType    PacketType = "Client::Gateway::Exec"
-	PacketClientGatewayConnectType PacketType = "Client::Gateway::Connect"
+	// client->agent connection
```
sandromello, line 9: You're not using the types:
- PacketClientGatewayConnectErrType
- PacketClientGatewayExecErrType
- PacketClientExecAgentOfflineType

Why are you defining them here?

agent/terminal.go
```diff
@@ -12,41 +12,43 @@ import (
...
 			pb.SpecGatewaySessionID: sessionID,
-		}).Write([]byte(`failed executing command`))
+		}).Write([]byte(err.Error()))
 		return
 	}
 	log.Printf("session=%v, tty=false - executing command=%q", string(sessionID), cmd.String())
-	spec := map[string][]byte{pb.SpecGatewaySessionID: sessionID}
-	stdoutWriter := pb.NewStreamWriter(a.client, pb.PacketTerminalClientWriteStdoutType, spec)
+
+	stdoutWriter := pb.NewStreamWriter(a.client, pb.PacketClientAgentExecOKType, pkt.Spec)
```
sandromello: Exec Ok means the command executed with success (exit-code) and you don't known that in advance, maybe calling `ExecResponse` makes more sense.

agent/terminal.go
```diff
@@ -12,41 +12,43 @@ import (
...
 		log.Printf("failed executing command, err=%v", err)
-		_, _ = pb.NewStreamWriter(a.client, pb.PacketTerminalCloseType, map[string][]byte{
+		_, _ = pb.NewStreamWriter(a.client, pb.PacketClientAgentExecErrType, map[string][]byte{
 			pb.SpecGatewaySessionID: sessionID,
-		}).Write([]byte(`failed executing command`))
+		}).Write([]byte(err.Error()))
 		return
 	}
 	log.Printf("session=%v, tty=false - executing command=%q", string(sessionID), cmd.String())
-	spec := map[string][]byte{pb.SpecGatewaySessionID: sessionID}
```
sandromello, line 37: Don't propagate the spec back to client, you could be leaking unwanted data to the client, this is not safe.

gateway/transport/client.go
```diff
@@ -153,13 +155,27 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
+	agentStream := getAgentStream(conn.AgentId)
+	if agentStream == nil {
+		log.Printf("agent not found for connection %s", connectionName)
+		return status.Errorf(codes.FailedPrecondition, fmt.Sprintf("agent not found for %v", c.AgentId))
+	}
+
+	if clientVerb == pb.ClientVerbConnect {
+		for _, p := range plugins {
+			if p.Plugin.Name() == pluginsreview.Name {
+				return status.Errorf(codes.PermissionDenied, fmt.Sprintf("this connection have the review plugin installed and cannot be connected.\n\nPlease use 'hoop exec %s {command}' instead", conn.Name))
```
sandromello: There's too much information to the user, the user doesn't even are aware of what's a plugin (developer). A better error message:
> This connection is subject to review. Please, use 'hoop exec ...` to interact with this connection

gateway/transport/client.go
```diff
@@ -281,6 +298,69 @@ func (s *Server) processClientConnect(pkt *pb.Packet,
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed encoding connection params err=%v", err)
+	}
+
+	pkt.Spec[pb.SpecConnectionType] = []byte(conn.Type)
+	pkt.Spec[pb.SpecAgentConnectionParamsKey] = encConnectionParams
+
+	_ = agentStream.Send(pkt)
+	return nil
```
sandromello: This is not even needed if you use the connect flow already implemented.
The connect flow already validates if the agent is alive and initializes a sessions with all credentials/config set

The `PacketClientGatewayExecWaitType` message could be implemented in the client after receiving a connect response.
I think we could move forward but it's good to create a task for addressing this in the near future, because now we have two ways of initializing a session between the client and session.

client/cmd/connect.go
```diff
@@ -139,6 +145,58 @@ func (c *connect) processPacket(pkt *pb.Packet) {
...
+
+		if input == "" {
+			input = "y"
+		}
+
+		input = input[0:1]
+
+		if input == "y" {
+			c.waitingReview.Type = string(pb.PacketClientGatewayExecType)
+			_ = c.client.Send(c.waitingReview)
```
sandromello, line 192: So in practice after receiving an approve the client could fake the output of the command and execute something else. I think it's ok for this version but we should add a technical debt to fix that.

gateway/transport/plugins/review/review.go
```diff
@@ -88,6 +100,9 @@ func (r *reviewPlugin) OnReceive(pluginConfig plugin.Config, config []string, pk
 		}, review); err != nil {
 			return err
 		}
+
+		b, _ := pb.GobEncode(review)
```
sandromello, line 104: Why are you gob encoding this whole type since you're using only the status to validate the review status? You could only add the review as string and compare it.

gateway/transport/client.go
```diff
@@ -281,6 +298,69 @@ func (s *Server) processClientConnect(pkt *pb.Packet,
...
+		ClientArgs: clientArgs,
+	})
+	if err != nil {
+		return fmt.Errorf("failed encoding connection params err=%v", err)
+	}
+
+	pkt.Spec[pb.SpecConnectionType] = []byte(conn.Type)
+	pkt.Spec[pb.SpecAgentConnectionParamsKey] = encConnectionParams
+
+	_ = agentStream.Send(pkt)
```
sandromello: Try to not mutate the packet the best you can, because we could starting leaking unwanted data, it's also hard to track which specs are being propagated to clients (agent/proxy)

gateway/transport/client.go
```diff
@@ -281,6 +298,69 @@ func (s *Server) processClientConnect(pkt *pb.Packet,
...
+	existingReviewData := pkt.Spec[pb.SpecReviewDataKey]
+	if existingReviewData != nil {
+		var review rv.Review
+		if err := pb.GobDecodeInto(existingReviewData, &review); err != nil {
+			return err
+		}
+		if review.Status != rv.StatusApproved {
+			pkt.Type = string(pb.PacketClientGatewayExecWaitType)
+			clientStream := getClientStream(client.SessionID)
+			_ = clientStream.Send(pkt)
```
sandromello: You're leaking the review struct.
Try to not mutate the packet the best you can, because we could starting leaking unwanted data, it's also hard to track which specs are being propagated to clients (agent/proxy).

client/cmd/connect.go
```diff
@@ -139,6 +145,58 @@ func (c *connect) processPacket(pkt *pb.Packet) {
 			string(sessionID), string(pkt.GetPayload()))
 		c.processGracefulExit(errMsg)
 
+	// exec
+	case pb.PacketClientExecAgentOfflineType:
```
sandromello, line 149: This is not used in any flow. Let's try to avoid this flow for now since there's other refactors to be done in this PR

client/cmd/connect.go
danielbonilha: I will provide the gateway return code to the medhod.

client/cmd/exec.go
danielbonilha: Will change the "command" nomenclature to use "input" instead, as this makes more sense to what it really is.

client/cmd/exec.go
danielbonilha: Yes, will remove.

client/cmd/exec.go
danielbonilha: To trim the input.

client/cmd/connect.go
danielbonilha: Lets use the payload from the DB (fetched in gateway) as the final command to be executed, so we avoid this exploit. This packet will server as metadata holder for the approval http interaction.

gateway/transport/plugins/review/review.go
danielbonilha: We are going to use the payload as from now, so it is better to keep the whole structure.

gateway/transport/client.go
danielbonilha: It is worth revisiting this flows to not reuse to package (mainly the spec data) back to the client.

client/cmd/connect.go
danielbonilha: This guy is supposed to be used, so I will fix the flow to send the agent offline info to the client in such cases.

agent/terminal/server-exec.go
```diff
@@ -21,7 +21,7 @@ type Command struct {
 	ptty     *os.File
 }
 
-type OnExecEndFn func(exitCode int, errMsg string, a ...any)
+type OnExecErrFn func(exitCode int, errMsg string, a ...any)
```
sandromello, line 24: You're right, this is indeed onExecErr

client/cmd/connect.go
```diff
@@ -149,13 +149,18 @@ func (c *connect) processPacket(pkt *pb.Packet) {
...
 			c, _ := reader.ReadByte()
-			result = strings.Trim(string(c), " \n")
+			result = string(c)
+			result = strings.Trim(result, " \n")
 			break
 		}
 
+		if result == "" {
+			result = "y"
+		}
```
sandromello, line 162: If the user type enter, without any character it will be executed. Is this intended ?

gateway/transport/client.go
```diff
@@ -312,37 +289,86 @@ func (s *Server) processClientExec(pkt *pb.Packet,
...
+		payload = []byte(review.Input)
 	}
 
+	if s.GcpDLPRawCredentials != "" {
+		spec[pb.SpecAgentGCPRawCredentialsKey] = []byte(s.GcpDLPRawCredentials)
+	}
+
+	agentStream := getAgentStream(conn.AgentId)
+	if agentStream == nil {
+		spec[pb.SpecClientExecArgsKey] = pkt.Spec[pb.SpecClientExecArgsKey]
```
sandromello, line 309: You're allowing the client to use a different argument in this flow (offline), you need to the same as you did with `payload` a few lines above.

gateway/transport/client.go
```diff
@@ -155,16 +155,16 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
 		return status.Errorf(codes.FailedPrecondition, err.Error())
 	}
 
-	agentStream := getAgentStream(conn.AgentId)
-	if agentStream == nil {
-		log.Printf("agent not found for connection %s", connectionName)
-		return status.Errorf(codes.FailedPrecondition, fmt.Sprintf("agent not found for %v", c.AgentId))
-	}
+	//agentStream := getAgentStream(conn.AgentId)
```
sandromello: You could remove this if it's not being used.

danielbonilha: Done

client/cmd/connect.go
danielbonilha: Yes, the default is yes

PR #65, opened 2022-11-28
agent retry

danielbonilha:
Agent retry to connect to gateway after unauthenticated trial.
Will try every 5 seconds, 100 times.

### DIFF
--- a/gateway/transport/client.go
+++ b/gateway/transport/client.go
@@ -224,7 +224,7 @@ func (s *Server) listenClientMessages(
 		err = s.processClientPacket(pkt, c, conn)
 		if err != nil {
 			fmt.Printf("session=%v - failed processing client packet, err=%v", c.SessionID, err)
-			return err
+			return status.Errorf(codes.FailedPrecondition, err.Error())
 		}
 	}
 }


gateway/transport/agent.go
```diff
@@ -82,6 +82,11 @@ func (s *Server) subscribeAgent(stream pb.Transport_ConnectServer, token string)
 	s.agentGracefulShutdown(ag)
 
 	log.Printf("successful connection hostname: [%s], machineId [%s], kernelVersion [%s]", hostname, machineId, kernelVersion)
+	stream.Send(&pb.Packet{
+		Type:    string(pb.PacketAgentGatewayConnectOK),
```
sandromello, line 86: The agent doesn't have anything to do with this flow, it makes sense to rename to `PacketGatewayConnectOK`

agent/main.go
```diff
@@ -33,22 +35,49 @@ func Run() {
...
+	return nil
+}
+
+func connect(conf *Config) (pb.ClientTransport, error) {
+	client, err := grpc.Connect(conf.ServerAddress, conf.Token, grpc.WithOption("origin", pb.ConnectionOriginAgent))
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
```
sandromello: Didn't understood this function, you could use directly:
```go
client, err := grpc.Connect(conf.ServerAddress, conf.Token, grpc.WithOption("origin", pb.ConnectionOriginAgent))
```

If you keep choose to keep it, then there's also no need to keep the err checks because you're not doing anything, you could return the output of the function instead:
```go
(...)
return grpc.Connect(conf.ServerAddress, conf.Token, grpc.WithOption("origin", pb.ConnectionOriginAgent))
```

danielbonilha: Indeed, no need for that. Will remove.

PR #66, opened 2022-11-29
Enhanc/send review url back to cli

danielbonilha:
- Send review URL to final user when input under review.
- Fixed BUG for stdin when user executed the input via | (grep) or EOF. The default stdin was the files, and therefore the stdin from keyboard when under review was failing to get from tty.

### DIFF
--- a/client/cmd/exec.go
+++ b/client/cmd/exec.go
@@ -69,9 +69,7 @@ func runExec(args []string) {
 	if len(pkt.Payload) > 0 {
 		pkt.Payload = []byte(strings.Trim(string(pkt.Payload), " \n"))
 	}
-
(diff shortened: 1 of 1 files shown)


client/cmd/exec.go
```diff
@@ -57,30 +57,21 @@ func runExec(args []string) {
...
+			input += s
 		}
+		pkt.Payload = []byte(input)
 	}
 
 	if len(pkt.Payload) > 0 {
 		pkt.Payload = []byte(strings.Trim(string(pkt.Payload), " \n"))
 	}
 
+	fmt.Println("### " + string(pkt.Payload))
```
sandromello: Is this necessary?

danielbonilha: Not at all.

PR #69, opened 2022-11-30
Exec and Demo Bugfixes & Enhancements

sandromello:
- [exec] Add a channel to proper call Wait function when using command pipes
- [exec] Error from agents now are propagated as stderr instead of stdout Prevent printing additional break lines
- [exec] Fix bug on parsing flags and stdin input from exec
  Reading content from stdin was causing a bug when passing a command without any input. It was being blocked in scanner 
- [demo] Add option to call the k8s demo to create multiple connections
- [demo] Add a hint to duplicated connections on k8s-demo
- [demo] Add cleanup instruction to k8s-demo

### DIFF
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -8,13 +8,16 @@ import (
 	"log"
 	"os"
 	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/briandowns/spinner"
 	"github.com/muesli/termenv"
+	"github.com/runopsio/hoop/client/cmd/styles"
 	"github.com/runopsio/hoop/client/proxy"
 	"github.com/runopsio/hoop/common/memory"
 	pb "github.com/runopsio/hoop/common/proto"
+	"github.com/runopsio/hoop/common/terminal"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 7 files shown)


client/cmd/connect.go
```diff
@@ -224,8 +227,14 @@ func (c *connect) processPacket(pkt *pb.Packet, config *Config) {
...
-		c.processGracefulExit(errors.New(string(pkt.Payload)))
-
+		if len(pkt.Payload) > 0 {
+			_, _ = os.Stderr.Write([]byte(styles.ClientError(string(pkt.Payload)) + "\n"))
+		}
+		exitCode, err := strconv.Atoi(string(pkt.Spec[pb.SpecClientExecExitCodeKey]))
+		if err != nil {
+			os.Exit(terminal.InternalErrorExitCode)
+		}
+		os.Exit(exitCode)
```
sandromello, line 237: Explanation: We need to propagate the exit code properly from the agent and write errors to stderr

client/cmd/exec.go
```diff
@@ -38,38 +86,14 @@ func runExec(args []string) {
...
-
-	if pkt.Payload == nil {
-		var input string
-		scanner := bufio.NewScanner(os.Stdin)
-		for scanner.Scan() {
-			s := scanner.Text()
-			input += s
-		}
-		pkt.Payload = []byte(input)
-	}
```
sandromello, line 67: I refactored the whole input thing because this was causing problems when executing a command without any input. Example:
```sh
hoop exec k8s-demo -- get namespaces
```

This condition is now triggered blocking the program because there's any input

PR #78, opened 2022-12-06
Scrips/delete org

danielbonilha:
- Required to be able to reach some xtdb server (127.0.0.1:3000 for instance)
- To interact with production, change the host to 127.0.0.1:8999
- Collects the org name as parameter via argument or stdin

**Deletes:**
  - Reviews
  - Plugins
  - Sessions
  - Connections
  - Agents
  - Users
  - Org itself

### DIFF
--- a/scripts/org-cleanup.go
+++ b/scripts/org-cleanup.go
@@ -279,7 +279,7 @@ func getItems(ednQuery []byte) ([]string, error) {
 func buildSubmitTxPayload(IDs []string) ([]byte, error) {
 	txOps := make([]any, 0)
 	for _, id := range IDs {
-		txOps = append(txOps, []any{"delete", id})
+		txOps = append(txOps, []any{"evict", id})
 	}
 	return json.Marshal(map[string]any{"tx-ops": txOps})
 }


scripts/org-cleanup.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	fmt.Printf("Found %d items\n", len(innerResult))
+	return innerResult, nil
+}
+
+func buildSubmitTxPayload(IDs []string) ([]byte, error) {
+	txOps := make([]any, 0)
+	for _, id := range IDs {
+		txOps = append(txOps, []any{"delete", id})
```
sandromello: Delete will not fetch all documents, depending on how we structured, this won't cleanup things properly.
Maybe evict is more suitable for this task: 
https://docs.xtdb.com/language-reference/datalog-transactions/#evict

scripts/org-cleanup.go
```diff
@@ -0,0 +1,285 @@
...
+
+	fmt.Printf("Cleaning up org [%s]...\n", org)
+
+	if err := cleanupOrg(org); err != nil {
+		fmt.Printf("Failed with error: %s\n", err.Error())
+		os.Exit(1)
+	}
+}
+
+func cleanupOrg(org string) error {
```
sandromello, line 54: ## Just thinking out loud
Maybe there's a better way to do that, like:
Find all resources that contains an org-id X. This would bring all xt/id references of an organization id.

andriosrobert: +1

danielbonilha: So far, we need to specify the {namespace}/{org} to do that, so :session/org, :user/org and so on. 

Please let me know if you do know a way to execute the suggestion.

scripts/org-cleanup.go
danielbonilha: Nice. Done.

PR #79, opened 2022-12-07
Bugfix/hoop demo/start

sandromello:
This PR fixes a bug where the demo is broken if the user has already logged in.

Ref: https://3.basecamp.com/5385186/buckets/28171754/todos/5616937470

### DIFF
--- a/common/clientconfig/clientconfig.go
+++ b/common/clientconfig/clientconfig.go
@@ -6,8 +6,8 @@ import (
 )
 
 const (
-	AgentFile string = "agent.toml"
-	ProxyFile string = "config.toml"
+	AgentFile  string = "agent.toml"
+	ClientFile string = "config.toml"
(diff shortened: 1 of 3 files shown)


common/clientconfig/clientconfig.go
```diff
@@ -0,0 +1,33 @@
+package clientconfig
+
+import (
+	"fmt"
+	"os"
+)
+
+const (
+	AgentFile string = "agent.toml"
+	ProxyFile string = "config.toml"
```
danielbonilha: We do not use proxy anywhere in the project to reference the client. Maybe it will be confusing to others. Maybe we should call it differently (ClientFile or ClientConfigFile or any other)

sandromello: Variable renamed

PR #80, opened 2022-12-08
Add ecs helper script

sandromello:
- Add `ecs-exec.sh` to address some limitations of the command `aws ecs execute-command` such as:
  - Prevent input from leaking shell because it doesn't accept content from stdin.
  - List a task based on the service and cluster name
  - Toggle the interactive / non-interactive options [using unbuffer](https://github.com/aws/amazon-ssm-agent/issues/354)
- Add permission `0600` by default to env-var persisted in the filesystem. It address problems with tools requiring a more strict permission of identity files such as `ssh`.
- Refactor Dockerfile

Related to https://github.com/hoophq/docs/pull/3

### DIFF
--- a/rootfs/usr/local/bin/ecs-exec.sh
+++ b/rootfs/usr/local/bin/ecs-exec.sh
@@ -99,6 +99,8 @@ if [ "$SHELL_INTERACTIVE" == "1" ]; then
   exit $?
 fi
 
+# unbuffer is required when running one-off tasks
+# https://github.com/aws/amazon-ssm-agent/issues/354#issuecomment-817274498
 STDIN_INPUT=$(cat -)
 if [ -n "$PIPE_EXEC" ]; then
   STDIN_INPUT="$(base64 <<< $STDIN_INPUT)"


rootfs/usr/local/bin/ecs-exec.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
```
sandromello, line 1: Ref: https://github.com/hoophq/docs/pull/3

k8s/deployment-agent.yaml
```diff
@@ -22,6 +22,6 @@ spec:
         envFrom:
         - secretRef:
             name: default-agent
-        image: hoophq/hoop
+        image: hoophq/hoopdev
```
sandromello, line 25: amd64 image. It has more tools on it

Dockerfile.dev
```diff
@@ -62,14 +61,6 @@ RUN curl -sL https://download.clojure.org/install/linux-install-$CLOJURE_VERSION
     rm clojure-install.sh && \
     clojure -e "(clojure-version)"
 
-RUN curl -sL https://hoopartifacts.s3.amazonaws.com/xtdb-in-memory-1.22.0-aarch64.tar.gz -o /app/xtdb-in-memory-1.22.0-aarch64.tar.gz && \
```
sandromello, line 65: no need for amd64 image

PR #81, opened 2022-12-12
Feat/access control

danielbonilha:
- Added access_control plugin. 
- Reads the plugin "access_control" for every get and list connections and perform the filtering in-memory
- For efficiency we should do the filtering at DB level with a single query that fetches/filters it as a whole (as next improvement)
- This affects all scopes of the system: grpc and http interactions

### DIFF
--- a/gateway/connection/service.go
+++ b/gateway/connection/service.go
@@ -4,7 +4,7 @@ import (
 	"github.com/google/uuid"
 	pb "github.com/runopsio/hoop/common/proto"
 	"github.com/runopsio/hoop/gateway/plugin"
-	pluginsrbac "github.com/runopsio/hoop/gateway/transport/plugins/access_control"
+	pluginsrbac "github.com/runopsio/hoop/gateway/transport/plugins/accesscontrol"
 	pluginsaudit "github.com/runopsio/hoop/gateway/transport/plugins/audit"
 	pluginsdlp "github.com/runopsio/hoop/gateway/transport/plugins/dlp"
 	"github.com/runopsio/hoop/gateway/user"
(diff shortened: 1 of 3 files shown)


gateway/transport/plugins/access_control/access_control.go
```diff
@@ -0,0 +1,42 @@
+package access_control
```
sandromello: avoid using underscore as package names / folders / files. There's no problem in using `accesscontrol/accesscontrol.go`, It's more aligned with conventions writing GO code.

TL;DR: https://go.dev/doc/effective_go#package-names

gateway/connection/service.go
```diff
@@ -4,7 +4,9 @@ import (
 	"github.com/google/uuid"
 	pb "github.com/runopsio/hoop/common/proto"
 	"github.com/runopsio/hoop/gateway/plugin"
-	"github.com/runopsio/hoop/gateway/transport/plugins/dlp"
+	pluginsrbac "github.com/runopsio/hoop/gateway/transport/plugins/access_control"
+	pluginsaudit "github.com/runopsio/hoop/gateway/transport/plugins/audit"
+	pluginsdlp "github.com/runopsio/hoop/gateway/transport/plugins/dlp"
```
sandromello: If in your next Pull Request the plugin tries to use any function of `gateway/connection`, it will return a circular dependency error. I wonder if it make more sense to shift this names to common instead.

danielbonilha: I think we can move when there is a need for it. Make sense?

gateway/transport/plugins/access_control/access_control.go
danielbonilha: WIll change this.

PR #86, opened 2022-12-14
added access control to admins only for some endpoints

danielbonilha:
Added access control to the following endpoints:

** Available to admins only: **
- GET /users
- GET /users/{id}
- PUT /users/{id}
- POST /users

- POST /connections
- PUT /connections/{id}

- POST /agents
- GET /agents

- POST /plugins
- PUT /plugins

- GET /sessions
- GET /sessions/{id}

** Available to non-admins: **
- GET /connections
- GET /connections/{id}

- GET /reviews
- GET /review/{id}
- PUT /reviews

- GET /plugins
- GET /plugins/{name}

### DIFF
--- a/gateway/api/middleware.go
+++ b/gateway/api/middleware.go
@@ -49,7 +49,7 @@ func (api *Api) AdminOnly(c *gin.Context) {
 	context := ctx.(*user.Context)
 
 	if !context.User.IsAdmin() {
-		c.AbortWithStatus(403)
+		c.AbortWithStatus(401)
 		return
 	}
 


gateway/api/middleware.go
```diff
@@ -43,6 +44,18 @@ func (api *Api) validateClaims(c *gin.Context) (string, error) {
 	return api.IDProvider.VerifyAccessToken(tokenParts[1])
 }
 
+func (api *Api) AdminOnly(c *gin.Context) {
+	ctx, _ := c.Get("context")
+	context := ctx.(*user.Context)
+
+	if !context.User.IsAdmin() {
+		c.AbortWithStatus(401)
```
sandromello, line 52: This seems authorization, so it makes sense to return 403 (forbidden). What do you think? It may confuse clients with authentication issues (401).

gateway/api/server.go
```diff
@@ -72,30 +72,80 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
...
+		api.AdminOnly,
+		api.UserHandler.FindAll)
+	route.GET("/users/:id",
+		api.Authenticate,
+		api.AdminOnly,
+		api.UserHandler.FindOne)
+	route.PUT("/users/:id",
+		api.Authenticate,
+		api.AdminOnly,
+		api.UserHandler.Put)
```
sandromello, line 86: So, I think it fix the issue below right?
> [enjoei] The user did an escalation of privileges changing his own user as admin (need investigation)
Ref: https://3.basecamp.com/5385186/buckets/28171754/messages/5632936861

gateway/api/middleware.go
danielbonilha: Yes, can confuse the client, however, I like to try to reduce the information we "give" to the user, or to a potential attacker.

When we send a 401, a potential attacker does not differentiate between invalid token and lack of authorization. For a 403, attacker knows the token is good. 

Not sure if the 401 x 403 to the user makes any difference. 

I am OK changing to 403, though.

gateway/api/server.go
danielbonilha: It fixes, indeed.

gateway/api/middleware.go
danielbonilha: Just did.

sandromello: Yeah, it makes sense. I read about this a while ago.
I think we could left 401 them.

PR #88, opened 2022-12-22
added magic bell notifications for review

danielbonilha:
- Added notification service (magic bell)
- If MAGIC_BELL_API_KEY or MAGIC_BELL_API_SECRET is null, then gracefully skip the notification
- Send email to all users that contain one of the groups of the review

### DIFF
--- a/gateway/notification/magicbell.go
+++ b/gateway/notification/magicbell.go
@@ -27,7 +27,7 @@ func NewMagicBell() *MagicBell {
 }
 
 func (m *MagicBell) Send(notification Notification) {
-	if m.apiKey != "" && m.apiSecret != "" {
+	if m.apiKey != "" && m.apiSecret != "" && len(notification.Recipients) > 0 {
 		url := "https://api.magicbell.com/notifications"
 		req, err := http.NewRequest(http.MethodPost, url, buildPayload(notification))
 		if err != nil {
--- a/gateway/user/storage.go
+++ b/gateway/user/storage.go
@@ -137,11 +137,12 @@ func (s *Storage) FindByGroups(context *Context, groups []string) ([]User, error
 		return make([]User, 0), nil
 	}
 
-	var payload = `{:query {
-		:find [(pull ?user [*])] 
-		:where [[?user :user/org "` + context.Org.Id + `"]
-		` + buildGroupsClause(groups) + `
-		]}}`
+	var payload = fmt.Sprintf(`{:query {
+		:find [(pull ?user [*])]
+	  	:in [org [g ...]]
+    	:where [[?user :user/org org]
+				[?user :user/groups g]]}
+		:in-args ["%s" %q]}}`, context.Org.Id, groups)
 
 	b, err := s.Query([]byte(payload))
 	if err != nil {
@@ -154,16 +155,6 @@ func (s *Storage) FindByGroups(context *Context, groups []string) ([]User, error
 	}
 
 	return users, nil
-}
-
-func buildGroupsClause(groups []string) string {
-	s := "(or "
-	for _, g := range groups {
-		s += fmt.Sprintf(`[?user :user/groups "%s"]`, g)
-	}
-
-	s += ")"
-	return s
 }
 
 func (s *Storage) getOrg(orgId string) (*Org, error) {


gateway/notification/magicbell.go
```diff
@@ -0,0 +1,69 @@
...
+		apiSecret: os.Getenv("MAGIC_BELL_API_SECRET"),
+	}
+}
+
+func (m *MagicBell) Send(notification Notification) {
+	if m.apiKey != "" && m.apiSecret != "" {
+		url := "https://api.magicbell.com/notifications"
+		req, err := http.NewRequest(http.MethodPost, url, buildPayload(notification))
+		if err != nil {
+			fmt.Printf("failed to send magic bell notification: %s", err.Error())
```
sandromello: This error happens on building the requesting not when performing the request. This is done by the `h.client.Do(req)`. You could change the error to:
`Failed building http request, err=%v`

You could use the placeholder `%v` and use only the variable `err` in the second argument.

Need to change from package `fmt` to `log`

gateway/notification/magicbell.go
```diff
@@ -0,0 +1,69 @@
...
+	MagicBell struct {
+		client    http.Client
+		apiKey    string
+		apiSecret string
+	}
+)
+
+func NewMagicBell() *MagicBell {
+	return &MagicBell{
+		client:    http.Client{},
```
sandromello: I think it may be a good idea to add timeout to requests for these services:

```go
(...)
client: http.Client{Timeout: 2 * time.Second}
(...)
```

gateway/user/storage.go
```diff
@@ -131,6 +132,40 @@ func (s *Storage) GetOrgByName(name string) (*Org, error) {
...
+func (s *Storage) FindByGroups(context *Context, groups []string) ([]User, error) {
+	if len(groups) == 0 {
+		return make([]User, 0), nil
+	}
+
+	var payload = `{:query {
+		:find [(pull ?user [*])] 
+		:where [[?user :user/org "` + context.Org.Id + `"]
+		` + buildGroupsClause(groups) + `
+		]}}`
```
sandromello: Two points here:

- This is something really hard to read when constructing queries with strings
- Constructing queries without using the `in-args` could be subjected to query injections.

> I know that the ID of an org comes from the database and it's a safe input, but I always need to think on that when I see queries constructed by joining strings.

Is it possible to build a more friendly query here?
We could fetch every users from an org and match the groups from that list in GO. It's not optimal but we could improve it later.

This query also works to find users by groups:
```edn
{:query {
    :find [(pull ?user [*])]
    :in [[grp ...]]
    :where [[?user :user/groups grp]]
}
:in-args [["dev" "test"]]}
```

See https://docs.xtdb.com/language-reference/1.22.1/datalog-queries/#_scalar_binding

It's possible to use some clojure functions also (need to find which of them are exposed)

gateway/notification/magicbell.go
```diff
@@ -0,0 +1,69 @@
...
+			return
+		}
+
+		req.Header.Set("content-type", "application/json")
+		req.Header.Set("X-MAGICBELL-API-KEY", m.apiKey)
+		req.Header.Set("X-MAGICBELL-API-SECRETY", m.apiSecret)
+
+		resp, err := m.client.Do(req)
+		if err != nil {
+			fmt.Printf("failed to send magic bell notification: %s", err.Error())
```
sandromello: Use the `log` package here instead of `fmt`

gateway/user/storage.go
danielbonilha: I tried to make the in-args work, but I kept receiving a java exception, without further explanation of the error. Maybe because of the "(or ..." function.

I tried in many ways, but didnt make it work. I know it is subject to injection, but it would be a very hard thing to do, and I believe only insiders could do that.

The attacker would need to know that we use clojure, that we use xtdb, need to know the structure of the query, as a lot of more things to inject something meaneaninful. 

Maybe neither of us could think about a string that would harm our database.

So I left this way for the sake of being fast.

gateway/notification/magicbell.go
danielbonilha: Done.

gateway/notification/magicbell.go
danielbonilha: Done.

gateway/notification/magicbell.go
danielbonilha: Done.

gateway/user/storage.go
sandromello: This is not something hard to know, download the quick start and you can know what is the database engine that we're using and xtdb is clojure/java.
We should not count with hope when writing code, if there's a security hole in our code we should avoid it.

Another point that I mentioned is about how hard is reading the code when joining strings. It's hard for me to make sense of this query because I need to join strings in my head to know which query is being executed.

I've created a task for us to deal with this later on: https://3.basecamp.com/5385186/buckets/28171754/todos/5670230109

danielbonilha: Can you provide some example of injection that could be done that could do some harm?
Can you provide a new way to doing the query without being so hard to read?
I can learn from that.

sandromello: So, this is more about being defensive. I really don't have all the knowledge to say how things could be exploited, but every time I see a query being composed joining strings I need to think if this input came from a safe source. It's more hard to review the code having to think about that.

So, not using a string to create inputs which could be processed in a complex parser it's coding on defensive. You don't know how anyone could exploit that, but you know that's possible. In the future we could allow more functions to be executed in the query, or a new feature or vulnerability is discovered. We don't know what could happen, but this opens a breach for an exploit in the present or in the future.

So, the `in-args` exists to be a safe place for dealing with arguments. We have the same thing with other drivers to, like `sql`:

```go
db.Query("SELECT * FROM album WHERE artist = ?", artist)
```

So it could be tempting to do something like:

```go
db.Query(fmt.Sprintf("SELECT * FROM album WHERE artist = %s", artist))
```

But there's a reason for the driver to let you pass inputs in such way, the driver will save you from the troubles of making sure this input is safe, otherwise we'll be needing do this from our side.

It's like driving your car maintaining a secure distance from the one ahead of you. If the car ahead makes something unexpected you will be able to react. It's [defensive programming](https://en.wikipedia.org/wiki/Defensive_programming)

Regarding writing the datalog queries, a more pleasant way of reading it is using placeholders, example:

```go
querystr := fmt.Sprintf(`
{:query {
    :find [(pull ?user [*])]
    :in [org [grp ...]]
    :where [
        [?user :user/org org]
        [?user :user/groups grp]
        ]
}
:in-args ["%s" ["%s" "%s"]]}`,
"a0ce0690-1518-4a4d-ad4b-7f90c494807b", "admin", "test")
```

I could get this query and paste on my postman or ident it to test in another place and it's more easier to read the query also.

But if I write something like this:

```go
org := "myorg-id"
querystr := "
{:query {
    :find [(pull ?user [*])]
    :in [org [grp ...]]
    :where [
        [?user :user/org " + org + "]
        [?user :user/groups" + myCustomFn(groups) +"]
        ]
}}"
```

It's really hard to understand the outcome of the query, I need to read a function to understand what's being executed in the xtdb.

PR #92, opened 2022-12-27
added list groups endpoint

danielbonilha:
- GET /users/groups

Response 200 - ok
`
[
  "admin",
  "devops",
  "sre",
...
]
`

### DIFF
--- a/gateway/user/api.go
+++ b/gateway/user/api.go
@@ -2,6 +2,7 @@ package user
 
 import (
 	pb "github.com/runopsio/hoop/common/proto"
+	"log"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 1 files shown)


gateway/user/api.go
```diff
@@ -133,3 +134,16 @@ func (a *Handler) Userinfo(c *gin.Context) {
...
 	c.JSON(http.StatusOK, context.User)
 }
+
+func (a *Handler) UsersGroups(c *gin.Context) {
+	ctx, _ := c.Get("context")
+	context := ctx.(*Context)
+
+	groups, err := a.Service.ListAllGroups(context)
+	if err != nil {
+		c.AbortWithStatus(http.StatusInternalServerError)
```
sandromello: Add a log statement here in case of errors

gateway/api/server.go
```diff
@@ -167,7 +170,7 @@ func (api *Api) CreateTrialEntities() error {
 		Name:   "hooper",
 		Email:  "tester@hoop.dev",
 		Status: "active",
-		Groups: []string{"admin"},
+		Groups: []string{"admin", "sre", "dba", "security", "devops", "support", "engineering"},
```
sandromello, line 173: This user is admin already, why does the test user need those groups?

PR #93, opened 2022-12-30
Feat/jit plugin

danielbonilha:
- Added jit (just-in-time) plugin
- Added jit endpoints:
 - GET /plugins/jit -> list all jits
 - GET /plugins/jit/{session} -> get single jit
 - PUT /plugins/jit/{session} -> approve / reject connection

**More info:**
- if JIT plugin is enabled, user will be as asked for how many minutes he wants to connect
- The JIT plugin configuration accepts a list of groups that needs to approve
- If a connection is marked with JIT plugin, then no exec will be allowed
- If a connection is marked with review plugin only, then no connect will be allowed
- If a connection is marked with both JIT and review plugins, than all is allowed
- The JIT review is sync, the same way it is for the review. If user drops the the session, than a new review is required.

### DIFF
--- a/gateway/transport/plugins/jit/jit.go
+++ b/gateway/transport/plugins/jit/jit.go
@@ -83,7 +83,7 @@ func (r *jitPlugin) OnStartup(config plugin.Config) error {
 
 func (r *jitPlugin) OnConnect(config plugin.Config) error {
 	if config.Org == "" || config.SessionId == "" {
-		return fmt.Errorf("failed processing review plugin, missing org_id and session_id params")
+		return fmt.Errorf("failed processing jit plugin, missing org_id and session_id params")
 	}
 
 	return nil
@@ -108,7 +108,7 @@ func (r *jitPlugin) OnReceive(pluginConfig plugin.Config, config []string, pkt *
 		if requestTime != nil {
 			timeInt, err = strconv.Atoi(string(requestTime))
 			if err != nil {
-				return errors.New("invalid just-in-time plugin configuration")
+				return errors.New("invalid timeout configuration")
 			}
 		}
 


client/cmd/connect.go
```diff
@@ -141,11 +142,127 @@ func (c *connect) processPacket(pkt *pb.Packet, config *Config) {
...
+		scanner := bufio.NewScanner(os.Stdin)
+		for scanner.Scan() {
+			s := scanner.Text()
+			input += s
+			break
+		}
+
+		if _, err := strconv.Atoi(input); err != nil {
+			c.processGracefulExit(fmt.Errorf(`invalid number %q`, input))
+			return
```
sandromello: Don't need to return if you're already gracefully exiting (this function with call os.Exit)

client/cmd/connect.go
```diff
@@ -141,11 +142,127 @@ func (c *connect) processPacket(pkt *pb.Packet, config *Config) {
...
+		}
+		c.processGracefulExit(errors.New("user cancelled the action"))
+	case pb.PacketClientGatewayConnectRejectType:
+		c.processGracefulExit(errors.New("connection rejected. Sorry"))
 	case pb.PacketClientAgentConnectErrType:
 		sessionID := pkt.Spec[pb.SpecGatewaySessionID]
 		errMsg := fmt.Errorf("session=%s - failed connecting with gateway, err=%v",
 			string(sessionID), string(pkt.GetPayload()))
 		c.processGracefulExit(errMsg)
```
sandromello, line 227: TIP: you don't need a type here, you could send a `PacketClientAgentConnectErrType` containing the error in the payload. I think it's ok to leave this way, but we could starting refactoring some flows to allow a more generic handling of actions

client/cmd/connect.go
```diff
@@ -141,11 +142,127 @@ func (c *connect) processPacket(pkt *pb.Packet, config *Config) {
...
+
+		if input == "" {
+			input = "y"
+		}
+
+		input = input[0:1]
+
+		if input == "y" {
+			c.waitingJit.Type = string(pb.PacketClientGatewayConnectType)
+			_ = c.client.Send(c.waitingJit)
```
sandromello, line 216: The spec containing the client args is missing

client/cmd/connect.go
```diff
@@ -141,11 +142,127 @@ func (c *connect) processPacket(pkt *pb.Packet, config *Config) {
...
+
+		if input == "" {
+			input = "y"
+		}
+
+		input = input[0:1]
+
+		if input == "y" {
+			pkt.Type = string(pb.PacketClientGatewayConnectType)
+			_ = c.client.Send(pkt)
```
sandromello, line 259: The spec containing the client args is missing

gateway/transport/plugins/jit/jit.go
```diff
@@ -0,0 +1,189 @@
...
+	if config.Org == "" || config.SessionId == "" {
+		return fmt.Errorf("failed processing review plugin, missing org_id and session_id params")
+	}
+
+	return nil
+}
+
+func (r *jitPlugin) OnReceive(pluginConfig plugin.Config, config []string, pkt *pb.Packet) error {
+	switch pb.PacketType(pkt.GetType()) {
+	case pb.PacketClientGatewayConnectType:
```
sandromello, line 94: This plugin phase is used when exchanging packets between the client/agent after the connect phase. Why aren't you using the OnConnect phase instead?

gateway/transport/plugins/jit/jit.go
```diff
@@ -0,0 +1,189 @@
...
+			return err
+		}
+
+		var timeInt int
+		requestTime := pkt.Spec[pb.SpecJitTimeout]
+		if requestTime != nil {
+			timeInt, err = strconv.Atoi(string(requestTime))
+			if err != nil {
+				return errors.New("invalid just-in-time plugin configuration")
+			}
```
sandromello: TIP: we could have a default time here instead of returning error. Or dealing with this validation in the client.

Another point here: why is it returning a "invalid plugin configuration"?  The input is coming from the user right?

gateway/transport/client.go
```diff
@@ -256,6 +291,39 @@ func (s *Server) processClientConnect(pkt *pb.Packet, client *client.Client, con
...
+				Type:    string(pb.PacketClientGatewayConnectRequestTimeoutType),
+				Spec:    spec,
+				Payload: pkt.Payload,
+			})
+
+			return nil
+		}
+		ctx, _ := context.WithTimeout(client.Context, time.Minute*jit.Time)
+		client.Context = ctx
+	}
```
sandromello, line 325: This is very specific code of JIT, why isn't in the plugin phases?

gateway/transport/client.go
```diff
@@ -191,11 +220,17 @@ func (s *Server) listenClientMessages(
...
 
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
+		case <-c.Context.Done():
+			_ = stream.Send(&pb.Packet{
+				Type: string(pb.PacketClientConnectTimeoutType),
+			})
+			return nil
```
sandromello, line 233: Why do you need to implement a timeout for this flow in the gateway? You can do that in the client which will be much more simpler.

client/cmd/connect.go
danielbonilha: The spec is already in place in the c.waitingJib packet.

client/cmd/connect.go
danielbonilha: Done.

client/cmd/connect.go
danielbonilha: It is not missing. It is already in place in pkt variable. We are just forwarding a packet that came from the client

gateway/transport/plugins/jit/jit.go
danielbonilha: For a number of reasons. The main reason is that the onConnect is called way before we know details of the packet. It is invoked in the loadConnectPlugins(), that is called during subscribeClient().

This means that we cannot hold the connection in memory at this point yet, because it was not even created. This is an architectural problem, that already appeared during the review plugin development.

gateway/transport/plugins/jit/jit.go
danielbonilha: This is just a defensive programming here because we do some string->int conversion.

The real validation happens on the client side, where the user inputs the timeout. However, since the client cannot be trusted, I do a double-check validation.

gateway/transport/client.go
danielbonilha: Because this flow is supposed to run after the onConnect, and before the onReceive. it is checking the existence of the plugin, so it can take actions on it.

gateway/transport/client.go
danielbonilha: The client cannot be trusted. The gateway is the one holding the plugin configuration, and therefore it is the gateway responsability to guarantee that the connection is broken.

gateway/transport/plugins/jit/jit.go
sandromello: Ok, got it. The "connect gateway" makes me confuse sometimes. As you pointed out in another PR, this flow could be inside the grpc connect which will make things more clear.

gateway/transport/plugins/jit/jit.go
sandromello: So this is input is coming from the user. I think it makes sense to change it to something else like:
"invalid timeout input". Otherwise we may confuse this with a invalid plugin configuration

gateway/transport/client.go
sandromello: There's any trust issue on relying a default timeout for the client. I just think that puts a lot of unnecessary logic in the gateway. But it's ok to go with this implementation.

PR #101, opened 2023-01-23
Feat/analytics

danielbonilha:
- Added analytics to gateway during API and gRPC usage
- Send to segment.io (that forwards to June)
- Identify the user and the company during signup (analytics.identify and analytics.track)
- Track all user activity in the gateway API (generic middleware that tracks the requests but not the responses)
- Track all user activity during gRPC interactions (exec and connections)

Note:
The login was deliberately skept because there will probably be a refactor to remove the email from the flow, so it will be done in this PR.

### DIFF
--- a/gateway/transport/client.go
+++ b/gateway/transport/client.go
@@ -223,6 +223,9 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
 		"hostname":        hostname,
 		"machine-id":      machineId,
 		"kernel-version":  kernelVersion,
+		"client-version":  c.Version,
+		"go-version":      c.GoVersion,
+		"platform":        c.Platform,
 	})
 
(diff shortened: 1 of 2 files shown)


gateway/transport/client.go
```diff
@@ -216,6 +216,15 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
 	s.handleGracefulShutdown(c)
 
+	s.Analytics.Track(context.User.Id, clientVerb, map[string]any{
+		"sessionID":       sessionID,
+		"connection-name": connectionName,
+		"connection-type": conn.Type,
+		"hostname":        hostname,
+		"machine-id":      machineId,
+		"kernel-version":  kernelVersion,
+	})
```
sandromello: Hey, we can put more thing here like:
- c.Version (client version)
- c.GoVersion (client go version)
- c.Compiler (client go compiler)
- c.Platform (client os platform)
- c.Verb
- c.UserID

PR #104, opened 2023-01-24
sentry config

danielbonilha:
- Added sentry gin configuration as per docs.
It handles all panics.

https://docs.sentry.io/platforms/go/guides/gin/

### DIFF
--- a/gateway/agent/api.go
+++ b/gateway/agent/api.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"github.com/getsentry/sentry-go"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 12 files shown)


gateway/api/server.go
```diff
@@ -42,6 +44,24 @@ func (api *Api) StartAPI() {
 	if os.Getenv("PORT") == "" {
 		os.Setenv("PORT", "8009")
 	}
+
+	sentryInit := false
+	dsn := os.Getenv("SENTRY_DSN")
```
sandromello, line 49: TIP: You can do this is a single line:
```go
if dsn := os.Getenv("SENTRY_DSN"); dsn != "" {
   <your coder>
}
```

PR #107, opened 2023-01-27
Feat/add monitoring

sandromello:
- Add sentry to client/agent
- Add [pyroscope for continuous profiling](https://pyroscope.io/)

## Sentry

The agent will receive a DSN in runtime to capture errors, we could later on change it configuring the env var `AGENT_SENTRY_DSN`

The client has a hardcoded DSN, it reports errors in synchronous due to the characteristic of how the client works.

## Profiling

This PR introduces profiling of agent/gateway, this is an experimental change to add more observability to these components. Having profiling will help us to diagnose problems more precise and understand where to look for issues.

Pyroscope has features such as:

- Mutex monitoring (we had a recent problem with that!)
- Comparing / Diffing profiles to see if we introduce bugs
- Memory/CPU Profiling explain how much each object is allocating

The cloud solution is free for 100GB / monthly (14 days retention), after that it starts at U$$500/month for 1TB of ingestion.

I've tested the opensource solution and deployed on Kubernetes, it worked very well, we could host an instance if the cost is too high.

Gateway Profiling:

image

Comparing Profiles:
image

Agent Profiling:
image

## Introduced environments

- AGENT_SENTRY_DSN
- PYROSCOPE_AUTH_TOKEN
- PYROSCOPE_INGEST_URL

If these environment variables are not set, the gateway/agent will not start the profiler or sentry

### DIFF
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/getsentry/sentry-go"
 	"github.com/runopsio/hoop/common/monitoring"
 	pb "github.com/runopsio/hoop/common/proto"
 	"github.com/spf13/cobra"
@@ -35,7 +34,6 @@ var loginCmd = &cobra.Command{
 	PreRun: monitoring.SentryPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
 		if err := doLogin(args); err != nil {
-			sentry.CaptureException(fmt.Errorf("login - failed to authenticate, err=%v", err))
 			fmt.Printf("Failed to login, err=%v\n", err)
 			os.Exit(1)
 		}


client/cmd/login.go
```diff
@@ -27,11 +29,13 @@ const (
...
-	Use:   "login",
-	Short: "Authenticate at Hoop",
-	Long:  `Login to gain access to hoop usage.`,
+	Use:    "login",
+	Short:  "Authenticate at Hoop",
+	Long:   `Login to gain access to hoop usage.`,
+	PreRun: monitoring.SentryPreRun,
 	Run: func(cmd *cobra.Command, args []string) {
 		if err := doLogin(args); err != nil {
+			sentry.CaptureException(fmt.Errorf("login - failed to authenticate, err=%v", err))
```
danielbonilha: Maybe this is noise we should not capture?

gateway/transport/agent.go
```diff
@@ -126,7 +148,7 @@ func (s *Server) listenAgentMessages(config plugin.Config, ag *agent.Agent, stre
 			log.Printf("received error from agent, err=%v", err)
 			return err
 		}
-		if pkt.Type == pbgateway.KeepAlive {
+		if pkt.Type == pbgateway.KeepAlive || pkt.Type == "KeepAlive" {
```
danielbonilha, line 151: I didnt get this.

sandromello: Old clients, rs

client/cmd/login.go
sandromello: Yeah, I'm removing it

PR #128, opened 2023-02-28
added groups handling from IDP

danielbonilha:
**Fixes on authentication/authorization flow:**

- Now we fetch the sub from the access token during signup. This will prevent issues regarding different subs for access and id tokens. No more specific if/else per IDP is necessary
- ID tokens may or may not receive the custom claim "https://app.hoop.dev/org". If the claim is present, then it will take precedence over the email-parsing logic
- ID tokens may or may not receive the custom claim "https://app.hoop.dev/groups". If it is present, than hoop will consider this as source of truth. If not present, than hoop will continue the way it is
- groups that **contains** _"admin"_ in its name will be considered as "admin". This is because some clients will use custom groups with hoop, and "admin" is a generic word that could conflict with other groups. So some customers will use the pattern **_{org}\_admin_** to represent a group at their provider that belongs to hoop and is an admin

### DIFF
--- a/gateway/security/service.go
+++ b/gateway/security/service.go
@@ -114,10 +114,7 @@ func (s *Service) Callback(state, code string) string {
 
 	groupsClaim, _ := idTokenClaims[pb.CustomClaimGroups].([]any)
 	if len(groupsClaim) > 0 {
-		groups := make([]string, 0)
-		for _, g := range groupsClaim {
-			groups = append(groups, g.(string))
-		}
+		groups := mapGroupsToString(groupsClaim)
 
 		context.User.Groups = groups
 		if err := s.UserService.Persist(context.User); err != nil {
@@ -187,9 +184,7 @@ func (s *Service) signup(context *user.Context, sub string, idTokenClaims map[st
 		groups := make([]string, 0)
 		groupsClaim, _ := idTokenClaims[pb.CustomClaimGroups].([]any)
 		if len(groupsClaim) > 0 {
-			for _, g := range groupsClaim {
-				groups = append(groups, g.(string))
-			}
+			groups = mapGroupsToString(groupsClaim)
 		}
 		status := user.StatusReviewing
 
@@ -243,4 +238,16 @@ func (s *Service) signup(context *user.Context, sub string, idTokenClaims map[st
 func (s *Service) loginOutcome(login *login, outcome outcomeType) {
 	login.Outcome = outcome
 	s.Storage.PersistLogin(login)
+}
+
+func mapGroupsToString(groupsClaim []any) []string {
+	groups := make([]string, 0)
+	for _, g := range groupsClaim {
+		groupName, _ := g.(string)
+		if groupName == "" {
+			continue
+		}
+		groups = append(groups, groupName)
+	}
+	return groups
 }
--- a/gateway/user/service.go
+++ b/gateway/user/service.go
@@ -162,7 +162,7 @@ func isPublicDomain(domain string) bool {
 }
 
 func (user *User) IsAdmin() bool {
-	return pb.IsInList(GroupAdmin, user.Groups) || pb.StartsWithInList(GroupAdmin, user.Groups)
+	return pb.IsInList(GroupAdmin, user.Groups)
 }
 
 func isInStatus(status StatusType) bool {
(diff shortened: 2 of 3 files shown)


gateway/security/service.go
```diff
@@ -123,6 +130,12 @@ func (s *Service) Callback(state, code string) string {
 		return login.Redirect + "?error=pending_review"
 	}
 
+	groups, ok := idTokenClaims["groups"].([]string)
+	if ok {
+		context.User.Groups = groups
+		s.UserService.Persist(context.User)
```
sandromello: I think you must return an error in case this fails, right?

gateway/security/service.go
```diff
@@ -191,20 +204,24 @@ func (s *Service) signup(context *user.Context, sub string, profile map[string]a
...
 			status = user.StatusActive
-			groups = append(groups,
-				user.GroupAdmin,
-				user.GroupSecurity,
-				user.GroupSRE,
-				user.GroupDBA,
-				user.GroupDevops,
-				user.GroupSupport,
-				user.GroupEngineering)
+			if !existClaimGroup {
```
sandromello: This will only match if the key doesn't exists or the type cast doesn't happen. However, if it has empty entries it will not work. You could easily do this which will catch both conditions: key not found or group with empty groups

```go
groups, _ := profile["groups"].([]string)
if len(groups) == 0 {
  (...)
}
```

> The `_` is for safe keeping, It will not panic if the type is not a `[]string` or the key doesn't exists in the map

gateway/security/service.go
danielbonilha: Done

gateway/security/service.go
danielbonilha: I had to change the overall logic here because the cast from []interface{} to []string were raising an error:
`interface conversation: interface{} is []interface{}, not []string`

So i just refactored a little bit to cast each individual element.

However I changed the logic to use the len(groups) instead of the "ok" variable.

gateway/security/service.go
```diff
@@ -182,6 +185,12 @@ func (s *Service) signup(context *user.Context, sub string, profile map[string]a
 
 	if context.User == nil {
 		groups := make([]string, 0)
+		groupsClaim, _ := idTokenClaims[pb.CustomClaimGroups].([]any)
+		if len(groupsClaim) > 0 {
+			for _, g := range groupsClaim {
+				groups = append(groups, g.(string))
```
sandromello: I think it's good to put a condition to bypass empty groups in a list:

```go
groupName, _ := g.(string)
if groupName == "" {
  continue
}
```

gateway/user/service.go
```diff
@@ -162,7 +162,7 @@ func isPublicDomain(domain string) bool {
 }
 
 func (user *User) IsAdmin() bool {
-	return pb.IsInList(GroupAdmin, user.Groups)
+	return pb.IsInList(GroupAdmin, user.Groups) || pb.StartsWithInList(GroupAdmin, user.Groups)
```
sandromello: So, in this case:
- sre-admin - is admin
- dbaadmin-foo - is admin
- devadmin - is admin

If an administrator isn't aware of that, he'll be allowing an escalation of privileges by mistake.
I think it's better to define a static name for defining a user as admin that will not overlap with any group that they have, like: `hoop-admin`, `hoopadmin`. What do you think?

gateway/security/service.go
```diff
@@ -123,6 +112,20 @@ func (s *Service) Callback(state, code string) string {
 		return login.Redirect + "?error=pending_review"
 	}
 
+	groupsClaim, _ := idTokenClaims[pb.CustomClaimGroups].([]any)
+	if len(groupsClaim) > 0 {
+		groups := make([]string, 0)
+		for _, g := range groupsClaim {
+			groups = append(groups, g.(string))
```
sandromello: I think it's good to put a condition to bypass empty groups in a list.

gateway/user/service.go
danielbonilha: Yes, sounds good. I will remove this particular operation from this PR now because it is out of the scope. Since this needs more conversation, lets do it next.

gateway/security/service.go
danielbonilha: Done.

gateway/security/service.go
danielbonilha: Done.

PR #132, opened 2023-03-06
added idp custom scopes as optional env var

danielbonilha:
- Added IDP_CUSTOM_SCOPES as optional scope to be passed to the scope list.
- IDP_CUSTOM_SCOPES is a comma separated string. Ex: IDP_CUSTOM_SCOPES="my_scope,another_scope"

Note:
The default scope list currently is [openid, email, profile]. If the variable is set, then the items wll be appended:
[openid,email,profile,my_scope,another_scope]

### DIFF
--- a/gateway/security/idp/authenticator.go
+++ b/gateway/security/idp/authenticator.go
@@ -139,7 +139,9 @@ func NewProvider(profile string) *Provider {
 
 func addCustomScopes(scopes []string, customScope string) []string {
 	custom := strings.Split(customScope, ",")
-	scopes = append(scopes, custom...)
+	for _, c := range custom {
+		scopes = append(scopes, strings.Trim(c, " "))
+	}
 	return scopes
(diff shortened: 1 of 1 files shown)


gateway/security/idp/authenticator.go
```diff
@@ -130,6 +137,12 @@ func NewProvider(profile string) *Provider {
 	return provider
 }
 
+func addCustomScopes(scopes []string, customScope string) []string {
+	custom := strings.Split(customScope, ",")
+	scopes = append(scopes, custom...)
+	return scopes
```
sandromello: I think would be good to add a trim to each scope, this will avoid issues with:
```
# space base scopes
IDP_CUSTOM_SCOPES="scope01, scope02"
```

danielbonilha: Done.

PR #135, opened 2023-03-08
added connection exec

danielbonilha:
- Added new endpoint POST /api/connection/{name}/exec
- that accepts a script
```
{
  "script": "ls -la"
}
```
and return accordingly (based on runbooks)

### DIFF
--- a/gateway/connection/api.go
+++ b/gateway/connection/api.go
@@ -24,7 +24,7 @@ type (
 		Persist(httpMethod string, context *user.Context, c *Connection) (int64, error)
 		FindAll(context *user.Context) ([]BaseConnection, error)
 		FindOne(context *user.Context, name string) (*Connection, error)
-		ProcessClientExec(inputPayload []byte, encodedEnvVars []byte, client pb.ClientTransport) (int, error)
+		ProcessClientExec(inputPayload []byte, client pb.ClientTransport) (int, error)
 	}
 )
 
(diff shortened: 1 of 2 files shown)


gateway/connection/api.go
```diff
@@ -120,3 +128,91 @@ func (a *Handler) Put(c *gin.Context) {
...
+	go func() {
+		defer close(clientResp)
+		defer client.Close()
+		var encEnvVars []byte
+		if len(conn.Secret) > 0 {
+			encEnvVars, err = pb.GobEncode(conn.Secret)
+			if err != nil {
+				clientResp <- ExecResponse{err, nilExitCode}
+				return
+			}
```
sandromello: THis is not necessary, this will be fetch when connecting to the gRPC gateway.
The runbook has something similar, but it's to propagate environment variables from a runbook definition.

gateway/connection/api.go
```diff
@@ -120,3 +128,91 @@ func (a *Handler) Put(c *gin.Context) {
...
+	statusCode := http.StatusOK
+	if req.Redirect {
+		statusCode = http.StatusFound
+	}
+	select {
+	case resp := <-clientResp:
+		switch resp.ExitCode {
+		case nilExitCode:
+			// means the gRPC client returned an error in the client flow
+			c.JSON(statusCode, &ExecErrResponse{
```
sandromello: This should be a bad request error

gateway/connection/api.go
```diff
@@ -120,3 +128,91 @@ func (a *Handler) Put(c *gin.Context) {
...
+			c.JSON(http.StatusBadRequest, &ExecErrResponse{
+				SessionID: &sessionID,
+				Message:   fmt.Sprintf("%v", resp.Err),
+				ExitCode:  &resp.ExitCode})
+		}
+	case <-time.After(time.Second * 50):
+		// closing the client will force the goroutine to end
+		// and the result will return async
+		client.Close()
+		c.JSON(statusCode, gin.H{"session_id": sessionID, "exit_code": nil})
```
sandromello: This should be a 202 status, because the result will be returned async

PR #136, opened 2023-03-13
Feat/userinfo external validation

danielbonilha:
If there is a variable IDP_TOKEN_EXTERNAL_VALIDATION != ""

Then the application will validate the token at the userinfo endpoint. The request is being handled by the library.

### DIFF
--- a/gateway/security/idp/authenticator.go
+++ b/gateway/security/idp/authenticator.go
@@ -27,14 +27,13 @@ var invalidAuthErr = errors.New("invalid auth")
 
 type (
 	Provider struct {
-		Issuer                  string
-		Audience                string
-		ClientID                string
-		ClientSecret            string
-		Profile                 string
-		CustomScopes            string
-		ApiURL                  string
-		tokenExternalValidation string
+		Issuer       string
+		Audience     string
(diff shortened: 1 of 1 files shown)


gateway/security/idp/authenticator.go
```diff
@@ -25,33 +25,43 @@ const (
...
-	return a.Verify(a.Context, rawIDToken)
+	return p.Verify(p.Context, rawIDToken)
 }
 
-func (a *Provider) VerifyAccessToken(accessToken string) (string, error) {
-	token, err := jwt.Parse(accessToken, a.JWKS.Keyfunc)
+func (p *Provider) VerifyAccessToken(accessToken string) (string, error) {
+	if p.tokenExternalValidation != "" {
+		return p.UserInfoEndpoint(accessToken)
+	}
```
sandromello: Hey, instead of introducing another env variable which introduces more configuration options, why don't you verify the token format and apply this login based on this?
```
(...)
	if !isJWTBearerToken(accessToken) {
		return p.UserInfoEndpoint(accessToken)
	}
```
What do you think?

PR #148, opened 2023-03-22
Add zap library logging over log std library

sandromello:
Introduces zap as logging library to agent and gateway

- Add env `LOG_ENCODING=console|json` which allows controlling the encoding the format of output
- Add env `LOG_LEVEL=info|warn|error|debug` to increase/decrease the level of logging
- By default adds sampling of events in gateway (debug messages)
- Add context logging to add by default information about the user logged in (org,user-id,etc)
- Add logging gin routes with zap when `GIN_MODE=debug`
- Add recovery gin panic with zap

### DIFF
--- a/gateway/user/service.go
+++ b/gateway/user/service.go
@@ -3,9 +3,7 @@ package user
 import (
 	"strings"
 
-	"github.com/gin-gonic/gin"
 	pb "github.com/runopsio/hoop/common/proto"
-	"go.uber.org/zap"
 )
 
 type (
@@ -182,25 +180,4 @@ func isInStatus(status StatusType) bool {
 		}
 	}
 	return false
-}
-
-// ContextLogger do a best effort to get the context logger,
-// if it fail to retrieve, returns a noop logger
-func ContextLogger(c *gin.Context) *zap.SugaredLogger {
-	obj, _ := c.Get(ContextLoggerKey)
-	if logger := obj.(*zap.SugaredLogger); logger != nil {
-		return logger
-	}
-	return zap.NewNop().Sugar()
-}
-
-// ContextUser do a best effort to get the user context from the request
-// if it fail, it will return an empty one
-func ContextUser(c *gin.Context) *Context {
-	obj, _ := c.Get("context")
-	ctx := obj.(*Context)
-	if ctx == nil {
-		return &Context{}
-	}
-	return ctx
 }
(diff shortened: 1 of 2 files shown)


gateway/user/service.go
```diff
@@ -178,3 +183,24 @@ func isInStatus(status StatusType) bool {
 	}
 	return false
 }
+
+// ContextLogger do a best effort to get the context logger,
+// if it fail to retrieve, returns a noop logger
+func ContextLogger(c *gin.Context) *zap.SugaredLogger {
```
danielbonilha: I don't think we should inject webserver (c *gin.Context) logic into the user service. Thinking about layers, internal layers should not be aware of external layers.

Lets move this logic to the "api.go level", since this layer already handles/knows about Gin.

gateway/user/service.go
```diff
@@ -178,3 +183,24 @@ func isInStatus(status StatusType) bool {
...
+	return zap.NewNop().Sugar()
+}
+
+// ContextUser do a best effort to get the user context from the request
+// if it fail, it will return an empty one
+func ContextUser(c *gin.Context) *Context {
+	obj, _ := c.Get("context")
+	ctx := obj.(*Context)
+	if ctx == nil {
+		return &Context{}
```
danielbonilha: I think this scenario will never happen, because the auth layer will assure context will be there.

But if this happen, it can cause many NPE along the way, since we constantly use the context.User and context.Org in many places.

danielbonilha: I don't think we should inject webserver (c *gin.Context) logic into the user service. Thinking about layers, internal layers should not be aware of external layers.

Lets move this logic to the "api.go level", since this layer already handles/knows about Gin.

sandromello: What's NPE @danielbonilha ?

danielbonilha: Null pointer exeception

sandromello: But regarding your point about this never happen: I can't rely in the layer above because anything could happen in the future:

- Introduction of a bug that doesn't populate the org/user properly
- A bug in the gin library
- A misconfiguration in the gin routes

Thus, this function guarantees that we'll always have a non-panic context no matter what. The same thing with zap logger.

danielbonilha: This does not guarantee a non panic context because at some point we will try to get the User or the Org from the context, and it will fail anyhow. But again, I think this won't be a problem for us.

If this is caused by a bug, we should fix the bug, and not write code assuming there will be a bug.

PR #149, opened 2023-03-22
Enhanc/redacted fields count report

danielbonilha:
- Added DLP Count to sessions
- Added DLP Count to weekly report

**Note:**
Because of a limitation of xtdb (or limitation in our knowledge), the _order by_ clause was removed from the session list. Instead, it is being ordered DESC by startDate in-memory. This can impact pagination and should be reviewed later.

### DIFF
--- a/agent/dlp/gcp.go
+++ b/agent/dlp/gcp.go
@@ -128,7 +128,7 @@ func (c *client) DeidentifyContent(ctx context.Context, conf *deidentifyConfig, 
 		dataRowsBuffer := encodeToDataRow(responseTable)
 		return &Chunk{
 			index:                 chunkIndex,
-			transformationSummary: &pbdlp.TransformationSummary{},
+			transformationSummary: chunk.transformationSummary,
 			data:                  dataRowsBuffer}
 	}
 	chunk.data = bytes.NewBufferString(r.Item.GetValue())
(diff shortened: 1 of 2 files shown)


gateway/transport/plugins/audit/wal.go
```diff
@@ -219,3 +235,17 @@ func (p *auditPlugin) truncateTCPEventStream(eventStream []byte, connType string
...
+	sh.Len = 8
+	sh.Cap = 8
+	sh.Data = uintptr(unsafe.Pointer(&i))
+
+	return b[:]
+}
+
+func byteArrayToInt(b []byte) int64 {
+	return *(*int64)(unsafe.Pointer(&b[0]))
+}
```
sandromello: You could use BigEndian to encode this properly. We already do this for the native postgres protocol, way more simple, see this: https://go.dev/play/p/Nf6THCiaBPR

gateway/transport/plugins/audit/wal.go
```diff
@@ -64,28 +66,39 @@ func decodeWalHeader(data []byte) (*walHeader, error) {
...
+		return nil, -1, 0, fmt.Errorf("event stream in wrong format [event-type]")
 	}
 	eventType := eventStream[position-1]
+
+	// dlp counter uses 8-byte (int64)
+	position = bytes.LastIndexByte(eventStream, '\000')
+	if position == -1 {
+		return nil, -1, 0, fmt.Errorf("event stream in wrong format [dlp-count]")
+	}
+	eventDlpCounter := eventStream[position-8 : position]
```
sandromello: The event stream contains also the event of all protocols (postgres, tcp, terminal). So you can't use LastIndexByte, because the whole event stream could have this data.

You need to advance the position to get the dlp count. It also requires to change the position to  get the whole event properly when returning the event bytes (see line 100).

gateway/transport/plugins/audit/wal.go
danielbonilha: Nice. Tks.

gateway/transport/plugins/audit/wal.go
danielbonilha: Indeed. Just fixed. Tks.

PR #152, opened 2023-03-27
feat(connections): add icon name on GET return results

rogefm:
Add Icon_name on GET return results

### DIFF
--- a/gateway/connection/service.go
+++ b/gateway/connection/service.go
@@ -32,7 +32,7 @@ type (
 	BaseConnection struct {
 		Id             string         `json:"id"       	edn:"xt/id"`
 		Name           string         `json:"name"     	edn:"connection/name"      	binding:"required"`
-		IconName       string         `json:"icon_name"	edn:"connection/icon_name"`
+		IconName       string         `json:"icon_name"	edn:"connection/icon-name"`
 		Command        []string       `json:"command"  	edn:"connection/command"   	binding:"required"`
 		Type           Type           `json:"type"     	edn:"connection/type"      	binding:"required"`
 		AgentId        string         `json:"agent_id" 	edn:"connection/agent"     	binding:"required"`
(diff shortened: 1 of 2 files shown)


gateway/connection/service.go
```diff
@@ -30,12 +30,13 @@ type (
...
 	BaseConnection struct {
-		Id             string         `json:"id"       edn:"xt/id"`
-		Name           string         `json:"name"     edn:"connection/name"    binding:"required"`
-		Command        []string       `json:"command"  edn:"connection/command" binding:"required"`
-		Type           Type           `json:"type"     edn:"connection/type"    binding:"required"`
-		AgentId        string         `json:"agent_id" edn:"connection/agent"   binding:"required"`
-		SecretProvider SecretProvider `json:"-"        edn:"connection/secret-provider"`
+		Id             string         `json:"id"       	edn:"xt/id"`
+		Name           string         `json:"name"     	edn:"connection/name"      	binding:"required"`
+		IconName       string         `json:"icon_name"	edn:"connection/icon_name"`
```
danielbonilha: for the edn part of the tags, we are using kebab-case... this means that the edn should be "connection/icon-name"

gateway/connection/storage.go
```diff
@@ -19,7 +19,8 @@ type (
 	xtdb struct {
 		Id             string         `edn:"xt/id"`
 		OrgId          string         `edn:"connection/org"`
-		Name           string         `edn:"connection/name" `
+		Name           string         `edn:"connection/name"`
+		IconName       string         `edn:"connection/icon_name"`
```
danielbonilha: for the edn part of the tags, we are using kebab-case... this means that the edn should be "connection/icon-name"

gateway/connection/service.go
rogefm: Done!

gateway/connection/storage.go
rogefm: Done!

PR #164, opened 2023-04-05
added review exec

danielbonilha:
# Created new Endpoint:
POST /api/reviews/{id}/exec

This will cause the review to be executed the same way the runbooks and exec api does.
Returns the resulting log, or 202 if takes more than 50s to finish.

### DIFF
--- a/gateway/review/api.go
+++ b/gateway/review/api.go
@@ -170,6 +170,9 @@ func (h *Handler) RunExec(c *gin.Context) {
 
 	select {
 	case resp := <-clientResp:
+		review.Status = StatusExecuted
+		h.Service.Persist(ctx, review)
+
 		if resp.IsError() {
 			c.JSON(http.StatusBadRequest, &clientexec.ExecErrResponse{
 				SessionID: &resp.SessionID,
@@ -178,9 +181,6 @@ func (h *Handler) RunExec(c *gin.Context) {
 			})
 			return
 		}
-
-		review.Status = StatusExecuted
-		h.Service.Persist(ctx, review)
 
 		c.JSON(statusCode, resp)
 	case <-time.After(time.Second * 50):


gateway/review/api.go
```diff
@@ -111,3 +116,92 @@ func (a *Handler) FindOne(c *gin.Context) {
...
+	statusCode := http.StatusOK
+
+	select {
+	case resp := <-clientResp:
+		if resp.IsError() {
+			c.JSON(http.StatusBadRequest, &clientexec.ExecErrResponse{
+				SessionID: &resp.SessionID,
+				Message:   resp.ErrorMessage(),
+				ExitCode:  resp.ExitCode,
+			})
```
sandromello, line 181: You need to persist the review when the execution returns an error also, because it was executed, but it returned an error

gateway/review/api.go
```diff
@@ -111,3 +116,92 @@ func (a *Handler) FindOne(c *gin.Context) {
...
+
+		c.JSON(statusCode, resp)
+	case <-time.After(time.Second * 50):
+		// closing the client will force the goroutine to end
+		// and the result will return async
+		client.Close()
+
+		// we do not know the status of this in the future.
+		// replaces the current "PROCESSING" status
+		review.Status = StatusUnknown
```
sandromello, line 193: So, the state of this should be PROCESSING, having an `UNKNOWN` state could confuse users. The timeout is when the process is being run for more than 50 seconds. This state currently is being displayed at the webapp

image

danielbonilha: I think if we leave the PROCESSING is when we gonna confuse the users. 

Imagine that a task finished after 3 minutes. Two months later the review will still be PROCESSING. People will get confused if the task is still processing, or the status is wrong. They will never imagine that the task timed out and may have finished later.

If we use UNKNOWN, it is clear to the user that we lost track of the task, but the user is sure that this task started processing in the past.

gateway/review/api.go
danielbonilha: Agree. I moved this logic to the top of the first case, so both success and error will get the status EXECUTED.

gateway/review/api.go
sandromello: Indeed, makes sense. Forgot that we wouldn't be able to update the state of the execution

PR #166, opened 2023-04-06
fixed owner execute task only; changing review state if running via CLI

danielbonilha:
Fixed:
- Only task owner can execute it via API
- When running via terminal, updating the state of the review to PROCESSING

### DIFF
--- a/gateway/transport/client.go
+++ b/gateway/transport/client.go
@@ -441,7 +441,8 @@ func (s *Server) processSessionOpenExec(pkt *pb.Packet, client *client.Client, c
 		if review.Status == rv.StatusApproved {
 			review.Status = rv.StatusProcessing
 			if err := s.ReviewService.Persist(ctx, &review); err != nil {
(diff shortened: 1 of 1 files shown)


gateway/transport/client.go
```diff
@@ -434,6 +434,14 @@ func (s *Server) processSessionOpenExec(pkt *pb.Packet, client *client.Client, c
 			})
 			return nil
 		}
+		ctx := &user.Context{
+			Org:  &user.Org{Id: client.OrgId},
+			User: &user.User{Id: client.UserId},
+		}
+		if review.Status == rv.StatusApproved {
+			review.Status = rv.StatusProcessing
+			s.ReviewService.Persist(ctx, &review)
```
sandromello: You'll need to drop the connection if it fail to persist.

danielbonilha: Right. If it fails we should stop. Fixed.

PR #167, opened 2023-04-11
feature(notifications) add matterbridge as notification service

mtmr0x:
This implements matterbridge managed by the Gateway

### DIFF
--- a/client/cmd/start.go
+++ b/client/cmd/start.go
@@ -181,6 +181,7 @@ var startAgentCmd = &cobra.Command{
 	Short:        "Runs the agent component",
 	SilenceUsage: false,
 	Run: func(cmd *cobra.Command, args []string) {
+
 		agent.Run()
 	},
 }
@@ -190,6 +191,7 @@ var startGatewayCmd = &cobra.Command{
 	Short:        "Runs the gateway component",
 	SilenceUsage: false,
 	Run: func(cmd *cobra.Command, args []string) {
+		log.Printf("running gateway")
 		gateway.Run()
 	},
 }
--- a/gateway/main.go
+++ b/gateway/main.go
@@ -1,9 +1,14 @@
 package gateway
 
 import (
+	"encoding/json"
+	"fmt"
 	"os"
+	"os/exec"
+	"path/filepath"
 	"time"
 
+	"github.com/runopsio/hoop/common/clientconfig"
 	"github.com/runopsio/hoop/common/grpc"
 	"github.com/runopsio/hoop/common/log"
 	"github.com/runopsio/hoop/common/monitoring"
@@ -152,8 +157,54 @@ func Run() {
 }
 
 func getNotification() notification.Service {
-	if os.Getenv("SMTP_HOST") != "" {
+	if os.Getenv("NOTIFICATIONS_BRIDGE_CONFIG") != "" && os.Getenv("ORG_MULTI_TENANT") != "true" {
+		mBridgeConfigRaw := []byte(os.Getenv("NOTIFICATIONS_BRIDGE_CONFIG"))
+		var mBridgeConfigMap map[string]string
+		if err := json.Unmarshal(mBridgeConfigRaw, &mBridgeConfigMap); err != nil {
+			log.Fatalf("failed decoding notifications bridge config")
+		}
+		log.Printf("Bridge notifications selected")
+		matterbridgeConfig := `[slack]
+[slack.myslack]
+Token="%s"
+PreserveThreading=true
+
+[api.myapi]
+BindAddress="127.0.0.1:4242"
+Buffer=10000
+
+[[gateway]]
+name="hoop-notifications-bridge"
+enable=true
+
+[[gateway.in]]
+account="api.myapi"
+channel="api"
+
+[[gateway.out]]
+account="slack.myslack"
+channel="general"`
+		matterbridgeFolder, err := clientconfig.NewHomeDir("matterbridge")
+		if err != nil {
+			log.Fatal(err)
+		}
+		configFile := filepath.Join(matterbridgeFolder, "matterbridge.toml")
+		configFileBytes := []byte(fmt.Sprintf(matterbridgeConfig, mBridgeConfigMap["slackBotToken"]))
+		err = os.WriteFile(configFile, configFileBytes, 0600)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		err = exec.Command("matterbridge", "-conf", configFile).Start()
+
+		if err != nil {
+			log.Fatal(err)
+		}
+		return notification.NewMatterbridge()
+	} else if os.Getenv("SMTP_HOST") != "" {
+		log.Printf("SMTP notifications selected")
 		return notification.NewSmtpSender()
 	}
+	log.Printf("MagicBell notifications selected")
 	return notification.NewMagicBell()
 }
--- a/gateway/notification/matterbridge.go
+++ b/gateway/notification/matterbridge.go
@@ -1,0 +1,64 @@
+package notification
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"os"
+
+	"github.com/runopsio/hoop/common/log"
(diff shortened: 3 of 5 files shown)


client/cmd/start.go
```diff
@@ -190,6 +191,52 @@ var startGatewayCmd = &cobra.Command{
...
+account="api.myapi"
+channel="api"
+
+[[gateway.out]]
+account="slack.myslack"
+channel="general"`,
+			)
+			configFile, err := os.Create("/tmp/matterbridge.toml")
+			if err != nil {
+				panic(err)
```
sandromello: Don't panic, use log.Fatal(err) instead

client/cmd/start.go
```diff
@@ -190,6 +191,52 @@ var startGatewayCmd = &cobra.Command{
...
+
+[[gateway.in]]
+account="api.myapi"
+channel="api"
+
+[[gateway.out]]
+account="slack.myslack"
+channel="general"`,
+			)
+			configFile, err := os.Create("/tmp/matterbridge.toml")
```
sandromello: Remove this and use `os.WriteFile()` instead.
To write a configuration inside the template, use `fmt.Sprintf()`

Example:

```go
myvar := "value"
fmt.Sprintf("Token=%v", myvar)
```

client/cmd/start.go
```diff
@@ -190,6 +191,52 @@ var startGatewayCmd = &cobra.Command{
...
+		log.Printf("running gateway")
+		if os.Getenv("NOTIFICATIONS_BRIDGE") == "true" && os.Getenv("NOTIFICATIONS_BRIDGE_URL") == "" {
+			exec.Command(
+				"curl",
+				"-sL",
+				"https://github.com/42wim/matterbridge/releases/download/v1.26.0/matterbridge-1.26.0-darwin-arm64",
+				"-o",
+				"/tmp/matterbridge-1.26.0-darwin-arm64",
+			).Run()
+			exec.Command("chmod", "a+x", "/tmp/matterbridge-1.26.0-darwin-arm64").Run()
```
sandromello: You need to put all this to the Dockerfile.dev, not here. You'll need to download the proper binary depending on the architecture that you're building. I can assist you if don't know how to do that.

client/cmd/start.go
```diff
@@ -190,6 +191,52 @@ var startGatewayCmd = &cobra.Command{
...
+				panic(err)
+			}
+			defer configFile.Close()
+			writeCmd.Stdout = configFile
+			err = writeCmd.Start()
+			if err != nil {
+				panic(err)
+			}
+			writeCmd.Wait()
+			exec.Command("/tmp/matterbridge-1.26.0-darwin-arm64", "-conf", "/tmp/matterbridge.toml").Start()
```
sandromello: Use `exec.Command("").CombinedOutput()`, this will block until the command executes and return any errors. It's important to not let it fail silently.

client/cmd/start.go
```diff
@@ -190,6 +191,52 @@ var startGatewayCmd = &cobra.Command{
 	Short:        "Runs the gateway component",
 	SilenceUsage: false,
 	Run: func(cmd *cobra.Command, args []string) {
+		log.Printf("running gateway")
+		if os.Getenv("NOTIFICATIONS_BRIDGE") == "true" && os.Getenv("NOTIFICATIONS_BRIDGE_URL") == "" {
```
sandromello: I think this will be too cumbersome to explain and document it. I think it's better to put the whole structure into a env and parse it, example:

```go
mBridgeConfigRaw := []byte(os.Getenv("MATTERBRIDGE_CONFIG"))
var mBridgeConfigMap map[string]any
if err := json.Unmarshal(mBridgeConfigRaw, &mBridgeConfigMap); err != nil {
	log.Fatalf("failed decoding matter bridge config")
}
_ = mBridgeConfigMap["slack-bot-token"]
_ = mBridgeConfigMap["url"]
(...)
```

When runnning the gateway:

```sh
export MATTERBRIDGE_CONFIG='{"config01": "val01", "config02": "val02"}'
```

gateway/main.go
```diff
@@ -153,8 +153,13 @@ func Run() {
 }
 
 func getNotification() notification.Service {
-	if os.Getenv("SMTP_HOST") != "" {
+	if os.Getenv("NOTIFICATIONS_BRIDGE") == "true" && os.Getenv("ORG_MULTI_TENANT") != "false" {
```
sandromello: This should be `os.Getenv("ORG_MULTI_TENANT") != "true"`, only this condition will make the gateway to be a multi tenant setup.

gateway/notification/matterbridge.go
```diff
@@ -0,0 +1,58 @@
...
+		"application/json",
+		bytes.NewBuffer(jsonData),
+	)
+}
+
+func (m *Matterbridge) IsFullyConfigured() bool {
+	log.Printf("Testing bridge connection")
+	_, err := http.Get(m.bridgeUrl)
+	if err != nil {
+		log.Printf("Couldn't connect with bridge")
```
sandromello: Use `log.Errorf()`

gateway/notification/matterbridge.go
```diff
@@ -0,0 +1,62 @@
...
+	"encoding/json"
+	"net/http"
+	"os"
+
+	"github.com/runopsio/hoop/common/log"
+)
+
+func getBridgeUrl(configMap map[string]string) string {
+	if len(configMap["bridgeUrl"]) > 0 {
+		log.Printf("Using NOTIFICATIONS_BRIDGE_URL env var")
```
sandromello: This env doesn't exists anymore, change the log message.

gateway/notification/matterbridge.go
```diff
@@ -0,0 +1,62 @@
...
+	Matterbridge struct {
+		bridgeUrl     string
+		slackBotToken string
+	}
+)
+
+func NewMatterbridge() *Matterbridge {
+	mBridgeConfigRaw := []byte(os.Getenv("NOTIFICATIONS_BRIDGE_CONFIG"))
+	var mBridgeConfigMap map[string]string
+	json.Unmarshal(mBridgeConfigRaw, &mBridgeConfigMap)
```
sandromello: You're decoding map[string]any at the start and then decoding as map[string]string here, you should decode the same object when starting the gateway, then it will be safe to not check for this error.

gateway/notification/matterbridge.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return &Matterbridge{
+		bridgeUrl:     getBridgeUrl(mBridgeConfigMap),
+		slackBotToken: mBridgeConfigMap["slackBotToken"],
```
sandromello, line 37: This will panic if the line above fail, just check the error above to be more defensive.

gateway/notification/matterbridge.go
```diff
@@ -0,0 +1,62 @@
...
+		"application/json",
+		bytes.NewBuffer(jsonData),
+	)
+}
+
+func (m *Matterbridge) IsFullyConfigured() bool {
+	log.Printf("Testing bridge connection")
+	_, err := http.Get(m.bridgeUrl)
+	if err != nil {
+		log.Errorf("Couldn't connect with bridge")
```
sandromello: Add the error placeholder in the log message

client/cmd/start.go
```diff
@@ -190,6 +193,50 @@ var startGatewayCmd = &cobra.Command{
...
+			err = os.WriteFile(configFile, configFileBytes, 0600)
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			err = exec.Command("matterbridge", "-conf", configFile).Start()
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
```
sandromello: Instead of having repeated logic here and in `gateway/main.go#getNotification`, just put all this logic in the getNotification function, it will be called when starting the gateway. You could make the program die if a condition is not met.

mtmr0x: makes sense

PR #173, opened 2023-04-17
added dlp count with ordering and pagination

danielbonilha:


gateway/session/storage.go
```diff
@@ -184,6 +184,7 @@ func (s *Storage) FindAll(ctx *user.Context, opts ...*SessionOption) (*SessionLi
 					[a :session/event-size event-size]
 					[a :session/start-date start-date]
 					[a :session/end-date end-date]
+					[(get-attr a :session/dlp-count) [dlp-count]]
```
sandromello: This won't work with the schema that we have now, dlp-count will return nil and fail to decode to int64.
You also need to add the proper attributes on find also to return this attribute.

danielbonilha: Added a default to 0 (zero) in case it is null.

PR #182, opened 2023-04-28
Enhanc/private endpoints

danielbonilha:
**#Features** 

1) Created a new webserver on port 8020
- Only starts if ORG_MULTI_TENANT = false
- This port is **not** supposed to be exposed to the internet, because it contains unprotected endpoints
- The endpoints are made for usage by systems inside the private network, like the slack-integration service
- Only a single endpoint is exposed: PUT /reviews/{id}
- The slack-integration service will call this endpoint when a user interact with slack to do some review. 

2) Created new plugin: slack
- The configuration in the plugin is tuples of email and slack id, configured by the admin:
`[]config {
  daniel@hoop.dev, @U01LGAWT2D7, 
  jose@hoop.dev, @W01MGAWO9D2
}`

3) During an exec, if there is review in the flow, the code checks for the presence of the slack plugin. If present, than send a message to it. If not review or not slack, than nothing is done
- Done asynchronously

**#Bugs**
- Fixed POST /api/connections/{conn-name}/exec 
- When the connection has review, the endpoint was not returning the review info to the user
- Instead, it waited the 50s timeout and sent a session ID back to the user
- Now the endpoint returns the reviewURI to the user, likewise happens in the CLI

POST /api/connections{conn-name}/exec
`
{
  "message":"this connection requires review",
  "reviewURL": "http://localhost:8009/api/reviews/39d90e8c-3b22-43b7-9490-818611b18629"
}
`

### DIFF
--- a/client/cmd/admin/create-user.go
+++ b/client/cmd/admin/create-user.go
@@ -1,0 +1,96 @@
+package admin
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/runopsio/hoop/client/cmd/styles"
+	clientconfig "github.com/runopsio/hoop/client/config"
+	"github.com/runopsio/hoop/common/log"
+	"github.com/spf13/cobra"
+)
+
+var (
+	userDisableFlag   bool
+	userGroupsFlag    []string
+	userNameFlag      string
+	userSlackIDFlag   string
+	userOverwriteFlag bool
+)
+
+func init() {
+	createUserCmd.Flags().BoolVar(&userOverwriteFlag, "overwrite", false, "It will create or update it if a connection already exists")
+	createUserCmd.Flags().StringSliceVar(&userGroupsFlag, "groups", []string{}, "The list of groups this user belongs to, e.g.: admin,devops,...")
+	createUserCmd.Flags().StringVar(&userNameFlag, "name", "", "The display name of the user")
+	createUserCmd.Flags().StringVar(&userSlackIDFlag, "slackid", "", "The slack id of the user, only useful with slack plugin")
+	createUserCmd.Flags().BoolVar(&userDisableFlag, "disable", false, "Disable the user")
+}
+
+var createUserCmd = &cobra.Command{
+	Use:     "user NAME",
+	Aliases: []string{"users"},
+	Short:   "Create a user resource.",
+	PreRun: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			cmd.Usage()
+			styles.PrintErrorAndExit("missing resource name")
+		}
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		resourceName := args[0]
+		actionName := "created"
+		method := "POST"
+		userID, err := getUser(clientconfig.GetClientConfigOrDie(), resourceName)
+		if err != nil {
+			styles.PrintErrorAndExit("failed retrieving user %v, %v", resourceName, err)
+		}
+		if userID != "" && userOverwriteFlag {
+			log.Debugf("user %v/%v exists, updating", userID, resourceName)
+			actionName = "updated"
+			method = "PUT"
+		}
+		resourceArgs := []string{"users", userID}
+		apir := parseResourceOrDie(resourceArgs, method, outputFlag)
+		status := "active"
+		if userDisableFlag {
+			status = "inactive"
+		}
+		resp, err := httpBodyRequest(apir, method, map[string]any{
+			// immutable fields
+			"email": resourceName,
+			"name":  userNameFlag,
+			// updatable fields
+			"groups":   userGroupsFlag,
+			"slack_id": userSlackIDFlag,
+			"status":   status,
+		})
+		if err != nil {
+			styles.PrintErrorAndExit(err.Error())
+		}
+		if apir.decodeTo == "raw" {
+			jsonData, _ := resp.([]byte)
+			fmt.Println(string(jsonData))
+			return
+		}
+		fmt.Printf("user %v %v\n", resourceName, actionName)
+	},
+}
+
+func getUser(conf *clientconfig.Config, userName string) (string, error) {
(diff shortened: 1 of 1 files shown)


gateway/api/middleware.go
```diff
@@ -123,3 +123,60 @@ func parseHeaderForDebug(authTokenHeader string) string {
...
+// when slack plugin is activated only.
+// assumes the org is named 'default'
+func (api *Api) SetSlackContext(c *gin.Context) {
+	userSlackID := c.GetHeader("slack-id")
+	if userSlackID == "" {
+		c.AbortWithStatus(412)
+		return
+	}
+
+	org, err := api.UserHandler.Service.GetOrgByName("default")
```
sandromello: Use the constant `pb.DefaultOrgName`

gateway/api/middleware.go
```diff
@@ -123,3 +123,60 @@ func parseHeaderForDebug(authTokenHeader string) string {
...
+	if err != nil || org == nil {
+		c.AbortWithStatus(412)
```
sandromello: Try to always log things when an undesired state is present, otherwise we won't be able to help users.
Always return a message to the client to indicate what happened, like an attribute with `message` containing a generic message or something alike.

There're two states here:

- The organization default wasn't found
- It found an error on searching the organization (maybe xtdb was down or something else)

Both states are not something that the user could control and we should return something like 500 error code, because something is really wrong.

If these status code are expected by slack, at least put logging  to see when such cases happened.

gateway/clientexec/clientexec.go
```diff
@@ -97,6 +98,14 @@ func newError(err error) *Response {
 	return &Response{err: err}
 }
 
+func newReviewedResponse(reviewURI string) *Response {
+	exit := 0
+	return &Response{
+		ExitCode:  &exit,
```
sandromello: It should be nil, because the execution didn't started.

gateway/connection/api.go
```diff
@@ -203,6 +203,12 @@ func (h *Handler) RunExec(c *gin.Context) {
 			})
 			return
 		}
+		if resp.ReviewURI != "" {
+			c.JSON(http.StatusAccepted, gin.H{
+				"message":   "this connection requires review",
+				"reviewURI": resp.ReviewURI})
+			return
+		}
```
sandromello: I think it's better to add a boolean `has_review` attribute and put the url in the output attribute, then you won't have to change nothing in this logic. Otherwise you'll be require to change the exec in 3 places (runbook, conn-exec, review-exec).

The `message` attribute is being used only when there's an non 2xx status code, let's continue to keep this pattern.

gateway/transport/client.go
```diff
@@ -412,11 +413,15 @@ func (s *Server) processSessionOpenExec(pkt *pb.Packet, client *client.Client, c
...
 		}
 		if !(review.Status == rv.StatusApproved || review.Status == rv.StatusProcessing) {
+			reviewURL := fmt.Sprintf("%s/plugins/reviews/%s", s.IDProvider.ApiURL, review.Id)
 			clientStream := getClientStream(client.SessionID)
 			_ = clientStream.Send(&pb.Packet{
 				Type:    pbclient.SessionOpenWaitingApproval,
-				Payload: []byte(fmt.Sprintf("%s/plugins/reviews/%s", s.IDProvider.ApiURL, review.Id)),
+				Payload: []byte(reviewURL),
 			})
+			apiReviewURL := strings.Replace(reviewURL, "plugins", "api", -1)
```
sandromello: Try not to do this, it's more harder to read, just construct the URL again.

gateway/api/server.go
```diff
@@ -283,3 +283,40 @@ func (api *Api) CreateTrialEntities() error {
...
+
+	if sentryInit {
+		rg.Use(sentrygin.New(sentrygin.Options{
+			Repanic: true,
+		}))
+	}
+
+	api.buildPrivateRoutes(rg)
+	if err := route.Run(":" + port); err != nil {
+		panic("Failed to start HTTP server")
```
sandromello: Don't panic, use log.Fatal instead with the error in the message

gateway/transport/plugins/slack/service.go
```diff
@@ -0,0 +1,87 @@
...
+
+	req.Header.Set("content-type", "application/json")
+
+	resp, err := slack.client.Do(req)
+	if err != nil {
+		log.Errorf("failed to send slack message, err=%v", err)
+		sentry.CaptureException(err)
+		return
+	}
+	defer resp.Body.Close()
```
sandromello: Don't need to use defer here

gateway/api/middleware.go
```diff
@@ -123,3 +123,60 @@ func parseHeaderForDebug(authTokenHeader string) string {
...
+		return
+	}
+
+	var userID string
+	for _, conn := range p.Connections {
+		for i, g := range conn.Config {
+			if i > 0 && g == userSlackID {
+				userID = conn.Config[i-1]
+			}
```
sandromello: Don't you need to break here after finding the userID?

gateway/api/middleware.go
```diff
@@ -123,3 +123,60 @@ func parseHeaderForDebug(authTokenHeader string) string {
...
+		}
+	}
+
+	if userID == "" {
+		c.AbortWithStatus(401)
```
sandromello: Log something here

gateway/api/middleware.go
```diff
@@ -123,3 +123,60 @@ func parseHeaderForDebug(authTokenHeader string) string {
...
+		return
+	}
+
+	loggedUser, err := api.UserHandler.Service.FindOne(ctx, userID)
+	if err != nil || loggedUser == nil {
+		c.AbortWithStatus(412)
```
sandromello: Log something here wrapping the error and the user variable

PR #183, opened 2023-05-02
Make sessions to return when querying with a non-admin user

sandromello:
- Fix session unauthenticated error when browsing with non-admin user
- Make all non-admin routes to return 403 instead of 401
- Add trace log level for plugin hooks

### DIFF
--- a/gateway/session/storage.go
+++ b/gateway/session/storage.go
@@ -17,7 +17,7 @@ type (
 		*st.Storage
 	}
 	GenericStorageWriter struct {
-		persistFn func(*user.Context, *Session) (*st.TxResponse, error)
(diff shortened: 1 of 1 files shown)


gateway/session/storage.go
```diff
@@ -223,7 +237,9 @@ func (s *Storage) FindOne(ctx *user.Context, sessionID string) (*Session, error)
 }
 
 func (s *Storage) queryDecoder(query string, into any, args ...any) error {
-	httpBody, err := s.QueryRaw([]byte(fmt.Sprintf(query, args...)))
+	qs := fmt.Sprintf(query, args...)
+	fmt.Println("QS->>", qs)
```
mtmr0x: maybe this was left here?

PR #194, opened 2023-05-12
 feature(session) add session exec endpoint

mtmr0x:
This is still WIP, but wanted to have it remotely just in case

### DIFF
--- a/gateway/api/server.go
+++ b/gateway/api/server.go
@@ -233,6 +233,23 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 		api.TrackRequest,
 		api.SessionHandler.FindAll)
 
+	route.GET("/sessions/:session_id",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.FindOne)
+	route.GET("/sessions/:session_id/status",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.StatusHistory)
+	route.GET("/sessions",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.FindAll)
+	route.POST("/sessions/:session_id/exec",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.RunExec)
+
 	route.POST("/plugins/indexer/sessions/search",
 		api.Authenticate,
 		api.TrackRequest,
--- a/gateway/review/service.go
+++ b/gateway/review/service.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/google/uuid"
 	pb "github.com/runopsio/hoop/common/proto"
+	"github.com/runopsio/hoop/gateway/session"
+	st "github.com/runopsio/hoop/gateway/storage"
 	"github.com/runopsio/hoop/gateway/user"
 )
 
@@ -21,6 +23,8 @@ type (
 		FindAll(context *user.Context) ([]Review, error)
 		FindBySessionID(sessionID string) (*Review, error)
 		FindApprovedJitReviews(ctx *user.Context, connID string) (*Review, error)
+		PersistSessionAsReady(s *session.Session) (*st.TxResponse, error)
+		FindSessionBySessionId(sessionID string) (*session.Session, error)
 	}
 
 	transportService interface {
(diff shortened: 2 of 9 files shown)


gateway/api/server.go
```diff
@@ -233,6 +233,23 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
...
+		api.SessionHandler.StatusHistory)
+	route.GET("/sessions",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.FindAll)
+	route.POST("/sessions/:session_id/exec",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.RunExec)
+
```
mtmr0x, line 252: since we're moving session to core feature, I already duplicated these and I will update on the frontend to use these ones, when we know there's no one using the older ones, we remove them

gateway/review/service.go
```diff
@@ -185,6 +189,14 @@ func (s *Service) Review(context *user.Context, reviewID string, status Status)
...
 
 	if rev.Status == StatusApproved || rev.Status == StatusRejected {
+		currentSession, err := s.Storage.FindSessionBySessionId(rev.Session)
+		if err != nil {
+			return nil, err
+		}
+		_, err = s.Storage.PersistSessionAsReady(currentSession)
+		if err != nil {
+			return nil, err
+		}
```
sandromello, line 199: This is not safe, use s.Persist(context, currentSession) instead.

gateway/review/storage.go
```diff
@@ -150,6 +153,50 @@ func (s *Storage) FindById(ctx *user.Context, id string) (*Review, error) {
...
+	if strings.Contains(string(httpBody), ":xtdb.error") {
+		return fmt.Errorf(string(httpBody))
+	}
+	return edn.Unmarshal(httpBody, into)
+}
+
+func (s *Storage) PersistSessionAsReady(sess *session.Session) (*st.TxResponse, error) {
+	sess.Status = "ready"
+	return s.SubmitPutTx(sess)
+}
```
sandromello, line 171: Don't use this, use Service.Persist instead, it's more safe.

gateway/review/storage.go
```diff
@@ -150,6 +153,50 @@ func (s *Storage) FindById(ctx *user.Context, id string) (*Review, error) {
...
+
+func (s *Storage) FindSessionBySessionId(sessionID string) (*session.Session, error) {
```
sandromello, line 173: `Service.FindOne(...)` already fetches a session by id, this function is not necessary.

gateway/session/storage.go
```diff
@@ -171,6 +194,7 @@ func (s *Storage) FindAll(ctx *user.Context, opts ...*SessionOption) (*SessionLi
 					[a :session/user usr]
 					[a :session/user-id usr-id]
 					[a :session/user-name usr-name]
+					[a :session/status status]
```
sandromello: This will break previous sessions from being displayed, you need to fetch it conditionally, like:
```
[(get-attr a :session/dlp-count 0) [dlp-count]]
```

gateway/api/server.go
```diff
@@ -233,6 +233,23 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
...
+		api.TrackRequest,
+		api.SessionHandler.StatusHistory)
+	route.GET("/sessions",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.FindAll)
+	route.POST("/sessions/:session_id/exec",
+		api.Authenticate,
+		api.TrackRequest,
+		api.SessionHandler.RunExec)
```
sandromello, line 251: If you want to add a new endpoint with the same behavior, just reference the review handler instead of duplicating this.

gateway/review/service.go
```diff
@@ -185,6 +189,14 @@ func (s *Service) Review(context *user.Context, reviewID string, status Status)
 	}
 
 	if rev.Status == StatusApproved || rev.Status == StatusRejected {
+		currentSession, err := s.Storage.FindSessionBySessionId(rev.Session)
+		if err != nil {
+			return nil, err
+		}
+		_, err = s.Storage.PersistSessionAsReady(currentSession)
```
sandromello, line 196: This state is not ready because it could be rejected also.

gateway/transport/plugins/audit/wal.go
```diff
@@ -28,6 +28,7 @@ type (
 		UserEmail      string     `json:"user_email"`
 		ConnectionName string     `json:"connection_name"`
 		ConnectionType string     `json:"connection_type"`
+		Status         any        `json:"status"`
```
sandromello: Change to string here

gateway/transport/plugins/audit/wal.go
```diff
@@ -115,6 +116,7 @@ func (p *auditPlugin) writeOnConnect(pctx plugintypes.Context) error {
 		ConnectionName: pctx.ConnectionName,
 		ConnectionType: pctx.ConnectionType,
 		Verb:           pctx.ClientVerb,
+		Status:         pctx.ParamsData["status"],
```
sandromello: Use `pctx.ParamsData.GetString("status")`

PR #199, opened 2023-05-31
feature(sessions) add script and labels fields

mtmr0x:
This PR implements

- sessions to the new folder architecture
- an overall refactor that starts a simplification in the structure of the sessions in the Gateway
- adds the field Script to the session
- adds the field Labels to the session
- adds the labels in a Runbook session for the Runbook file name and parameters inserted by the user

### DIFF
--- a/gateway/connection/api.go
+++ b/gateway/connection/api.go
@@ -156,9 +156,9 @@ func (a *Handler) Evict(c *gin.Context) {
 	c.Writer.WriteHeader(204)
 }
 
-// This functions is legacy and should not be used anymore
-// Use sessionapi.RunExec at api/session/run-exec.go
+// DEPRECATED in flavor of POST /api/sessions
 func (h *Handler) RunExec(c *gin.Context) {
+	log.Warnf("executing connection run-exec - deprecated endpoint")
 	ctx := user.ContextUser(c)
 	storageCtx := storagev2.ParseContext(c)
 
--- a/gateway/review/api.go
+++ b/gateway/review/api.go
@@ -112,10 +112,11 @@ func (h *Handler) FindOne(c *gin.Context) {
 }
 
 // DEPRECATED
-// Use sessionapi.RunExec instead
+// This function/endpoint is deprecated in flavor of POST /api/sessions/:session-id/exec
 func (h *Handler) RunExec(c *gin.Context) {
 	ctx := user.ContextUser(c)
 	log := user.ContextLogger(c)
+	log.Warnf("executing review run-exec - deprecated endpoint")
 	reviewID := c.Param("id")
 	if isLockedForExec(reviewID) {
 		errMsg := fmt.Sprintf("the review %v is already being processed", reviewID)
--- a/gateway/runbooks/api.go
+++ b/gateway/runbooks/api.go
@@ -93,8 +93,7 @@ func (h *Handler) FindAll(c *gin.Context) {
 	c.PureJSON(http.StatusOK, list)
 }
 
-// DEPRECATED
-// Use sessionapi.RunExec instead
+// TODO: Refactor to use sessionapi.RunExec
 func (h *Handler) RunExec(c *gin.Context) {
 	ctx := user.ContextUser(c)
 	log := user.ContextLogger(c)
--- a/gateway/session/api.go
+++ b/gateway/session/api.go
@@ -165,8 +165,7 @@ func getAccessToken(c *gin.Context) string {
 	return ""
 }
 
-// DEPRECATED
-// Use sessionapi.RunExec instead
+// TODO: Refactor to use sessionapi.RunExec
 func (h *Handler) RunExec(c *gin.Context) {
 	ctx := user.ContextUser(c)
 	log := user.ContextLogger(c)
--- a/gateway/api/server.go
+++ b/gateway/api/server.go
@@ -184,7 +184,7 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 		api.Authenticate,
 		api.TrackRequest,
 		api.ReviewHandler.Put)
-	// DEPRECATED in flavor of POST /sessions
+	// DEPRECATED in flavor of POST /api/sessions/:session_id/exec
 	route.POST("/reviews/:id/exec",
 		api.Authenticate,
 		api.TrackRequest,
@@ -272,7 +272,6 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 		api.Authenticate,
 		api.TrackRequest,
 		sessionapi.Post)
-	// DEPRECATED in flavor of POST /sessions
 	route.POST("/sessions/:session_id/exec",
 		api.Authenticate,
 		api.TrackRequest,


gateway/api/session/session.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/runopsio/hoop/gateway/user"
+)
+
+type SessionPostBody struct {
+	Script     string `json:"script"     edn:"session/script"`
+	Verb       string `json:"verb"       edn:"session/verb"` // TODO remove this in the future and try to understand which verb based on the connection
+	Connection string `json:"connection" edn:"session/connection"`
+	Labels     types.SessionLabels
+	ClientArgs []string `json:"clientArgs" edn:"session/client-args"`
+}
```
sandromello: You're mixing the storage layer with the api.

You need to have a struct at `gateway/storagev2/types` with edn tags and a separate struct only for the api layer.

gateway/api/session/session.go
```diff
@@ -0,0 +1,47 @@
...
+
+func Post(c *gin.Context) {
+	ctx := user.ContextUser(c)
```
sandromello, line 26: You don't need this anymore, `storagev2.ParseContext` will give the user context also.

gateway/transport/client.go
```diff
@@ -87,6 +89,16 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
 	connectionName := mdget(md, "connection-name")
 	clientVerb := mdget(md, "verb")
 	clientOrigin := mdget(md, "origin")
+	sessionRequestBody := mdget(md, "session-request")
```
sandromello: You're using an HTTP header to propagate the request. This has some problems like:

- Headers have a limited size in most load balancer setups (nginx uses a default of 16k)
- It is not compatible with runops (we have customers that create tasks with more than 16KB), meaning that this will break the connection with the gRPC gateway.

This also will not be compatible with other clients (only executions made from api will have the input script).

Instead of propagating via header when connecting the more compatible way is doing this in the open session phase. Actually the "session-request" is already being propagated in this phase for the clientexec.

You just need to implement a phase to catch it in the gateway and save the input in the xtdb. I can help you to explain how to do that.

gateway/storagev2/types/types.go
```diff
@@ -29,3 +33,41 @@ type Client struct {
...
+	Verb        string             `json:"verb"         edn:"session/verb"`
+	Status      string             `json:"status"       edn:"session/status"`
+	DlpCount    int64              `json:"dlp_count"    edn:"session/dlp-count"`
+	EventStream SessionEventStream `json:"event_stream" edn:"session/event-stream"`
+	// Must NOT index streams (all top keys are indexed in xtdb)
+	NonIndexedStream SessionNonIndexedEventStreamList `json:"-"          edn:"session/xtdb-stream"`
+	EventSize        int64                            `json:"event_size" edn:"session/event-size"`
+	StartSession     time.Time                        `json:"start_date" edn:"session/start-date"`
+	EndSession       *time.Time                       `json:"end_date"   edn:"session/end-date"`
+}
```
sandromello, line 74: You're mixing the api layer with the storage one. Let's separate this properly, it's really hard to reason about this structure when it's mixed.

gateway/storagev2/connection/connection.go
```diff
@@ -0,0 +1,38 @@
...
+	"olympos.io/encoding/edn"
+)
+
+func GetOneByName(storage *storage.Context, userCtx *user.Context, name string) (*types.Connection, error) {
+	var payload = `{:query {
+		:find [(pull ?connection [*])] 
+		:in [name org]
+		:where [[?connection :connection/name name]
+                [?connection :connection/org org]]}
+		:in-args ["` + name + `" "` + userCtx.Org.Id + `"]}`
```
sandromello: Try to use `fmt.Sprintf(":in-args [%q %q]", name, ...)` for composing the queries, it's harder to read the query when concatenating the strings in this way.

gateway/api/session/session.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	if body.Connection == "" {
+		c.JSON(http.StatusBadRequest, gin.H{"message": "You must provide the connection name"})
+		return
+	}
+
+	connection, err := connectionstorage.GetOneByName(storageCtx, ctx, body.Connection)
+	if connection == nil {
+		c.JSON(http.StatusBadRequest, gin.H{"message": "Connection not found"})
```
sandromello, line 41: It should be status not found instead of bad request

gateway/connection/api.go
```diff
@@ -151,69 +151,34 @@ func (a *Handler) Evict(c *gin.Context) {
...
-		client.Close()
-		c.JSON(http.StatusAccepted, gin.H{"session_id": client.SessionID(), "exit_code": nil})
+		c.JSON(http.StatusBadRequest, gin.H{"message": err.Error()})
+		return
 	}
+
+	// appending info to body to be like SessionPostBody struct
+	body.Connection = connectionName
+	// As this endpoint is exclusive for exec, we're forcing the Verb to be exec
+	body.Verb = "exec"
```
sandromello: Use constants instead: `common/proto/const.go#ClientVerbExec`

gateway/review/api.go
```diff
@@ -111,6 +111,8 @@ func (h *Handler) FindOne(c *gin.Context) {
 	c.PureJSON(http.StatusOK, review)
 }
 
+// This function is being deprecated
+// Use sessionapi.RunExec instead
```
sandromello: Put the name `DEPRECATED <explanation>` at the beginning, the go tooling will indicate that this function is deprecated when hovering on IDE's

gateway/runbooks/api.go
```diff
@@ -87,6 +90,8 @@ func (h *Handler) FindAll(c *gin.Context) {
 	c.PureJSON(http.StatusOK, list)
 }
 
+// This function is being deprecated
+// Use sessionapi.RunExec instead
```
sandromello: Use the syntax `DEPRECATED <description>`

gateway/storagev2/session/session.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"github.com/runopsio/hoop/gateway/storagev2"
+	"github.com/runopsio/hoop/gateway/storagev2/types"
+)
+
+func Write(storage *storagev2.Context, session *types.Session) error {
+	_, err := storage.Put(session)
+	return err
+}
```
sandromello: Is this being used in somewhere?

gateway/storagev2/types/types.go
mtmr0x: Move this struct partially to the API layer with the fields needed there to the SessionResponse

gateway/storagev2/session/session.go
mtmr0x: Just changed some stuff on the code after we spoke and it became used

gateway/storagev2/types/types.go
mtmr0x: @sandromello can't we assume these structures can be reused around the app when they have no reason to be different on the payload? That would just avoid a lot of places with very close structs

sandromello: Yes. The goal is to use the storage structure in any place that you want it.
The only exception is avoiding using it with json tags.

mtmr0x: is there a why? The json tags allow me to use it to respond on the API without pretty much duplicating all this structure

sandromello: - The storage model may include fields that we don't want to expose to the api layer
- Sometimes we may wish to expose fields to the api layer that we don't want to expose to the db layer
- We may wish to compute some field for the API that we don't want for the storage layer
- We can easily forget to save an important field because we lack any sort of constraints and both layers right now

It's very difficulty to understand and read the contract of the API if we're mixing these two layers. I can't know for sure the fields of an structure looking to a mixed structure, because there are fields that aren't relevant to the API contract.

The reasoning about this is having a more legible code and I think this decoupling helps with that.

gateway/session/api.go
```diff
@@ -242,7 +248,37 @@ func (h *Handler) RunExec(c *gin.Context) {
...
+		ID:           review.Session,
+		OrgID:        ctx.Org.Id,
+		Labels:       session.Labels,
+		Script:       session.Script,
+		UserEmail:    ctx.User.Email,
+		UserID:       ctx.User.Id,
+		UserName:     ctx.User.Name,
+		Type:         session.Type,
+		Connection:   session.Connection,
+		Verb:         "exe",  // TODO use a const
```
sandromello: Did you forget to use the constants here?

gateway/storagev2/connection/connection.go
```diff
@@ -0,0 +1,39 @@
...
+
+	if len(connections) == 0 {
+		return nil, nil
+	}
+
+	conn := connections[0][0]
+	// secret, err := storage.getSecret(conn.SecretId)
+	if err != nil {
+		return nil, err
+	}
```
sandromello: You don't need this anymore, you're already checking when unmarshaling it.

gateway/storagev2/session/session.go
```diff
@@ -0,0 +1,37 @@
...
+	b, err := storageCtx.Query(payload)
+	if err != nil {
+		return nil, err
+	}
+
+	var sessions [][]types.Session
+	if err := edn.Unmarshal(b, &sessions); err != nil {
+		return nil, err
+	}
+	session := sessions[0][0]
```
sandromello: This could panic, try to at least check if the first record has any content first:
```go
len(sessions) > 0 {
  return &sessions[0][0], nil
}
```

In practice, if a 200 status code doesn't return anything, then it will return an empty list `[]`

gateway/transport/client.go
```diff
@@ -113,10 +113,11 @@ func (s *Server) subscribeClient(stream pb.Transport_ConnectServer, token string
...
 	sessionID := mdget(md, "session-id")
-	if sessionID != "" {
-		if err := s.validateSessionID(sessionID); err != nil {
-			return status.Errorf(codes.AlreadyExists, err.Error())
-		}
+	session, err := s.SessionService.FindOne(userCtx, sessionID)
+	// Checks if the context org and the session org matches for security reasons
+	// some clients can provide the session ID and we need to double check with this operation
+	if session != nil && session.OrgID != userCtx.Org.Id {
+		return status.Errorf(codes.Aborted, "You don't have access to this session")
```
sandromello: There're a few details here:

1. You're already verifying if the session belongs to the org inside the `FindOne` function
2. You need to only perform this validation when the `session-id` is set, which means that a client is trying to derive the session id beforehand (in this case the API layer). This will prevent from breaking other clients (hoop cli)
3. You need to check for the error
4. If the session is empty, then it means that the user doesn't have access to the session.

gateway/transport/plugins/audit/wal.go
```diff
@@ -197,6 +205,14 @@ func (p *auditPlugin) writeOnClose(sessionID string) error {
 		idx++
 	}
 	endDate := time.Now().UTC()
+
+	newStorage := storagev2.NewStorage(nil)
+	storageContext := storagev2.NewContext(wh.UserID, wh.OrgID, newStorage)
+	session, err := sessionStorage.FindOne(storageContext, wh.SessionID)
+	if err != nil {
+		return err
+	}
```
sandromello: If you're getting this information from the database, it's better to permit the `storageWriter` to fetch the session and populate this attributes, then you don't need this additional logic in this layer. See:
https://github.com/hoophq/hoop/blob/0f9e2961ee3ebe14c02414cede40bb5802743731/gateway/session/storage.go#L346-L382

Then the command below will return the error in case it doesn't find the session (line 236)

This function also could panic because you're checking if the session obj is not nil.

gateway/session/api.go
mtmr0x: no, I left it because this constant is not created, this would require to change to the const on more points of the application and we're running out of time to this be on air so we can run some frontend development

PR #205, opened 2023-06-13
Review inside SessionById

rogefm:
The idea of this PR is to bring review inside the request GET session/:id
This will empower the front end to deliver more actions inside the sessions page.

**What was implemented**

- Adapt reviews to the new struct pattern of the gateway.
  - It wasn't possible to bring FindAll to the pattern;
  - It was added Review inside the Session type (I've talked with Sandro and Matheus about it. The Matheus concerns will be validated by Sandro on tests in CLI and Agent).


gateway/api/review/review.go
```diff
@@ -0,0 +1,32 @@
+package sessionapi
```
sandromello: We're already using the package sessionapi on `../session/session.go`, it will be better to change to `reviewapi`

gateway/api/review/review.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"net/http"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/gin-gonic/gin"
+	"github.com/runopsio/hoop/gateway/storagev2"
+	reviewStorage "github.com/runopsio/hoop/gateway/storagev2/review"
```
sandromello: avoid to use camel case on imports, because it could easily overlap with local variable instances, use `reviewstorage` instead of `reviewStorage`

gateway/api/review/review.go
rogefm: Sorry, It was copied code that I forgot

gateway/api/review/review.go
rogefm: Ok!

gateway/storagev2/types/types.go
```diff
@@ -48,6 +48,39 @@ type Connection struct {
...
 
+type ReviewOwner struct {
+	Id    string `json:"id,omitempty"   edn:"xt/id"`
+	Name  string `json:"name,omitempty" edn:"review-user/name"`
+	Email string `json:"email"          edn:"review-user/email"`
+}
+type ReviewConnection struct {
+	Id   string `json:"id,omitempty" edn:"xt/id"`
+	Name string `json:"name"         edn:"review-connection/name"`
+}
```
sandromello: This structure will not be indexed by xtdb (and it's ok not to be) because it's being used in a nested structure inside the review, so you don't need to use these edn prefixes, instead you could use just:

- `edn:"connection-id"`
- `edn:"connection-name"`
- `edn:"review-username"`
- `edn:"review-email"`

The review group is a special kind, because this is a full entity on xtdb and we could partially update it.

PR #217, opened 2023-07-05
refactor(connections): add command default to postgres and mysql

rogefm:


gateway/connection/api.go
```diff
@@ -85,6 +85,15 @@ func (a *Handler) Post(c *gin.Context) {
...
 	}
 
+	// TODO: Move this alternative solution to other place
+	if connection.Type == "postgres" {
+		connection.Command = []string{"psql", "-A", "-F\t", "-P", "pager=off", "-h", "$HOST", "-U", "$USER", "--port=$PORT", "$DB"}
+	}
+
+	if connection.Type == "mysql" {
+		connection.Command = []string{"mysql", "-h$HOST", "-u$USER", "--port=$PORT", "-D$DB"}
+	}
```
sandromello: Nice! I recommend using the constants instead of named strings:

- `pb.ConnectionTypePostgres`
- `pb.ConnectionTypeMySQL`

rogefm: I will adjust to it

PR #225, opened 2023-07-07
[REFACTOR] Remove session duplication after executing a reviewed session

mtmr0x:


gateway/session/api.go
```diff
@@ -206,19 +201,21 @@ func (h *Handler) RunReviewedExec(c *gin.Context) {
...
+		errMsg := fmt.Sprintf("the session %v is already being processed", sessionId)
 		c.JSON(http.StatusConflict, &clientexec.ExecErrResponse{Message: errMsg})
 		return
 	}
 
 	// locking the execution per review id prevents race condition executions
 	// in case of misbehavior of clients
-	lockExec(reviewID)
-	defer unlockExec(reviewID)
+	lockExec(sessionId)
```
mtmr0x, line 217: @sandromello can I do this? Like, execution locker has both review and session IDs and it will work? I guessed so looking at the application, but it doesn't hurt to ask as well

gateway/session/api.go
```diff
@@ -307,13 +277,13 @@ func (h *Handler) RunReviewedExec(c *gin.Context) {
 		defer close(clientResp)
 		defer client.Close()
 		select {
-		case clientResp <- client.Run([]byte(review.Input), nil):
+		case clientResp <- client.Run([]byte(session.Script["data"]), nil):
```
mtmr0x: this will conflict with what you're doing, but it's fine, I change after you merge yours @sandromello

gateway/session/api.go
sandromello: Yes, no problem on locking the sessionid

PR #227, opened 2023-07-11
[ENHANCEMENT]  bind plugins on connection creation

mtmr0x:
This binds plugins Runbooks, Indexer, Editor and Slack when a connection is created.

https://3.basecamp.com/5385186/buckets/28171754/todos/6310801176

### DIFF
--- a/gateway/connection/service.go
+++ b/gateway/connection/service.go
@@ -107,8 +107,30 @@ func (s *Service) Persist(httpMethod string, context *user.Context, c *Connectio
 	}
 
 	if strings.ToUpper(httpMethod) == "POST" {
-		s.bindAuditPlugin(context, c)
-		s.bindDLPPlugin(context, c)
+		// This automatically sets plugins we
+		// want to be active out of the box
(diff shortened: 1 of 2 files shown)


gateway/connection/service.go
```diff
@@ -107,8 +107,30 @@ func (s *Service) Persist(httpMethod string, context *user.Context, c *Connectio
...
+		// plugins that are simply configured by
+		// the connection name can be passed here
+		for _, plugin := range basicPlugins {
+			s.bindBasicPlugin(context, c, plugin)
+		}
+		// plugins that need to have a
+		// config field are binded here
+		for _, plugin := range pluginsWithConfig {
+			s.bindPluginWithConfig(context, c, plugin)
+		}
```
mtmr0x, line 133: @sandromello I refactored the functions we had and based on those functions I made them be "simple" and "withConfiguration". I figured that this would imply touching as less as possible on how it's passed to the plugin creation, that's why I've done this way. Let me know if you think differently

PR #305, opened 2023-12-29
[REFACTOR] Slack messages and interactivity of review

rogefm:
On this PR was changed:
 - Feedback from Modal to messages, as direct messages and ephemeral messages;
 - Bug fix: Add slack ID into the packet of plugin review. Now we pass SlackID to the owner reviewer when creating a review;
 - Notifies the user that the session can be executed;
 - Change the message of a Review;
 - Notifies the user when they can't review because not being registered, and provides a link to do so.

### DIFF
--- a/gateway/slack/service.go
+++ b/gateway/slack/service.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/runopsio/hoop/common/log"
 	"github.com/slack-go/slack"
 	"github.com/slack-go/slack/socketmode"
 )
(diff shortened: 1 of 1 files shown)


gateway/slack/service.go
```diff
@@ -288,3 +288,26 @@ func (s *SlackService) SendDirectMessage(sessionID, slackID string) error {
...
 	return err
 }
+
+func (s *SlackService) PostMessage(SlackID string, message string) error {
+	channelID, timestamp, err := s.apiClient.PostMessage(SlackID, slack.MsgOptionText(message, false))
+	if err != nil {
+		return fmt.Errorf("failed post message to the channel %v at %v, err=%v", channelID, timestamp, err)
+	}
+
+	fmt.Printf("Message successfully sent to the channel %s at %s", channelID, timestamp)
```
sandromello: Don't use fmt to print logs to stdout. Use the log package "github.com/runopsio/hoop/common/log" instead:
```
log.Infof("<msg>")
```

PR #397, opened 2024-07-08
[readme] add install instructions to readme

andriosrobert:
Improved ux as users have fewer pages to navigate and can get started without leaving readme. I believe we should push this even further


sandromello:
LGTM

sandromello (approved):

PR #407, opened 2024-07-17
Feat/connections access modes

rogefm:
- [gateway] Add access modes policies to block the usage of certain verbs (connect, exec, runbooks)
- [gateway] Add option to toggle the introspection schema of a connection
- [gateway] Add option env to disable webapp users management, `env WEBAPP_USERS_MANAGEMENT=off`

### DIFF
--- a/client/cmd/admin/create-conn.go
+++ b/client/cmd/admin/create-conn.go
@@ -3,6 +3,7 @@ package admin
 import (
 	"encoding/base64"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/hoophq/hoop/client/cmd/styles"
@@ -13,19 +14,17 @@ import (
 )
 
 var (
-	connAgentFlag              string
-	connPuginFlag              []string
-	reviewersFlag              []string
-	connRedactTypesFlag        []string
-	connTypeFlag               string
-	connTagsFlag               []string
-	connSecretFlag             []string
-	connAccessModeRunbooksFlag string
-	connAccessModeExecFlag     string
-	connAccessModeConnectFlag  string
-	connSchemaFlag             string
-	skipStrictValidation       bool
-	connOverwriteFlag          bool
+	connAgentFlag        string
+	connPuginFlag        []string
+	reviewersFlag        []string
+	connRedactTypesFlag  []string
+	connTypeFlag         string
+	connTagsFlag         []string
+	connSecretFlag       []string
+	connAccessModesFlag  []string
+	connSchemaFlag       string
+	skipStrictValidation bool
+	connOverwriteFlag    bool
 )
 
 func init() {
@@ -38,9 +37,7 @@ func init() {
 	createConnectionCmd.Flags().BoolVar(&skipStrictValidation, "skip-validation", false, "It will skip any strict validation")
 	createConnectionCmd.Flags().StringSliceVarP(&connSecretFlag, "env", "e", nil, "The environment variables of the connection")
 	createConnectionCmd.Flags().StringSliceVar(&connTagsFlag, "tags", nil, "Tags to identify connections in a key=value format")
-	createConnectionCmd.Flags().StringVarP(&connAccessModeRunbooksFlag, "access-mode-runbooks", "", "enabled", "Enabled or disabled the access mode for runbooks (enabled/disabled)")
-	createConnectionCmd.Flags().StringVarP(&connAccessModeExecFlag, "access-mode-exec", "", "enabled", "Enabled or disabled the access mode for exec (enabled/disabled)")
-	createConnectionCmd.Flags().StringVarP(&connAccessModeConnectFlag, "access-mode-connect", "", "enabled", "Enabled or disabled the access mode for connect (enabled/disabled)")
+	createConnectionCmd.Flags().StringSliceVar(&connAccessModesFlag, "access-modes", nil, "Access modes enabled for this connection. Access modes are: runbooks, exec, connect")
 	createConnectionCmd.Flags().StringVarP(&connSchemaFlag, "schema", "", "", "Enabled or disabled the schema for this connection on the WebClient (enabled/disabled)")
 	createConnectionCmd.MarkFlagRequired("agent")
 }
@@ -127,6 +124,9 @@ var createConnectionCmd = &cobra.Command{
 			redactEnabled = true
 		}
 
+		access_mode_runbooks := verifyAccessModeStatus("runbooks")
+		access_mode_exec := verifyAccessModeStatus("exec")
+		access_mode_connect := verifyAccessModeStatus("connect")
 		connSchemaFlag = verifySchemaStatus(connSchemaFlag, connType)
 
 		connectionBody := map[string]any{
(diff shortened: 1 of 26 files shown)


client/cmd/admin/create-conn.go
```diff
@@ -34,6 +37,8 @@ func init() {
 	createConnectionCmd.Flags().BoolVar(&skipStrictValidation, "skip-validation", false, "It will skip any strict validation")
 	createConnectionCmd.Flags().StringSliceVarP(&connSecretFlag, "env", "e", nil, "The environment variables of the connection")
 	createConnectionCmd.Flags().StringSliceVar(&connTagsFlag, "tags", nil, "Tags to identify connections in a key=value format")
+	createConnectionCmd.Flags().StringSliceVar(&connAccessModesFlag, "access-modes", nil, "Access modes enabled for this connection. Access modes are: runbooks, exec, connect")
+	createConnectionCmd.Flags().StringVarP(&connSchemaFlag, "schema", "", "", "Enabled or disabled the schema for this connection on the WebClient (enabled/disabled)")
```
sandromello: - Use `StringVar()` if you're not using a short flag.
- grammar: "Enable or disable the schema for this connection on the WebClient. Accepted values: [disabled enabled]"

client/cmd/admin/create-conn.go
```diff
@@ -118,17 +123,27 @@ var createConnectionCmd = &cobra.Command{
 		if len(connRedactTypesFlag) > 0 {
 			redactEnabled = true
 		}
+
+		access_mode_runbooks := verifyAccessModeStatus("runbooks")
+		access_mode_exec := verifyAccessModeStatus("exec")
+		access_mode_connect := verifyAccessModeStatus("connect")
```
sandromello: With GO we use CamelCase to name variables.

client/cmd/admin/create-conn.go
```diff
@@ -179,6 +194,30 @@ var createConnectionCmd = &cobra.Command{
...
 
+func verifyAccessModeStatus(mode string) string {
+	if slices.Contains(connAccessModesFlag, mode) {
+		return "enabled"
+	} else if !slices.Contains(connAccessModesFlag, mode) {
+		return "disabled"
+	}
+
+	styles.PrintErrorAndExit("invalid access mode: %q", mode)
+	return "disabled"
```
sandromello: This is better and more cleaner:

```go
if slices.Contains(connAccessModesFlag, mode) {
  return "enabled"
}
return "disabled"
```

client/cmd/admin/create-conn.go
```diff
@@ -179,6 +194,30 @@ var createConnectionCmd = &cobra.Command{
...
+	if schema == "" && connType == "database" {
+		return "enabled"
+	} else if schema == "" {
+		return "disabled"
+	} else if schema == "enabled" || schema == "disabled" {
+		return schema
+	}
+
+	styles.PrintErrorAndExit("invalid schema status: %q", schema)
+	return "disabled"
```
sandromello: - Inform the user the valid options instead of just informing that something is incorrect.
Example: "invalid value for schema status %s, accepted values are: [enabled disabled]"

- The `return "disabled'` will never be returned, return an empty string instead

gateway/api/connections/connections.go
```diff
@@ -68,18 +72,23 @@ func Post(c *gin.Context) {
...
+		Type:               string(req.Type),
+		SubType:            req.SubType,
+		Envs:               coerceToMapString(req.Secrets),
+		Status:             req.Status,
+		ManagedBy:          nil,
+		Tags:               req.Tags,
+		AccessModeRunbooks: string(req.AccessModeRunbooks),
+		AccessModeExec:     string(req.AccessModeExec),
+		AccessModeConnect:  string(req.AccessModeConnect),
+		AccessSchema:       string(req.AccessSchema),
```
sandromello: Remove the type cast here, it isn't necessary anymore

gateway/pgrest/connections/connections.go
```diff
@@ -91,18 +95,31 @@ func (c *connections) Upsert(ctx pgrest.OrgContext, conn pgrest.Connection) erro
 	if conn.Status == "" {
 		conn.Status = pgrest.ConnectionStatusOffline
 	}
+
+	if conn.AccessSchema == "" {
+		if conn.Type == "database" {
+			conn.AccessSchema = "enabled"
+		} else {
+			conn.AccessSchema = "disabled"
+		}
```
sandromello: There's no need for the else statement here, this is more cleaner and easier to read:

```go
conn.AccessSchema = "disabled"
if conn.Type == "database" {
  conn.AccessSchema = "enabled"
}
```

gateway/transport/interceptors/auth/auth.go
```diff
@@ -168,15 +168,19 @@ func (i *interceptor) getConnection(name string, userCtx *pguserauth.Context) (*
...
+		SubType:            conn.SubType,
+		CmdEntrypoint:      conn.Command,
+		Secrets:            conn.AsSecrets(),
+		AgentID:            conn.AgentID,
+		AgentMode:          conn.Agent.Mode,
+		AgentName:          conn.Agent.Name,
+		AccessModeRunbooks: string(conn.AccessModeRunbooks),
+		AccessModeExec:     string(conn.AccessModeExec),
+		AccessModeConnect:  string(conn.AccessModeConnect),
+		AccessSchema:       string(conn.AccessSchema),
```
sandromello: There's no need for type casting a string.

gateway/transport/server.go
```diff
@@ -106,9 +106,26 @@ func (s *Server) PreConnect(ctx context.Context, req *pb.PreConnectRequest) (*pb
 	return resp, nil
 }
 
+func GetAccessModesFromConnect(clientVerb string, clientOrigin string) string {
+	switch {
+	case clientVerb == pb.ClientVerbExec && clientOrigin == pb.ConnectionOriginClient:
+		return "exec"
+	case clientVerb == pb.ClientVerbExec && clientOrigin == pb.ConnectionOriginClientAPI:
+		return "exec"
```
sandromello: Since you're returning the same value for distinct statements in a case block, just use comma to separate expressions and return once.

Example:

```go
switch {
	case clientVerb == pb.ClientVerbExec && clientOrigin == pb.ConnectionOriginClient,
               clientVerb == pb.ClientVerbExec && clientOrigin == pb.ConnectionOriginClientAPI:
		return "exec"
         (...)
```

gateway/transport/server.go
```diff
@@ -167,6 +184,20 @@ func (s *Server) Connect(stream pb.Transport_ConnectServer) (err error) {
...
+	currentAccessMode := GetAccessModesFromConnect(clientVerb[0], clientOrigin[0])
+
+	AccessModes := map[string]string{
+		"exec":     gwctx.Connection.AccessModeExec,
+		"connect":  gwctx.Connection.AccessModeConnect,
+		"runbooks": gwctx.Connection.AccessModeRunbooks}
+
+	if AccessModes[currentAccessMode] == "disabled" {
+		return status.Error(codes.FailedPrecondition,
+			fmt.Sprintf("the %v access mode connection has the %v feature disabled", gwctx.Connection.Name, currentAccessMode))
```
sandromello: The error message seems a bit confuse. Let's keep the same error from the proxy manager:

```go
fmt.Sprintf("the %s connection has the access mode %v feature disabled", gwctx.Connection.Name, currentAccessMode))
```

sandromello (changes requested):

sandromello (approved):

PR #444, opened 2024-08-20
Update Segment Identify Event Handler

andriosrobert:
Identify user when email is first available

### DIFF
--- a/webapp/src/webapp/auth/views/signup.cljs
+++ b/webapp/src/webapp/auth/views/signup.cljs
@@ -49,6 +49,7 @@
                            (.preventDefault e)
                            (reset! loading (not @loading))
                            (rf/dispatch [:segment->track "SignUp - Create organization"])
+                           (rf/dispatch [:segment->identify])
                            (rf/dispatch [:auth->signup @organization-name @user-name]))}
              (when login-error
                [:div {:class "pb-small"}
--- a/webapp/src/webapp/auth/views/signup.cljs
+++ b/webapp/src/webapp/auth/views/signup.cljs
@@ -49,7 +49,7 @@
                            (.preventDefault e)
                            (reset! loading (not @loading))
                            (rf/dispatch [:segment->track "SignUp - Create organization"])
-                           (rf/dispatch [:segment->identify])
+                           (rf/dispatch [:segment->identify] (:data @userinfo)
                            (rf/dispatch [:auth->signup @organization-name @user-name]))}
              (when login-error
                [:div {:class "pb-small"}
--- a/webapp/src/webapp/events/segment.cljs
+++ b/webapp/src/webapp/events/segment.cljs
@@ -15,12 +15,11 @@
 
 (rf/reg-event-fx
  :segment->identify
- (fn [{:keys [db]} _]
-   (let [user (-> db :users->current-user :data)
-         user-id (:id user)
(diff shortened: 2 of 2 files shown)


rogefm:
Andrios is the opposite, this does not guarantee that the user's data is fresh, the other way did guarantee it, because now you are fetching the user we already have in the state. Before, we first updated the user's state and then proceeded, ensuring the updated user. 
Was there any place where you felt a difference with this implementation?

webapp/src/webapp/auth/views/signup.cljs
```diff
@@ -49,6 +49,7 @@
                            (.preventDefault e)
                            (reset! loading (not @loading))
                            (rf/dispatch [:segment->track "SignUp - Create organization"])
+                           (rf/dispatch [:segment->identify])
```
rogefm: And here you only need to pass the userinfo
```suggestion
                           (rf/dispatch [:segment->identify] (:data @userinfo)
```

andriosrobert:
> Andrios is the opposite, this does not guarantee that the user's data is fresh, the other way did guarantee it, because now you are fetching the user we already have in the state. Before, we first updated the user's state and then proceeded, ensuring the updated user. Was there any place where you felt a difference with this implementation?

@rogefm this was a bad description, my main goal here was to trigger the identify call, the potential improvement is secondary.

andriosrobert:
And from what I understood the previous identify implementation didn't use the user information, it had only the user id pulled from the user received as a parameter

also: I didn't find calls to the identify method, I don't think it was being used before

rogefm:
> And from what I understood the previous identify implementation didn't use the user information, it had only the user id pulled from the user received as a parameter
> 
> also: I didn't find calls to the identify method, I don't think it was being used before

Yes they do! here:
`(rf/reg-event-fx
 :segment->identify
 (fn [{:keys [db]} [_ user]]
   (let [user-id (:id user)
         analytics (-> db :segment->analytics)]
     (if (nil? analytics)
       {:fx [[:dispatch [:segment->load [:segment->identify user]]]]}
       (do
         (.identify analytics user-id (clj->js user))   <-------
         {})))))`

rogefm:
> > Andrios is the opposite, this does not guarantee that the user's data is fresh, the other way did guarantee it, because now you are fetching the user we already have in the state. Before, we first updated the user's state and then proceeded, ensuring the updated user. Was there any place where you felt a difference with this implementation?
> 
> @rogefm this was a bad description, my main goal here was to trigger the identify call, the potential improvement is secondary.

Cool, so can we keep it the same as before? The improvement can provide false data if the user is updated any time ago.

webapp/src/webapp/auth/views/signup.cljs
```diff
@@ -49,6 +49,7 @@
                            (.preventDefault e)
                            (reset! loading (not @loading))
                            (rf/dispatch [:segment->track "SignUp - Create organization"])
+                           (rf/dispatch [:segment->identify] (:data @userinfo)
```
andriosrobert, line 52: @rogefm do we have userinfo at this point? note that it is right before creating the org. the main thing I'm looking for here is the email to be present

andriosrobert (commented):
I rolled back the identify change

rogefm (approved):

PR #446, opened 2024-08-20
MS Presidio as DLP option

mtmr0x:
Add MS Presidio to terminal exec, console connection types.

### DIFF
--- a/agent/controller/agent.go
+++ b/agent/controller/agent.go
@@ -199,6 +199,18 @@ func (a *Agent) processSessionOpen(pkt *pb.Packet) {
 		}
 	}
 
+	if a.connStore.Get(msPresidioAnonymizerURLKey) == nil {
+		if anonymizerURL, ok := pkt.Spec[pb.SpecAgentMSPresidioAnonymizerURL]; ok {
+			a.connStore.Set(msPresidioAnonymizerURLKey, anonymizerURL)
+		}
+	}
+
+	if a.connStore.Get(msPresidioAnalyzerURLKey) == nil {
+		if analyzerURL, ok := pkt.Spec[pb.SpecAgentMSPresidioAnalyzerURL]; ok {
+			a.connStore.Set(msPresidioAnalyzerURLKey, analyzerURL)
+		}
+	}
+
 	go func() {
 		if err := a.checkTCPLiveness(pkt, connParams.EnvVars); err != nil {
 			_ = a.client.Send(&pb.Packet{
@@ -436,6 +448,18 @@ func (a *Agent) getGCPCredentials() string {
 
 func (a *Agent) getDlpProvider() string {
 	obj := a.connStore.Get(dlpProviderKey)
+	v, _ := obj.([]byte)
+	return string(v)
+}
+
+func (a *Agent) getMSPresidioAnalyzerURL() string {
+	obj := a.connStore.Get(msPresidioAnalyzerURLKey)
+	v, _ := obj.([]byte)
+	return string(v)
+}
+
+func (a *Agent) getMSPresidioAnonymizerURL() string {
+	obj := a.connStore.Get(msPresidioAnonymizerURLKey)
 	v, _ := obj.([]byte)
 	return string(v)
 }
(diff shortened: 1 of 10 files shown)


webapp/package-lock.json
sandromello, line 1: Why are you changing this file?

gateway/appconfig/appconfig.go
```diff
@@ -89,19 +98,22 @@ func Load() error {
...
+		apiHost:                 apiRawURL.Host,
+		askAICredentials:        askAICred,
+		pgCred:                  pgCred,
+		migrationPathFiles:      migrationPathFiles,
+		licenseSigningKey:       licensePrivKey,
+		licenseSignerOrgID:      allowedOrgID,
+		gcpDLPJsonCredentials:   gcpJsonCred,
+		dlpProvider:             dlpProvider,
+		msPresidioAnalyzerURL:   msPresidioAnalyzerURL,
+		msPresidioAnonymizerURL: msPresidioAnonymizerURL,
```
sandromello: You can pass the `os.Getenv("<ENV>")` directly, in case it doesn't exists the default value will be an empty string. There's no need to create a function to parse those values.

agent/controller/postgres.go
```diff
@@ -49,14 +49,17 @@ func (a *Agent) processPGProtocol(pkt *pb.Packet) {
...
-		"dlp_info_types":        strings.Join(connParams.DLPInfoTypes, ","),
-		"dlp_masking_character": "#",
+		"hostname":                  connenv.host,
+		"port":                      connenv.port,
+		"username":                  connenv.user,
+		"password":                  connenv.pass,
+		"sslmode":                   connenv.postgresSSLMode,
+		"dlp_provider":              a.getDlpProvider(),
+		"mspresidio_analyzer_url":   a.getMSPresidioAnalyzerURL(),
+		"mspresidio_anonymizer_url": a.getMSPresidioAnonymizerURL(),
```
sandromello: Remove these values, they are not being used in the Postgres implementation

agent/controller/agent.go
```diff
@@ -193,6 +193,24 @@ func (a *Agent) processSessionOpen(pkt *pb.Packet) {
...
+		if anonymizerURL, ok := pkt.Spec[pb.SpecAgentMSPresidioAnonymizerURL]; ok {
+			a.connStore.Set(msPresidioAnonymizerURLKey, anonymizerURL)
+		}
+	}
+
+	if a.connStore.Get(msPresidioAnalyzerURLKey) == nil {
+		if analyzerURL, ok := pkt.Spec[pb.SpecAgentMSPresidioAnalyzerURL]; ok {
+			a.connStore.Set(msPresidioAnalyzerURLKey, analyzerURL)
+		}
+	}
```
sandromello, line 212: TIP: you can save them in a single struct using a single key, but it's ok to leave it like for now.

sandromello (changes requested):

agent/controller/postgres.go
mtmr0x: my bad, when first navigating in the code, I added then here, then saw it was a mistake and never fixed

webapp/package-lock.json
mtmr0x: this file is auto generated when building the webapp, as I didn't performed changes at it, I will just checkout from the PR

mtmr0x: This file doesn't do any harm, and we can safely publish it. It's a mess, but it helps solve dependency problems when coding in nodejs. It will be all the time rebuilt

mtmr0x: also, the weird thing I never bought, it's actually recommended to track and commit this file always.

node ¯\_(ツ)_/¯

rogefm: Pay attention to this file @mtmr0x, because you are using an outdated package-lock, which is a little bit dangerous.
We always do an `npm install` in our build, but you should verify your source code and check this behavior as soon as possible.

sandromello (approved):

PR #453, opened 2024-08-27
chore: Create maxGroupsSize logic in SlackService

andriosrobert:
image (2)


sandromello (changes requested):
This change will also impact the size of the script which could lead to more complain from users. I think it's best to create a different constant or apply the truncate logic directly in the group list.

sandromello:
LGTM!  @andriosrobert just make sure to change the PR to "Ready for review" if you want to merge it.

andriosrobert:
@sandromello still see merging is blocked

sandromello (approved):

PR #454, opened 2024-08-27
feat: Add API key authentication middleware

andriosrobert:
# Add API Key Authentication for Connection Management

## Summary
This PR introduces API key authentication for creating and deleting connections. This feature allows for automated connection management while maintaining security.

## Changes
- Added an `ApiKeyAuthMiddleware` to validate API keys for specific routes
- Implemented API key validation using an environment variable (`ORG_API_KEY`)
- Applied the new middleware to POST and DELETE connection routes

## Security Considerations
While not the most secure method, this implementation offers a balance between functionality and security:

1. **Optional Feature**: Customers can choose not to enable this feature by simply not setting the `ORG_API_KEY` environment variable.

2. **Limited Scope**: The API key authentication only works for creating or deleting connections. It cannot be used to read any information, sensitive or otherwise, making it relatively harmless from a data security standpoint.

3. **Potential for Service Disruption**: The main risk is the potential deletion of connections, which could disrupt service. However, this risk is mitigated by:
   - The requirement of both admin authentication and the API key
   - The ability to easily restore connections using database backups

4. **No Data Exposure**: This feature does not expose any sensitive data, as it's limited to connection management operations.

## Usage
To enable this feature, set the `ORG_API_KEY` environment variable with a secure, randomly generated key. Use this key in the `X-API-Key` header when making POST or DELETE requests to the connections endpoints.

## Testing
- [ ] Verify that connection creation works with a valid API key
- [ ] Verify that connection deletion works with a valid API key
- [ ] Confirm that these operations fail without a valid API key
- [ ] Ensure that other endpoints are not affected by this change

### DIFF
--- a/gateway/api/server.go
+++ b/gateway/api/server.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -221,6 +222,7 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 	route.PUT("/connections/:nameOrID",
 		AdminOnlyAccessRole,
 		api.Authenticate,
+		apiKeyAuth,
 		api.TrackRequest(analytics.EventUpdateConnection),
 		AuditApiChanges,
 		apiconnections.Put)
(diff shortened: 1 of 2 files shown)


sandromello (changes requested):
@andriosrobert , this won't work because the middleware of api.Authenticate. You need to perform this conditionally in this middleware if you want to introduce this feature.

andriosrobert:
@sandromello sorry, it was supposed to be a draft, I'm still working on it, thanks for the review though

andriosrobert:
it is in draft, I'll ping you once it is ready

PR #458, opened 2024-08-31
[FEAT]: Dashboard page

rogefm:


sandromello (approved):

webapp/src/webapp/subs.cljs
```diff
@@ -358,3 +359,28 @@
  :reports->session
  (fn [db _]
    (:reports->session db)))
+
+(re-frame/reg-sub
+ :reports->redact-data-by-date
+ (fn [db _]
+   (:reports->redact-data-by-date db)))
```
mtmr0x: it's a minor, but still helps readability to have the `redact` in its verbal time right as `redacted` across this subs and events at the `reports/cljs`. Saying `redact` means an action, versus what this demonstrate, which is redacted information

mtmr0x (changes requested):
I have this naming comment only to enforce readability

webapp/src/webapp/events/reports.cljs
```diff
@@ -12,14 +28,136 @@
...
 (rf/reg-event-fx
- ::reports->set-session
+ ::reports->set-report-by-session-id
  (fn
    [{:keys [db]} [_ report]]
    {:db (assoc db :reports->session {:status :ready
                                      :data report})}))
 
+(rf/reg-event-fx
+ :reports->get-redact-data-by-date
```
mtmr0x: sorry if I wasn't very clear on my last comment, but I meant across the events like this one too

mtmr0x: all of the events in this file actually, to change to `redacted`

webapp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webapp",
-  "version": "1.26.0",
+  "version": "1.27.1",
```
mtmr0x: As we didn't released anything yet, we can have this as 1.27.0 instead of bumping the version during the PR time

rogefm: It's complicated because when we create a new commit after changing the package.json, it becomes difficult to track the deployment version. So I "always" (if I finished the feature before the other commit) prefer creating a new version in the package.json for the commits.

mtmr0x (approved):

PR #463, opened 2024-09-05
Local Auth and beta api token implementation

mtmr0x:


.env.sample
```diff
@@ -17,6 +17,13 @@ LICENSE_SIGNING_KEY=
 
 # auth configuration
 API_URL=http://localhost:8009
+# If you don't set a IDP and want to go with AUTH_METHOD as local, you must provide a JWK_KEY.
+AUTH_METHOD=
+# Generate a random string and place it here. This key shall not be lost or changed.
+# If so, all the tokens generated with it will be invalid and all users will need
+# to login again. This value is only necessary if AUTH_METHOD is set to local.
+JWT_KEY=
```
mtmr0x: jwt_secret_key

gateway/api/localauth/login.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/google/uuid"
+	"github.com/hoophq/hoop/gateway/appconfig"
+	"github.com/hoophq/hoop/gateway/pgrest"
+	pglocalauthsession "github.com/hoophq/hoop/gateway/pgrest/localauthsession"
+	pgusers "github.com/hoophq/hoop/gateway/pgrest/users"
+	"golang.org/x/crypto/bcrypt"
+)
+
+type Claims struct {
```
mtmr0x: put this on a types file

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+		if totalUsers > 0 {
+			return "", fmt.Errorf("You can not access this instance. Please contact your administrator")
+		}
+
+		return org.ID, nil
+	}
+}
+
+func Register(c *gin.Context) {
+	fmt.Printf("Register\n")
```
mtmr0x: print

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+		Token:     tokenString,
+		ExpiresAt: expirationTime.Format(time.RFC3339),
+	}
+	_, err = pglocalauthsession.CreateSession(newLocalAuthSession)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to store session"})
+		return
+	}
+
+	// TODO: generate token and return it
```
mtmr0x: authorize this header below for frontend to save session

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
+	tokenString = strings.Replace(tokenString, "Bearer ", "", 1)
+	if tokenString == "" {
+		c.JSON(http.StatusUnauthorized, gin.H{"error": "Missing authorization token"})
+		c.Abort()
+		return
+	}
+
+	jwtKey := appconfig.Get().JWTSecretKey()
+	claims := &localauthapi.Claims{}
+	token, err := jwt.ParseWithClaims(tokenString, claims, func(token *jwt.Token) (interface{}, error) {
```
mtmr0x, line 60: this already validates expiration

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
+		return
+	}
+
+	sessionByToken, err := pglocalauthsession.GetSessionByToken(tokenString)
 	if err != nil {
-		tokenHeader := c.GetHeader("authorization")
-		log.Infof("failed authenticating, %v, length=%v, reason=%v",
-			parseHeaderForDebug(tokenHeader), len(tokenHeader), err)
-		c.AbortWithStatus(http.StatusUnauthorized)
+		fmt.Printf("err: %v\n", err)
```
mtmr0x: print

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
+		c.JSON(http.StatusUnauthorized, gin.H{"error": "Invalid or expired session"})
+		return
+	}
+
+	// TODO change ExpiresAt at the database for date with timezone
```
mtmr0x: remove comment

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+	case "local":
+		a.localAuthMiddleware(c)
+	case "api-key":
+		allowed, _ := c.Get("allow-api-key")
+		fmt.Printf("allowed: %v\n", allowed)
+		if allowed != nil && allowed.(bool) {
+			a.apiKeyMiddleware(c)
+			return
+		}
+		c.AbortWithStatus(http.StatusUnauthorized)
```
mtmr0x, line 177: see if there we need a return

gateway/security/idp/authenticator.go
```diff
@@ -223,7 +227,9 @@ func setProviderConfFromEnvs(p *Provider) error {
 	}
 	p.authWithUserInfo = issuerURL.Query().Get("_userinfo") == "1"
 	if p.ClientSecret == "" || p.ClientID == "" {
-		return fmt.Errorf("missing IDP credentials: IDP_CLIENT_ID, IDP_CLIENT_SECRET")
+		// defaults to local auth
+		return nil
```
mtmr0x, line 230: review this

webapp/src/webapp/subs.cljs
```diff
@@ -384,3 +389,9 @@
  :reports->today-session-data
  (fn [db _]
    (:reports->today-session-data db)))
+
+(re-frame/reg-sub
+ :auth->auth-method
+ (fn [db _]
+   (:auth->auth-method db)))
+
```
mtmr0x: this is trash, delete

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+	apiKeyReq := c.GetHeader("Api-Key")
+	// split apiKey by | getting the first and the second part
+	// the first part is the orgID and the second part is the apiKey
+	orgID := strings.Split(apiKeyReq, "|")[0]
+	// apiKey := strings.Split(apiKeyReq, "|")[1]
+	org, err := pgorgs.New().FetchOrgByID(orgID)
+	if err != nil || org == nil {
+		c.AbortWithStatus(http.StatusUnauthorized)
+		return
+	}
```
sandromello: Relying on the organization ID inside the API KEY is not secure. The key is configured as an environment variable and you're trying to obtain the organization from the database. If a key leaks, it could be used to impersonate other organizations.

If this is intentional for **self-hosted installations**, deny the utilization of API KEYS for multi-tenant setups.

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+
+	if apiKeyReq == configuredApiKey {
```
sandromello, line 113: I recommend validating the key before hitting the database. It avoids a roundtrip to the database when the key is not valid.

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+	switch authMethod {
+	case "local":
+		a.localAuthMiddleware(c)
+	case "api-key":
+		allowed, _ := c.Get("allow-api-key")
+		fmt.Printf("allowed: %v\n", allowed)
+		if allowed != nil && allowed.(bool) {
+			a.apiKeyMiddleware(c)
+			return
+		}
```
sandromello: Rewrite to:


```go
if _, ok := c.Get("allow-api-key"); ok {
  a.apiKeyMiddleware(c)
  return
}
```

gateway/api/middleware.go
sandromello: there's no need

gateway/api/user/user.go
```diff
@@ -62,12 +63,27 @@ func Create(c *gin.Context) {
...
+	// why we create this separated variable so we can modify it
+	// accordingly to the auth method
+	userSubject := newUser.Email
+	if appconfig.Get().AuthMethod() == "local" {
+		hashedPassword, err := bcrypt.GenerateFromPassword([]byte(newUser.Password), bcrypt.DefaultCost)
+		newUser.Password = string(hashedPassword)
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to hash password"})
+			return
+		}
```
sandromello: Log the reason (errorf) in the logs as well.

gateway/api/user/user.go
```diff
@@ -62,12 +63,27 @@ func Create(c *gin.Context) {
...
+	}
+
 	newUser.Verified = false
 	pguser := pgrest.User{
 		ID:       newUser.ID,
-		Subject:  newUser.Email,
+		Subject:  userSubject,
 		OrgID:    ctx.OrgID,
 		Name:     newUser.Name,
+		Password: newUser.Password,
```
sandromello: Having an explicit name in the database and on structs would be better to identify that this field is hashed just by reading it. Example: `HashedPassword`

gateway/pgrest/localauthsession/localauthsession.go
```diff
@@ -0,0 +1,26 @@
+package pglocalauthsession
```
sandromello: As we discussed: consider validating the jwt token instead of relying on a local state table

gateway/pgrest/orgs/helpers.go
```diff
@@ -23,8 +23,7 @@ func CreateDefaultOrganization() (pgrest.OrgContext, error) {
 			return nil, fmt.Errorf("failed creating the default organization, err=%v", err)
 		}
 		return pgrest.NewOrgContext(orgID), nil
-	case len(orgList) == 1:
+	default:
 		return pgrest.NewOrgContext(orgList[0].ID), nil
```
sandromello: The problem with this change is that you leave it to potential issues if the installation has multiple organizations.
I recommend changing this behavior in a separate PR that would introduce multiple organizations for single-tenant setups.

gateway/pgrest/orgs/orgs.go
```diff
@@ -55,10 +56,11 @@ func (o *org) CreateOrGetOrg(name string, licenseDataJSON []byte) (orgID string,
 		return org.ID, nil
 	}
 	orgID = uuid.NewString()
-	licenseData, err := decodeLicenseToMap(licenseDataJSON)
-	if err != nil {
-		return "", fmt.Errorf("unable to encode license data properly: %v", err)
-	}
+	licenseData, _ := decodeLicenseToMap(licenseDataJSON)
```
sandromello: We need to return an error if the license has any data. Add a condition to fix this:
```go
var licenseData []byte
if len(licenseDataJSON) > 0 {
  licenseData, err = decodeLicenseToMap(licenseDataJSON)
  if err != nil {
    return "", fmt.Errorf("unable to encode license data properly: %v", err)
  }
}
```

gateway/pgrest/orgs/orgs.go
```diff
@@ -93,6 +95,19 @@ func (o *org) FetchOrgByName(name string) (*pgrest.Org, int64, error) {
 	return &org, total, nil
 }
 
+// FetchOrgByID returns an organization
+func (o *org) FetchOrgByID(id string) (*pgrest.Org, error) {
```
sandromello: You don't need this additional method, just use the method below (`FetchOrgByContext`) instead by creating a new context with the organization ID string:
```go
orgCtx := pgrest.NewOrgContext(orgID)
_, _ = pgorgs.New().FetchOrgByContext(orgCtx)
```

gateway/pgrest/userauth/userauth.go
```diff
@@ -60,7 +60,7 @@ func (u *userAuth) FetchUserContext(subject string) (*Context, error) {
 }
 
 func fetchOneBySubject(subject string) (*pgrest.User, error) {
-	path := fmt.Sprintf("/users?select=*,groups,orgs(id,name,license,license_data)&subject=eq.%v&verified=is.true", subject)
+	path := fmt.Sprintf("/users?select=*,groups,orgs(id,name,license,license_data)&subject=eq.%v", subject)
```
sandromello: Let's not change that for now. Just set the user as verified for local auth method.

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
-func (a *Api) Authenticate(c *gin.Context) {
-	roleName := RoleFromContext(c)
-	subject, err := a.validateAccessToken(c)
+func (a *Api) localAuthMiddleware(c *gin.Context) {
+	tokenString := c.GetHeader("Authorization")
+	// remove bearer from token
+	tokenString = strings.Replace(tokenString, "Bearer ", "", 1)
+	if tokenString == "" {
+		c.JSON(http.StatusUnauthorized, gin.H{"error": "Missing authorization token"})
+		c.Abort()
```
sandromello: You're already returning unauthorized, remove this line.

gateway/api/middleware.go
sandromello: You're not using any claims, use the function Parse instead.

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
+
+	jwtKey := appconfig.Get().JWTSecretKey()
+	claims := &localauthapi.Claims{}
+	token, err := jwt.ParseWithClaims(tokenString, claims, func(token *jwt.Token) (interface{}, error) {
+		return jwtKey, nil
+	})
+
+	if err != nil || !token.Valid {
+		c.JSON(http.StatusUnauthorized, gin.H{"error": "Invalid or expired token"})
+		c.Abort()
```
sandromello: There's no need for this abort, already calling unauthorized above

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
+	sessionExpiresAt, err := time.Parse("2006-01-02T15:04:05+00:00", sessionByToken.ExpiresAt)
+
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "Error processing session"})
+		c.Abort()
+		return
+	}
+	if time.Now().After(sessionExpiresAt) {
+		c.JSON(http.StatusUnauthorized, gin.H{"error": "Session has expired"})
+		c.Abort()
```
sandromello: There's no need to validate the token `exp` claim. It's already checked when using the `Parse` method.

gateway/api/middleware.go
```diff
@@ -26,23 +32,62 @@ import (
...
 	}
 
-	ctx, err := pguserauth.New().FetchUserContext(subject)
+	user, err := pgusers.GetOneByEmail(sessionByToken.UserEmail)
+	if err != nil {
+		log.Errorf("failed fetching user, subject=%v, err=%v", user.Subject, err)
+		c.AbortWithStatus(http.StatusUnauthorized)
+		return
+	}
+	ctx, err := pguserauth.New().FetchUserContext(user.Subject)
```
sandromello: You can get the subject straight from the token claims. It's safer after it's verified by the Parse function.

gateway/security/idp/authenticator.go
mtmr0x: yes, we need this for when local auth is set, we these values are not set

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
-		if !ctx.IsEmpty() {
-			break
+
+	c.Set(storagev2.ContextKey,
+		storagev2.NewContext(ctx.UserSubject, ctx.OrgID).
+			WithUserInfo(ctx.UserName, ctx.UserEmail, string(ctx.UserStatus), ctx.UserPicture, ctx.UserGroups).
+			WithSlackID(ctx.UserSlackID).
+			WithOrgName(ctx.OrgName).
+			WithOrgLicense(ctx.OrgLicense).
+			// WithApiURL(a.IDProvider.ApiURL).
```
sandromello: Is there any reason to not propagate this data?

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+			WithGrpcURL(a.GrpcURL),
+	)
+
+	c.Next()
+}
+
+// AllowApiKey is a middleware that allows the request to proceed if the request has an api key.
+// It sets the "allow-api-key" key in the context to true so the Auth Middleware can authorize it
+func (a *Api) AllowApiKey(c *gin.Context) {
+	c.Set("allow-api-key", true)
```
sandromello: Use a constant value for this.

gateway/api/middleware.go
```diff
@@ -26,23 +31,48 @@ import (
...
+		return
+	}
+
+	sessionByToken, err := pglocalauthsession.GetSessionByToken(tokenString)
 	if err != nil {
-		tokenHeader := c.GetHeader("authorization")
-		log.Infof("failed authenticating, %v, length=%v, reason=%v",
-			parseHeaderForDebug(tokenHeader), len(tokenHeader), err)
-		c.AbortWithStatus(http.StatusUnauthorized)
+		fmt.Printf("err: %v\n", err)
```
rogefm:
```suggestion
		log.Errorf("err: %v\n", err)
```

Also describe better the error

gateway/api/middleware.go
```diff
@@ -51,70 +81,187 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+	case "api-key":
+		allowed, _ := c.Get("allow-api-key")
+		if allowed != nil && allowed.(bool) {
+			a.apiKeyMiddleware(c)
+			return
+		}
+		c.AbortWithStatus(http.StatusUnauthorized)
+	default:
+		roleName := RoleFromContext(c)
+		fmt.Printf("roleName: %v\n", roleName)
```
rogefm: print

gateway/api/middleware.go
```diff
@@ -51,70 +81,187 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+		subject, err := a.validateAccessToken(c)
+		fmt.Printf("subject: %v\n", subject)
```
rogefm: print

gateway/api/middleware.go
```diff
@@ -51,70 +81,187 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+		if err != nil {
+			tokenHeader := c.GetHeader("authorization")
+			log.Infof("failed authenticating, %v, length=%v, reason=%v",
+				parseHeaderForDebug(tokenHeader), len(tokenHeader), err)
+			c.AbortWithStatus(http.StatusUnauthorized)
+			return
+		}
+
+		ctx, err := pguserauth.New().FetchUserContext(subject)
+		fmt.Printf("ctx: %+v\n", ctx)
```
rogefm: print

gateway/api/server.go
```diff
@@ -148,6 +150,15 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
 	route.GET("/login", loginHandler.Login)
 	route.GET("/callback", loginHandler.LoginCallback)
 	route.GET("/healthz", apihealthz.LivenessHandler())
+	///////////////////////
+	// local auth routes //
+	///////////////////////
+	// TODO do not allow these routes be called if auth is not set as local
```
rogefm: Was this TODO made? Sounds important

gateway/pgrest/orgs/orgs.go
```diff
@@ -55,10 +56,11 @@ func (o *org) CreateOrGetOrg(name string, licenseDataJSON []byte) (orgID string,
...
 	orgID = uuid.NewString()
-	licenseData, err := decodeLicenseToMap(licenseDataJSON)
-	if err != nil {
-		return "", fmt.Errorf("unable to encode license data properly: %v", err)
-	}
+	licenseData, _ := decodeLicenseToMap(licenseDataJSON)
+	log.Debugf("licenseData: %v", licenseData)
+	// if err != nil {
+	// 	return "", fmt.Errorf("unable to encode license data properly: %v", err)
+	// }
```
rogefm:
```suggestion

```
Is this comment important?

gateway/pgrest/users/users.go
```diff
@@ -19,17 +19,33 @@ type UserGroups struct {
...
 type user struct{}
 
+func GetOneByEmail(email string) (*pgrest.User, error) {
+	var usr pgrest.User
+	err := pgrest.New("/users?email=eq.%v", url.QueryEscape(email)).
+		FetchOne().
+		DecodeInto(&usr)
+	if err != nil {
+		if err == pgrest.ErrNotFound {
+			fmt.Printf("user not found")
```
rogefm: print

webapp/src/webapp/auth/local/login.cljs
```diff
@@ -0,0 +1,51 @@
...
+       [:> Flex {:direction "column"}
+        [forms/input {:label "Email"
+                      :value @email
+                      :type "email"
+                      :on-change #(reset! email (-> % .-target .-value))}]
+        [forms/input {:label "Password"
+                      :value @password
+                      :type "password"
+                      :on-change #(reset! password (-> % .-target .-value))}]
+        [button/tailwind-primary
```
rogefm: here you can use the button of Radix default color indigo.
Example:
`[:> Button {:color "indigo"
                :size "2"
                :variant "solid"
                :radius "medium"}
     "Login"]`

webapp/src/webapp/auth/local/register.cljs
```diff
@@ -0,0 +1,62 @@
...
+                      :on-change #(reset! email (-> % .-target .-value))}]
+        [forms/input {:label "Password"
+                      :value @password
+                      :type "password"
+                      :on-change #(reset! password (-> % .-target .-value))}]
+        [forms/input {:label "Confirm Password"
+                      :value @confirm-password
+                      :type "password"
+                      :on-change #(reset! confirm-password (-> % .-target .-value))}]
+        [button/tailwind-primary
```
rogefm: Here as well.

gateway/transport/interceptors/auth/auth.go
```diff
@@ -125,33 +131,109 @@ func (i *interceptor) StreamServerInterceptor(srv any, ss grpc.ServerStream, inf
...
+				sessionExpiresAt, err := time.Parse("2006-01-02T15:04:05+00:00", sessionByToken.ExpiresAt)
+
+				if err != nil {
+					log.Debugf("failed verifying access token, reason=%v", err)
+					return status.Errorf(codes.Unauthenticated, "invalid authentication")
+				}
+				if time.Now().After(sessionExpiresAt) {
+					log.Debugf("failed verifying access token, reason=%v", err)
+					return status.Errorf(codes.Unauthenticated, "invalid authentication")
+				}
```
sandromello: There's no need to validate the session expiration if you use the `Parse` method to validate the token.

gateway/transport/interceptors/auth/auth.go
```diff
@@ -125,33 +131,109 @@ func (i *interceptor) StreamServerInterceptor(srv any, ss grpc.ServerStream, inf
...
+			connectionName := commongrpc.MetaGet(md, "connection-name")
+			conn, err := i.getConnection(connectionName, userCtx)
+			if err != nil {
+				return err
+			}
+			if conn == nil {
+				return status.Errorf(codes.NotFound, "connection not found")
+			}
+			gwctx.Connection = *conn
+			ctxVal = gwctx
```
sandromello: This part of the code is repeated, use it once by putting in the outer block of the condition

gateway/api/middleware.go
mtmr0x: this was removed already

mtmr0x: thanks

gateway/transport/interceptors/auth/auth.go
```diff
@@ -125,33 +131,109 @@ func (i *interceptor) StreamServerInterceptor(srv any, ss grpc.ServerStream, inf
...
-			return err
-		}
-		if conn == nil {
-			return status.Errorf(codes.NotFound, "connection not found")
+			if conn == nil {
+				return status.Errorf(codes.NotFound, "connection not found")
+			}
+			gwctx.Connection = *conn
+			ctxVal = gwctx
+		} else {
```
sandromello: Remove this else block by adding a return in the if block above. It's unnecessary and hard to read it.

gateway/pgrest/orgs/orgs.go
mtmr0x: noice

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+		c.AbortWithStatus(http.StatusUnauthorized)
+		return
+	}
+
+	if apiKeyReq == configuredApiKey {
+		ctx := &pguserauth.Context{
+			OrgID:       orgID, // "400342d2-fa79-4a05-aa8a-b4d4d1989acb",
+			OrgName:     org.Name,
+			OrgLicense:  org.License,
+			UserUUID:    "API_KEY",
```
sandromello: I recommend using a deterministic UUID for this one, I'm not sure how this information is used in the underline layers. We usually rely on UUID type in the database.

To generate a deterministic UUID based on that name:
```go
uuidStr := uuid.NewSHA1(uuid.NameSpaceURL, []byte(`API_KEY`))
```

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+	err = pgusers.New().Upsert(pgrest.User{
+		ID:       userID,
+		Subject:  fmt.Sprintf("local|%v", userID),
+		OrgID:    newOrgID,
+		Email:    user.Email,
+		Name:     user.Name,
+		Status:   "active",
+		Password: string(hashedPassword),
+		Groups:   []string{adminGroupName},
+	})
```
sandromello: Add `Verified: true`, it's required to be considered a valid user

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+		// the default organization. Instead, they should get an invitation
+		if totalUsers > 0 {
+			return "", fmt.Errorf("You can not access this instance. Please contact your administrator")
+		}
+
+		return org.ID, nil
+	}
+}
+
+func Register(c *gin.Context) {
```
sandromello, line 22: We need to be more defensive here and checking if the secret key is set (!= "") before allowing this route to be used.

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+	if err := c.BindJSON(&user); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+	fmt.Printf("Registering user %+v", user)
+
+	log.Debugf("looking for existing user %v", user.Email)
+	// fetch user by email
+	existingUser, err := pgusers.GetOneByEmail(user.Email)
+	if existingUser != nil {
```
sandromello, line 36: Check the error first.

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
+package localauthapi
+
+import (
+	"fmt"
+	"libhoop/log"
```
sandromello: use the log from `common` package

gateway/api/middleware.go
```diff
@@ -51,70 +96,188 @@ func (a *Api) Authenticate(c *gin.Context) {
...
+			UserSubject: "API_KEY",
+			UserName:    "API_KEY",
+			UserEmail:   "API_KEY",
+			UserStatus:  "active",
+			UserGroups:  []string{"admin"},
```
sandromello: Use `types.GroupAdmin` instead

gateway/transport/interceptors/auth/auth.go
```diff
@@ -125,33 +131,109 @@ func (i *interceptor) StreamServerInterceptor(srv any, ss grpc.ServerStream, inf
...
+			ctx := &pguserauth.Context{
+				OrgID:       orgID,
+				OrgName:     org.Name,
+				OrgLicense:  org.License,
+				UserUUID:    "API_KEY",
+				UserSubject: "API_KEY",
+				UserName:    "API_KEY",
+				UserEmail:   "API_KEY",
+				UserStatus:  "active",
+				UserGroups:  []string{"admin"},
```
sandromello: Use `types.GroupAdmin` instead

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+// Otherwise, a new organization is created for the user.
+func manageOrgCreation(user pgrest.User) (string, error) {
+	var tenancy string
+	if pgusers.IsOrgMultiTenant() {
+		tenancy = "multi-tenant"
+	} else {
+		tenancy = "single-tenant"
+	}
+	switch tenancy {
+	case "multi-tenant":
```
sandromello: What's the motivation to add multi tenant to local auth? Our current multi tenant cloud setup won't be using this.

I'm not against having this, but there's too many things to test it in a single PR.
Did you test this setup running the gateway in multi tenant mode?

gateway/api/localauth/register.go
```diff
@@ -0,0 +1,142 @@
...
+		log.Debug("Creating new organization")
+		newOrgID, err := pgorgs.New().CreateOrGetOrg(fmt.Sprintf("%q Orgnization", user.Email), nil)
```
sandromello: Fix this to:
```go
fmt.Sprintf("%v Organization")
```

sandromello (changes requested):
Please, remove all `fmt.Print...` statements and add make sure to add a log.Infof|Warnf|Error to http errors 4xx and 5xx.

gateway/api/server.go
mtmr0x: good catch

gateway/api/middleware.go
mtmr0x: I'm aware, and an improvement will come soon. For now, this will do for what we want to do

gateway/pgrest/userauth/userauth.go
mtmr0x: makes sense

gateway/api/middleware.go
mtmr0x: good catch. Nevertheless, even though this data is the same loaded in API_URL, I disagree it would come from `IDProvider`. Maybe it should come directly from the config with the configured API_URL.

gateway/api/middleware.go
mtmr0x: nice one

gateway/api/localauth/register.go
mtmr0x: that was added at another moment, on another comprehension of the feature. I just removed it

gateway/pgrest/orgs/helpers.go
mtmr0x: you are right. This was added in the beginning of the code due to a limitation, but then I found out how to go around that limitation. I will revert this file and keep it as is

gateway/pgrest/orgs/orgs.go
mtmr0x: this method was made necessary because of the API key. At the interceptor layer, there's no `orgCtx` and to parse the API token it was necessary

gateway/transport/interceptors/auth/auth.go
mtmr0x: I've removed the else statement, but this is not necessarily a DRY problem. If I move this code to the end of the file, I'd have to move a lot of variables to the top of the block and cause collateral effects on them procedurally; it not only makes the code harder to read but also saves a couple of lines in favor of adding more on the top of the block. Also, the contexts and some other values are slightly different, and making the names different in favor of readability at the cost of some repetition is fairly okay in this case.

gateway/api/localauth/register.go
mtmr0x: Agreed! I added it to the app configuration. When AUTH_METHOD is set to local, we should reinforce for a JWT secret.

gateway/api/server.go
mtmr0x: as we plan to expand local auth, I've created a middleware called `LocalAuthOnly` to manage this

gateway/pgrest/orgs/orgs.go
sandromello: To create an organization context use this method, the orgID is a parameter to the create the context.
```go
orgID := "<my-org-uuid>"
orgCtx := pgrest.NewOrgContext(orgID)
```

gateway/api/localauth/login.go
```diff
@@ -0,0 +1,58 @@
...
+		UserID:      dbUser.ID,
+		UserEmail:   dbUser.Email,
+		UserSubject: dbUser.Subject,
+		RegisteredClaims: jwt.RegisteredClaims{
+			ExpiresAt: jwt.NewNumericDate(expirationTime),
+		},
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	tokenString, err := token.SignedString(appconfig.Get().JWTSecretKey())
```
sandromello, line 48: I was meaning that we need to strict validate if the secret key is empty, otherwise we could emit insecure tokens.

mtmr0x: it is being done at app configuration level. This is never empty at this point

sandromello (approved):

PR #467, opened 2024-09-09
Bundle of fixes and small implementations

rogefm:
This PR contains:
- Change the `Run` button and the `Go to Configurations` button of runbooks;
- Fix the font of modals;
- Change the default state of the sidebar to `close`
- Inside the session list we are truncating the big names or emails;
- Change icons in runbooks;
- Refactored the spacing of the side menu;
- Only show runbooks allowed in connections;


sandromello (commented):
Hey @rogefm , could you describe in more details what this PR does? It's important to have the list of changes, it could help us to diagnose issues in releases based on the outlined changes.

Could you also change the title as well? What's "Big Catado"?

PR #470, opened 2024-09-11
Add agent integration with HashiCorp Vault

sandromello:
Add integration with Vault KV Store (v1, v2).
- Allow fetching from distinct key value store in the same connection
- Allow specifying the path of a secret to be able to use custom mount paths or versioned secrets

## Example

The example below shows how to obtain secret from version 1 and 2

```sh
# versioned secrets (v2)
vault kv put -mount=secret dbsecret HOST=127.0.0.1 PASS=mydbsecret
vault kv put -mount=secret dbsecret HOST=127.0.0.1 PASSWD=mydbsecret

# non versioned (v1)
vault kv put kv/dbsecret2 USER=mydbuser

hoop admin create conn bash -a default \
  -e HOST=_vaultkv2:dbsecret:HOST \
  -e PASS=_vaultkv2:/secret/data/dbsecret?version=2:PASSWD \
  -e USER=_vaultkv1:/kv/dbsecret2:USER \
  -- bash
```


client/config/config.go
```diff
@@ -124,7 +124,7 @@ func (c *Config) GrpcClientConfig() (grpc.ClientConfig, error) {
 	return grpc.ClientConfig{
 		ServerAddress: srvAddr,
 		Token:         c.Token,
-		Insecure:      hasInsecureScheme(c.GrpcURL),
+		Insecure:      c.InsecureGRPC,
```
sandromello, line 127: It's a missing change from last improvements on docker-compose to allow client to connect insecurely.

deploy/docker-compose/rootfs/etc/nginx/nginx-default.conf
```diff
@@ -18,7 +18,6 @@ http {
     }
 
     server {
-        server_name 192.168.15.48;
```
sandromello, line 21: it's safe to remove: there's no need to use server_name for non tls setup,

mtmr0x (approved):

PR #482, opened 2024-09-24
enhancements(tracking) track license type and auth method

mtmr0x:


gateway/api/middleware.go
```diff
@@ -315,10 +316,26 @@ func (a *Api) TrackRequest(eventName string) func(c *gin.Context) {
 			c.Next()
 			return
 		}
+		org, err := pgorgs.New().FetchOrgByContext(ctx)
```
sandromello: I recommend moving this to the middleware layer because we're already obtaining this information from the user (license data, org, etc). This part should only use this information contained in the `ctx` variable.

gateway/api/middleware.go
```diff
@@ -315,10 +316,26 @@ func (a *Api) TrackRequest(eventName string) func(c *gin.Context) {
...
+		if err != nil {
+			log.Errorf("failed fetching org, reason=%v", err)
+		}
```
sandromello: Make sure to return after an error, otherwise it will blow up with nil pointer exception

mtmr0x: good point. I will adjust it, but the thing is, it should not fail if we fail to parse the license. Sounds like a problem we want to avoid. The ideal scenario here is that the license arrives at the tracker, but if it doesn't, the request should not fail, it should just send an empty field for the license maybe

sandromello (approved):

PR #486, opened 2024-09-26
[FIX]: Copy in the tcp and ssh create connections

rogefm:
- Copy in the TCP and SSH creating connections
- Fix the Dashboard available in the side menu collapsed


mtmr0x (approved):
dale

PR #487, opened 2024-09-26
[GEN-758][AUTH] Remove authentication locks for multitenant and implement invited user flow under the system control

mtmr0x:
This PR adjusts behavior, so single-tenant installations keep working as is but simplifies user control for multi-tenant usage, like:

 - adds enum type "invite" to user status so when a user is invited, we apply this status;
 - remove logic that would complicate behavior for user sign up, both at sign up (this just keeps working as is; no change was applied directly to this) and signing, allowing users into their organization who invited them or creating a new one for fresh new users;
 - removes status active logic from the model layer and places it at controllers so we can rule best what to do in each case (part of complicated behavior removal)
 - removes "verified" status locks for letting people create their accounts;
 - creates a FetchInvitedUser method at model layer to apply signup rules for invited users


gateway/api/user/user.go
```diff
@@ -88,8 +88,8 @@ func Create(c *gin.Context) {
 		HashedPassword: newUser.HashedPassword,
 		Picture:        newUser.Picture,
 		Email:          newUser.Email,
-		Verified:       newUser.Verified, // DEPRECATED in flavor of role
-		Status:         string(openapi.StatusActive),
+		Verified:       newUser.Verified,
```
rogefm, line 91: Is it no longer "Deprecated in favor of role"?

gateway/api/openapi/types.go
```diff
@@ -30,8 +30,10 @@ type (
 )
 
 const (
-	StatusActive   StatusType = "active"
-	StatusInactive StatusType = "inactive"
+	StatusActive    StatusType = "active"
+	StatusInactive  StatusType = "inactive"
+	StatusReviewing StatusType = "reviewing"
```
rogefm, line 35: Do we have these values duplicated?
I will comment on the duplicates.

gateway/storagev2/types/const.go
```diff
@@ -29,6 +29,7 @@ const (
 	UserStatusActive    UserStatusType = "active"
 	UserStatusReviewing UserStatusType = "reviewing"
 	UserStatusInactive  UserStatusType = "inactive"
+	UserStatusInvited   UserStatusType = "invited"
```
rogefm: Duplicate value.

gateway/api/login/login.go
```diff
@@ -143,39 +144,44 @@ func (h *handler) LoginCallback(c *gin.Context) {
...
 	// multi tenant won't sync users
 	if pgusers.IsOrgMultiTenant() {
-		isNewUser, err := registerMultiTenantUser(uinfo, login.SlackID)
-		if err != nil {
-			login.Outcome = fmt.Sprintf("failed registering multi tenant user subject=%s, email=%s, reason=%v",
-				uinfo.Subject, uinfo.Email, err)
-			log.With("multitenant", true).Error(login.Outcome)
-			sentry.CaptureException(err)
+		isNewUser := false
+		if ctx.UserStatus == string(types.UserStatusInactive) {
```
sandromello, line 155: Actually, the method `FetchUserContext` only return users that are under the `active` state. So this condition isn't required anymore

gateway/api/login/login.go
```diff
@@ -143,39 +144,44 @@ func (h *handler) LoginCallback(c *gin.Context) {
...
+			log.With("multitenant", true).Warnf("user %s is inactive. They need to be edited to active before trying to signin", uinfo.Email)
 			c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
 			return
 		}
-		// the signup process is performed by /api/signup when the gateway is running multi tenant mode
-		// track as a login event
-		if isNewUser {
-			h.analyticsTrack(false, userAgent, ctx)
+		if ctx.UserStatus != string(types.UserStatusActive) {
```
sandromello, line 160: I recommend using `ctx.IsEmpty()` method that makes this condition more clear to read. The user will always be in `active` state if he exists.

gateway/api/login/login.go
```diff
@@ -198,29 +204,56 @@ func (h *handler) LoginCallback(c *gin.Context) {
...
+		if err := pgorgs.New().CreateOrg(newOrgId, newOrgName, nil); err != nil {
+			return false, fmt.Errorf("failed setting new org, reason=%v", err)
+		}
+		newUser := pgrest.User{
+			ID:       uuid.NewString(),
+			OrgID:    newOrgId,
+			Subject:  uinfo.Subject,
+			Name:     uinfo.Profile,
+			Email:    uinfo.Email,
+			Verified: *uinfo.EmailVerified,
```
sandromello: You must validate if this attribute is null before trying to access it directly, otherwise it could blow with nil pointer exception:

```
emailVerified := false
if uinfo.EmailVerified != nil {
  emailVerified = uinfo.EmailVerified
}
```

gateway/api/middleware.go
```diff
@@ -67,6 +67,10 @@ func (a *Api) localAuthMiddleware(c *gin.Context) {
 	}
 
 	ctx, err := pguserauth.New().FetchUserContext(claims.UserSubject)
+	if ctx.UserStatus != string(types.UserStatusActive) {
```
sandromello: This method already validates that. If you want to validate if the user doesn't exists, i recommend using `ctx.IsEmpty()`.

gateway/api/middleware.go
```diff
@@ -184,6 +188,10 @@ func (a *Api) Authenticate(c *gin.Context) {
 		}
 
 		ctx, err := pguserauth.New().FetchUserContext(subject)
+		if ctx.UserStatus != string(types.UserStatusActive) {
```
sandromello: If the user is not active it will return in the `err` variable. There's not need for this condition

gateway/api/openapi/types.go
mtmr0x: I don't fully support having them duplicated, but one is for OpenAPI documentation spec, and the other is internally used. We could use one, but we are using both. We can change that on another refactor, so we don't go changing all around, as it's not the focus of this PR

gateway/api/user/user.go
mtmr0x: no, we will keep "verified" because it's an email-verified feature from IDPs. Roles play another function on all of this

gateway/api/openapi/types.go
```diff
@@ -30,8 +30,10 @@ type (
...
+	StatusInvited   StatusType = "invited"
```
sandromello, line 36: This type doesn't exists in the database as an enum type, it requires a migration otherwise it will return an error when creating a user with this status.

gateway/storagev2/types/const.go
sandromello: It's better to use the openapi types. Eventually these types are going to be deprecated

gateway/transport/interceptors/auth/auth.go
```diff
@@ -120,6 +120,9 @@ func (i *interceptor) StreamServerInterceptor(srv any, ss grpc.ServerStream, inf
 			return status.Errorf(codes.Unauthenticated, "invalid authentication")
 		}
 		userCtx, err := pguserauth.New().FetchUserContext(sub)
+		if userCtx.UserStatus != string(types.UserStatusActive) {
+			return status.Errorf(codes.Unauthenticated, "user is not active")
+		}
```
sandromello, line 129: Remove this, it's already being validated. When the context is not active or not found.

gateway/transport/interceptors/auth/auth.go
```diff
@@ -210,6 +213,9 @@ func (i *interceptor) StreamServerInterceptor(srv any, ss grpc.ServerStream, inf
...
```
sandromello, line 220: Remove this, it's already being validated. When the context is not active or not found.

gateway/api/user/user.go
```diff
@@ -88,8 +88,8 @@ func Create(c *gin.Context) {
...
+		Status:         string(openapi.StatusInvited),
```
sandromello, line 92: This will change the behavior of single tenant installations.

sandromello (changes requested):

gateway/api/openapi/types.go
mtmr0x: @sandromello instructed me on removing it, so it's done. Thanks for bringing it up

sandromello (approved):

PR #490, opened 2024-09-30
[webapp] update webapp packages to fix vulnerabilities

sandromello:
Changes provided by `npm audit fix` command

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -5986,21 +5986,21 @@
       "dev": true
     },
     "node_modules/body-parser": {
-      "version": "1.20.1",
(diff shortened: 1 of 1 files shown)


mtmr0x:
Hey @sandromello, did you have the chance to try running dev and release modes for this? These updates usually do no harm, but double-checking is good to make sure we don't have to manually update the dependency in the `package.json` file due to some incompatibility. It doesn't require much testing; if it builds, it should be enough, but if you try to access it and everything seems normal, even better.

sandromello:
Hey @mtmr0x , yes I did. I actually tested as well the release build. These changes was generated by the command `npm audit fix` command.

mtmr0x (approved):

PR #500, opened 2024-10-04
[CHORE] refactor models layer: user and user groups

mtmr0x:


gateway/api/localauth/login.go
```diff
@@ -25,17 +25,12 @@ func Login(c *gin.Context) {
 		return
 	}
 
-	dbUser, err := pgusers.GetOneByEmail(user.Email)
+	dbUser, err := models.GetUserByEmail(user.Email)
```
sandromello, line 28: You need to check if dbuser is nil

gateway/models/user_groups.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func InsertUserGroups(userGroups []UserGroup) error {
+	log.Debugf("inserting user groups")
+	if len(userGroups) == 0 {
+		return nil
+	}
+	if err := DB.Create(&userGroups).Error; err != nil {
+		return err
+	}
```
sandromello: Just return the error directly:
```go
return DB.Create(...)
```

gateway/models/user_groups.go
```diff
@@ -0,0 +1,57 @@
...
+
+	return nil
+}
+
+func DeleteUserGroupsByUserID(userID string) error {
+	log.Debugf("deleting user groups for user=%s", userID)
+	if err := DB.Where("user_id = ?", userID).Delete(&UserGroup{}).Error; err != nil {
+		log.Errorf("failed to delete user groups, reason=%v", err)
```
sandromello: Don't log it here, propagate the error and log it in the caller

gateway/api/localauth/register.go
```diff
@@ -81,6 +79,19 @@ func Register(c *gin.Context) {
...
 
+	adminUserGroup := models.UserGroup{
+		OrgID:  org.ID,
+		UserID: userID,
+		Name:   adminGroupName,
+	}
+
+	err = models.InsertUserGroups([]models.UserGroup{adminUserGroup})
+	if err != nil {
+		log.Debugf("failed creating user group, err=%v", err)
```
sandromello: It's an internal error, use log.Errorf

gateway/api/login/login.go
```diff
@@ -204,7 +204,7 @@ func (h *handler) LoginCallback(c *gin.Context) {
 }
 
 func registerMultiTenantUser(uinfo idp.ProviderUserInfo, slackID string) (isNewUser bool, err error) {
-	iuser, err := pgusers.New().FetchInvitedUser(&pguserauth.Context{}, uinfo.Email)
+	iuser, err := models.GetUserByEmail(uinfo.Email)
```
sandromello: You're changing the logic, the previous method fetch an invited user, this one fetches a valid user.

gateway/api/login/login.go
```diff
@@ -307,16 +336,28 @@ func syncSingleTenantUser(ctx *pguserauth.Context, uinfo idp.ProviderUserInfo) (
...
+					OrgID:  org.ID,
+					UserID: iuser.ID,
+					Name:   ctx.UserGroups[i],
+				})
+			}
+			// update the user groups if it fails, we don't return an error
+			// because it's only a sync and the database won't allow to save
+			// a duplicated user group due to the unique constraint for
+			// user_id and name in user_groups table
+			_ = models.InsertUserGroups(invitedUserGroups)
```
sandromello: You must differentiate between errors. This method could return an internal error which is different from your goal. The request should fail under these conditions.

gateway/api/login/login.go
```diff
@@ -280,10 +290,28 @@ func syncSingleTenantUser(ctx *pguserauth.Context, uinfo idp.ProviderUserInfo) (
...
+				OrgID:  ctx.OrgID,
+				UserID: ctx.UserUUID,
+				Name:   userGroups[i],
+			})
+		}
+		// update the user groups if it fails, we don't return an error
+		// because it's only a sync and the database won't allow to save
+		// a duplicated user group due to the unique constraint for
+		// user_id and name in user_groups table
+		_ = models.InsertUserGroups(newUserGroups)
```
sandromello: One of the goal of this function is to sync the groups when there's is a change, it's important the following statements are true when this function is executed:

- It's able to replace the old groups from a user with the new ones?
- It's able to remove all groups if the request contains empty groups?
- It's able to add new groups if doesn't contain any record?

Does this function guarantees all these requirements?

gateway/api/user/user.go
```diff
@@ -50,14 +52,14 @@ func Create(c *gin.Context) {
 		return
 	}
 
-	existingUser, err := pgusers.New().FetchOneByEmail(ctx, newUser.Email)
-	if err != nil {
+	existingUser, err := models.GetUserByEmailAndOrg(newUser.Email, ctx.OrgID)
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
```
sandromello: You're already checking if the record exists in the underline function

gateway/api/user/user.go
```diff
@@ -50,14 +52,14 @@ func Create(c *gin.Context) {
...
 		log.Errorf("failed fetching existing invited user, err=%v", err)
 		sentry.CaptureException(err)
 		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed fetching existing invited user"})
 		return
 	}
-	if existingUser != nil {
+	if existingUser.Email == newUser.Email {
```
sandromello: Revert the previous logic here, see my comment above.

gateway/api/user/user.go
```diff
@@ -141,24 +158,40 @@ func Update(c *gin.Context) {
 	ctx := storagev2.ParseContext(c)
 	userID := c.Param("id")
 
-	existingUser, err := pgusers.New().FetchOneBySubject(ctx, userID)
+	existingUser, err := models.GetUserBySubjectAndOrg(userID, ctx.OrgID)
+	if errors.Is(err, gorm.ErrRecordNotFound) {
```
sandromello: You're already validating this condition inside the db function

gateway/api/user/user.go
```diff
@@ -141,24 +158,40 @@ func Update(c *gin.Context) {
...
+		c.JSON(http.StatusNotFound, gin.H{"message": "not found"})
+		return
+	}
 	if err != nil {
 		log.Errorf("failed getting user %s, err=%v", userID, err)
 		sentry.CaptureException(err)
 		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed getting user"})
 		return
 	}
-	if existingUser == nil {
```
sandromello, line 151: Revert this condition

gateway/api/user/user.go
```diff
@@ -141,24 +158,40 @@ func Update(c *gin.Context) {
...
-	}
+
 	var req openapi.User
 	if err := c.ShouldBindJSON(&req); err != nil {
 		c.JSON(http.StatusBadRequest, gin.H{"message": err.Error()})
 		return
 	}
+
+	userGroups, err := models.GetUserGroupsByUserID(existingUser.ID)
+	fmt.Printf("userGroups: %+v\n", userGroups)
```
sandromello: remove this print statement

gateway/api/user/user.go
```diff
@@ -172,22 +205,47 @@ func Update(c *gin.Context) {
...
+				Name:   req.Groups[i],
+			})
+		}
+		if err := models.InsertUserGroups(newUserGroups); err != nil {
+			log.Errorf("failed persisting user groups, err=%v", err)
+			sentry.CaptureException(err)
+			c.JSON(http.StatusInternalServerError, gin.H{"message": "failed persisting user groups"})
+			return
+		}
+	}
```
sandromello: This is the problem with not using transactions / full queries for cases likes this:
This method is now non idempotent and could leave the user in a inconsistent state.

We are trading "simplicity" over inconsistency.

gateway/api/user/user.go
```diff
@@ -218,29 +276,40 @@ func Update(c *gin.Context) {
...
 		sentry.CaptureException(err)
 		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed listing users"})
 		return
 	}
-	userList := []openapi.User{}
-	for _, u := range users {
-		userList = append(userList,
+
+	usersList := []openapi.User{}
+	for i, u := range users {
```
sandromello, line 284: There's no need to use this variable, just hide it with `_` and use `u.ID`

gateway/api/user/user.go
```diff
@@ -218,29 +276,40 @@ func Update(c *gin.Context) {
...
+		userGroups, err := models.GetUserGroupsByUserID(users[i].ID)
```
sandromello: This is very bad, because you're executing an statement for each user record.
You could improve this by:
- [good] Performing one query to select all groups from an organization and them build the list payload
- [better] Perform a single query that return the groups in a single object

gateway/api/user/user.go
```diff
@@ -256,22 +325,18 @@ func List(c *gin.Context) {
...
 	subject := c.Param("id")
-	user, err := pgusers.New().FetchOneBySubject(ctx, subject)
+	user, err := models.GetUserBySubjectAndOrg(subject, ctx.OrgID)
 	if err != nil {
 		log.Errorf("failed getting user %s, err=%v", subject, err)
 		sentry.CaptureException(err)
 		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed getting user"})
 		return
 	}
-	if user == nil {
```
sandromello, line 266: Revert this condition

gateway/api/user/user.go
```diff
@@ -293,34 +358,44 @@ func Delete(c *gin.Context) {
...
-		user, err = pgusers.New().FetchOneBySubject(ctx, emailOrID)
+		user, err = models.GetUserBySubjectAndOrg(emailOrID, ctx.OrgID)
 	}
 	if err != nil {
 		log.Errorf("failed getting user %s, err=%v", emailOrID, err)
 		sentry.CaptureException(err)
 		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed getting user"})
 		return
 	}
-	if user == nil {
```
sandromello, line 309: Revert this condition

gateway/api/user/user.go
```diff
@@ -293,34 +358,44 @@ func Delete(c *gin.Context) {
...
 		Status:   openapi.StatusType(user.Status),
 		Verified: user.Verified, // DEPRECATED in flavor of role
-		Role:     toRole(*user),
 		SlackID:  user.SlackID,
 		Picture:  user.Picture,
-		Groups:   user.Groups,
-	})
+	}
+	userResponse.Groups = []string{}
+	for ug := range userGroups {
```
sandromello: You could just get the object by adding one more variable:

```go
for _, ug := range userGroups {
 	userResponse.Groups = append(userResponse.Groups, ug.Name)
}
```

gateway/api/user/user.go
```diff
@@ -412,9 +487,10 @@ func GetUserInfo(c *gin.Context) {
 //	@Router			/users/self/slack [patch]
 func PatchSlackID(c *gin.Context) {
 	ctx := storagev2.ParseContext(c)
-	u, err := pgusers.New().FetchOneBySubject(ctx, ctx.UserID)
-	if err != nil || u == nil {
```
sandromello, line 416: Revert this condition, your method now returns nil for not found

gateway/transport/plugins/slack/events.go
```diff
@@ -26,8 +25,12 @@ func (p *slackPlugin) processEventResponse(ev *event) {
 		ev.msg.ID, ev.msg.Status)
 
 	// validate if the slack user is able to review it
-	orgCtx := pgrest.NewOrgContext(ev.orgID)
-	slackApprover, err := pgusers.New().FetchOneBySlackID(orgCtx, ev.msg.SlackID)
+	slackApprover, err := models.GetUserByOrgIDAndSlackID(ev.orgID, ev.msg.SlackID)
+	slackApproverGroups, err := models.GetUserGroupsByUserID(slackApprover.ID)
```
sandromello, line 35: You must check the error first

gateway/models/users.go
```diff
@@ -0,0 +1,127 @@
...
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	return user, nil
+}
+
+func GetUserByEmailAndOrg(email, orgID string) (*User, error) {
```
sandromello, line 73: Let's try to keep a pattern where the orgID comes first since most of the resources derive from an organization. Therefore, there's no need to call the function of `...EmailAndOrg`, it's implicit if you're passing this information:

```go
func GetUserByEmail(orgID, email string) (*User, error) {
  (...)
}
```

gateway/models/database.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/hoophq/hoop/common/log"
+	"github.com/hoophq/hoop/gateway/appconfig"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+)
+
+// This makes the DB generally available to the application
+var DB *gorm.DB
+
+func InitDatabase() {
```
sandromello, line 17: We need to make sure this function tries to reconnect in case of a failure,  otherwise the whole application will be unavailable.

sandromello (changes requested):
Before proceeding, make sure to **comment** all unused functions at `./gateway/pgrest/users`, it will enforce the deprecation of these functions.

It's important to keep it commented for us to compare in case of any inconsistent during this process of migration.

gateway/models/database.go
mtmr0x: we don't, if it fails, it fails. This function expects a valid database connection, no magic, just explicitly.

gateway/api/user/user.go
mtmr0x: indeed it is, but how we designed our user groups right now is far from ideal, so the controller layer will have to pay this price during this refactor until we match the simplicity we want

gateway/api/user/user.go
mtmr0x: potential inconsistency, actually. And it's fine, like I've said in the other comment, the controller will pay the price. The one thing tho we might do to avoid this is to create a specific function at the model layer to handle a db transaction with all these operations. Still considering the overall availability of the DB using GORM I see this as a low-risk action to take. Whether doing it or not is a matter of detail

gateway/api/user/user.go
mtmr0x: that's true, nice catch. I've moved some logic to the inside and forgot to remove it here

gateway/api/login/login.go
mtmr0x: you're right, I will force sync these groups

gateway/models/users.go
```diff
@@ -0,0 +1,156 @@
...
+	log.Debugf("deleting user=%s for org=%s", subject, orgID)
+	if err := DB.Where("org_id = ? AND subject = ?", orgID, subject).Delete(&User{}).Error; err != nil {
+		log.Errorf("failed to delete user, reason=%v", err)
+		return err
+	}
+
+	return nil
+}
+
+func UpdateUserAndUserGroups(user *User, userGroups []UserGroup) error {
```
sandromello, line 129: Don't log error in this method, just append it to the error and log in the caller
```go
return fmt.Errorf("failed doing whatever, reason=%v", err)
```

gateway/models/users.go
```diff
@@ -0,0 +1,156 @@
...
+	// create new user groups
+	if len(userGroups) > 0 {
+		if err := tx.Create(&userGroups).Error; err != nil {
+			tx.Rollback()
+			log.Errorf("failed to create user groups, reason=%v", err)
+			return err
+		}
+	}
+
+	tx.Commit()
```
sandromello: This will return an error, make sure to return it

gateway/models/users.go
mtmr0x: my bad

sandromello (approved):

PR #502, opened 2024-10-08
[FEAT]: New create connection

rogefm:
Figma -> https://www.figma.com/design/GhiQ0r5qL37UV9WTS8LBsj/Main?node-id=2631-17772&node-type=instance&m=dev

Checklist:

- [x] New create connection UI
- [x] New update connection UI
- [x] Rules for creating a new connection
- [x] Rules for updating a connection
- [x] New UI components
  - [x] Dialog
  - [x] Modal
  - [x] Accordion
  - [x] Input
  - [x] Select
  - [x] Text area
- [x] Delete old files
- [x] Colors in Tailwind/Radix

We removed icon_name from the create and update connection source code in this new implementation.


webapp/src/webapp/components/accordion.cljs
```diff
@@ -0,0 +1,109 @@
...
+      (when show-configure? [configure-button])
+      (when show-icon? [status-icon status])
+
+      [:> ChevronRight {:size 16 :className "text-[--gray-12] transition-transform duration-300 group-data-[state=open]:rotate-90"}]]]]
+
+   [:> (.-Content Accordion)
+    [:> Box {:px "5" :py "7" :className "bg-white border-t border-[--gray-a6] rounded-b-6"}
+     content]]])
+
+(defn main
```
mtmr0x: I don't think that this component should have to worry about a list. It's better to the component in the upper scope to call the accordion component as many times as it needs, instead giving this responsibility here.

webapp/src/webapp/components/accordion.cljs
```diff
@@ -0,0 +1,109 @@
...
+   [reagent.core :as r]))
+
+(defn badge [text]
+  [:> Badge {:color "red" :variant "soft" :size "2"} text])
+
+(defn configure-button []
+  [:> Button {:variant "soft" :size "3"} "Configure"])
+
+(defn checkbox []
+  [:> Checkbox])
```
mtmr0x: shouldn't these be separated from this file? Checkbox and buttons are important first class components that should be organized in its own component

webapp/src/webapp/components/accordion.cljs
```diff
@@ -0,0 +1,109 @@
...
+                  :fallback (r/as-element (if avatar-icon
+                                            avatar-icon
+                                            [:> User {:size 16}]))}]
+
+      [:div {:className "flex flex-col items-start"}
+       [:> Text {:size "5" :weight "bold" :className "text-[--gray-12]"} title]
+       [:> Text {:size "3" :className "text-[--gray-11]"} subtitle]]]
+
+     [:div {:className "flex space-x-3 items-center"}
+      (when show-badge? [badge "Badge"])
```
mtmr0x: is this text "Badge" hard coded here?

webapp/src/webapp/components/dialog.cljs
```diff
@@ -48,52 +49,75 @@
...
+;;                              ((:on-success @dialog-state))
+;;                              (rf/dispatch [:dialog->close]))}
+;;         (if (:text-action-button @dialog-state)
+;;           (:text-action-button @dialog-state)
+;;           "Delete")]
+;;        [:button {:type "button"
+;;                  :class "mt-3 inline-flex w-full justify-center rounded-md bg-white px-3 py-2 text-sm font-semibold text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 hover:bg-gray-50 sm:mt-0 sm:w-auto"
+;;                  :on-click #(rf/dispatch [:dialog->close])
+;;                  :data-autofocus true}
+;;         "Cancel"]]]]]]]]
```
mtmr0x: do we need this comment for something or do we delete it?

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
```diff
@@ -0,0 +1,193 @@
...
+
+         (and (= @connection-type "custom")
+              (not= @connection-subtype "ssh"))
+         [:<>
+          [:> Grid {:columns "5" :gap "7"}
+           [:> Flex {:direction "column" :grid-column "span 2 / span 2"}
+            [:> Text {:size "4" :weight "bold"} "Load environment variables"]
+            [:> Text {:size "3"} "Add variables values to use in your connection."]]
+           [:> Grid {:columns "2" :gap "2" :grid-column "span 3 / span 3"}
+            (println @configs)
```
mtmr0x: println

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
```diff
@@ -0,0 +1,193 @@
...
+                           :classes "whitespace-pre overflow-x"
+                           :placeholder "* * * *"
+                           :type "password"
+                           :value @config-value}]]
+            [:> Box {:grid-column "span 2 / span 2" :class "justify-self-center"}
+             [:> Button {:size "2"
+                         :variant "ghost"
+                         :type "button"
+                         :color "gray"
+                         :on-click #(on-click->add-more-key-value)}
```
mtmr0x, line 117: Buttons should be at component layer in favor of keeping consistency

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
```diff
@@ -0,0 +1,193 @@
...
+            [forms/textarea {:label "Content"
+                             :placeholder "Paste your file content here"
+                             :on-change #(reset! config-file-value (-> % .-target .-value))
+                             :value @config-file-value}]
+            [:> Box {:class "justify-self-center"}
+             [:> Button {:size "2"
+                         :variant "ghost"
+                         :type "button"
+                         :color "gray"
+                         :on-click #(on-click->add-more-file-content)}
```
mtmr0x, line 142: same comment on button

webapp/src/webapp/connections/views/create_update_connection/connection_type_form.cljs
```diff
@@ -0,0 +1,120 @@
...
+
+    (= value "ssh") (and (= value connection-subtype)
+                         (= connection-type "custom"))
+
+    (= value "tcp") (and (= value connection-subtype)
+                         (= connection-type "application"))
+
+    (= value "application") (and (= value connection-type)
+                                 (not= connection-subtype "tcp"))))
+
```
mtmr0x, line 82: I understand the need of this in here, but it makes very complicated to understand. Is it because you have to manage some information for the form submit? If so, why not make it simpler here using only the value to set the styles and let the submit worry about these details?

webapp/src/webapp/connections/views/create_update_connection/main.cljs
```diff
@@ -0,0 +1,298 @@
...
+
+(defn- convertStatusToBool [status]
+  (if (= status "enabled")
+    true
+    false))
+
+(defn add-new-configs
+  [config-map config-key config-value]
+
+  (println "config-map" config-map)
```
mtmr0x: print

webapp/src/webapp/connections/views/create_update_connection/connection_type_form.cljs
rogefm: It's not. It isn't very easy because, SSH and TCP are not a type, but we are using them as a type in the UI and UX.
In this case, I'm using it to tell me which card of connection type is selected.
So I'm comparing the value selected with the atoms. I would like to compare only `@connection-type (TCP) = value (TCP)` but it's not true

webapp/src/webapp/components/accordion.cljs
rogefm: We don't need a checkbox in this implementation, it was only to follow the Figma, I will remove it
The Button I didn't make yet as I said in the Description of the PR.

webapp/src/webapp/components/accordion.cljs
rogefm: Same here, it's not necessary for this implementation, I'm only following the Figma, and I will remove it to simplify.

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
rogefm: The Button I didn't make yet as I said in the Description of the PR.

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
rogefm: The Button I didn't make yet as I said in the Description of the PR.

sandromello (approved):

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
rogefm: I will create a PR for this implementation.

webapp/src/webapp/connections/views/create_update_connection/connection_environment_form.cljs
rogefm: I will create a PR for this implementation.

webapp/src/webapp/connections/views/create_update_connection/connection_type_form.cljs
rogefm: We will change how the form works today. We will send the management state to reframe. 
But it's a big change in the logic of this feature so we will do it in the new PR.

mtmr0x (approved):

PR #524, opened 2024-10-28
Add guard rail rules feature that blocks executions based on rules

sandromello:
- [logging] Change to not output stack trace when logging error
- [gateway] Add CRUD routes to create guard rail rules resources (GET, POST, PUT, DELETE)
- [gateway] Add connection attribute to create rule associations with connections
- [gateway] Refactor connections to use gorm over postgrest


gateway/models/connections.go
```diff
@@ -0,0 +1,272 @@
...
+			GROUP BY rc.connection_id
+		), ARRAY[]::TEXT[]) AS guardrail_rules,
+		(
+			SELECT json_agg(r.input) FROM private.guardrail_rules r
+			INNER JOIN private.guardrail_rules_connections rc ON rc.connection_id = c.id AND rc.rule_id = r.id
+		) AS guardrail_input_rules,
+		(
+			SELECT json_agg(r.output) FROM private.guardrail_rules r
+			INNER JOIN private.guardrail_rules_connections rc ON rc.connection_id = c.id AND rc.rule_id = r.id
+		) AS guardrail_output_rules
```
mtmr0x: we don't need to make this complicated join. This complexity is easily replaceable by responding the IDs of the guardrails rules, and the clients request them at their endpoint.

sandromello: - The `guardrail_input_rules` and `guardrail_output_rules` is used internally and  is not exposed to the api.

We already have an endpoint that exposes the guard rail rules, the `guardrail_rules` join is the association of rules with a connection.
Are you saying that we should return the `id` of the `private.guardrail_rules_connections` table instead of obtaining their names? The client will need to perform an additional request (e.g.: listing guard rail rules) to obtain the name for this case, are you ok with this? If this is the case, IMO we're just shifting complexity to another layer (WebApp / Clients)

mtmr0x: there's no complexity on this behavior, it's standard REST API plus keeping the concepts transparent and explicit. So yes, let the clients do extra requests

gateway/api/guardrails/guardrails.go
```diff
@@ -0,0 +1,204 @@
...
+	req := openapi.GuardRailRuleRequest{}
+	if err := c.ShouldBindJSON(&req); err != nil {
+		log.Errorf("failed parsing request payload, err=%v", err)
+		c.JSON(http.StatusBadRequest, gin.H{"message": err.Error()})
+		return nil
+	}
+	return &req
+}
+
+func toJSON(c *gin.Context, statusCode int, obj any) {
```
mtmr0x: this function would be an unnecessary abstraction, as they are used at two methods (List and Get)

gateway/models/connections.go
```diff
@@ -0,0 +1,266 @@
...
+	AccessModeRunbooks string            `gorm:"column:access_mode_runbooks"`
+	AccessModeExec     string            `gorm:"column:access_mode_exec"`
+	AccessModeConnect  string            `gorm:"column:access_mode_connect"`
+	AccessSchema       string            `gorm:"column:access_schema"`
+	Envs               map[string]string `gorm:"column:envs;serializer:json;->"`
+	GuardRailRules     pq.StringArray    `gorm:"column:guardrail_rules;type:text[];->"`
+
+	// Read Only fields
+	GuardRailInputRules  []byte         `gorm:"column:guardrail_input_rules;->"`
+	GuardRailOutputRules []byte         `gorm:"column:guardrail_output_rules;->"`
```
mtmr0x: can we remove these fields from here, considering we will use the ids only?

gateway/api/server.go
```diff
@@ -466,4 +467,34 @@ func (api *Api) buildRoutes(route *gin.RouterGroup) {
...
+	route.GET("/guardrail-rules/:id",
+		api.AllowApiKey,
+		api.Authenticate,
+		apiguardrails.Get)
+
+	route.DELETE("/guardrail-rules/:id",
+		api.AllowApiKey,
+		api.Authenticate,
+		api.TrackRequest(analytics.EventDeleteGuardRailRules),
+		apiguardrails.Delete)
```
mtmr0x: for consistency, in frontend and design, we defined the names as "guardrails", so to keep that let's make these endpoints be `/guardrails` instead of `/guardrail-rules`

rogefm (approved):

mtmr0x (approved):

PR #569, opened 2024-11-26
[REFACTOR]: new behavior on accordions at creating a connection

rogefm:
- New accordion behaviors
- New helpers file to concentrate reusable functions
- Add auto-fill of connection name based on your type
- Fix warning in the textarea input
- Fix the problem of not sending the value of the command when choosing the SSH type


sandromello (approved):
LGTM

PR #571, opened 2024-11-27
[gateway] Add honeycomb http tracing

sandromello:


common/monitoring/monitoring.go
```diff
@@ -71,13 +72,19 @@ func NewOpenTracing(apiURL string) (ShutdownFn, error) {
 	// Enable multi-span attributes
 	bsp := honeycomb.NewBaggageSpanProcessor()
 
+	hcApiKey := honeycombApiKey
+	if hcApiKey == "" {
+		hcApiKey = os.Getenv("HONEYCOMB_API_KEY")
+	}
```
mtmr0x: I have two questions on this: 
1) I couldn't find `honeycombApiKey` in this file; is it some golang magic or am I missing something?
2) at line 77, shouldn't we get this value from our appconfig module?

mtmr0x (approved):
I've commented two things in it, but not a blocker

common/monitoring/monitoring.go
sandromello: 1. The value of honeycombApiKey is dynamically added in the build time when a new release is generated
2. This value is used only for development, I will add a comment there.

PR #574, opened 2024-11-29
Provide a way to specify GroupsClaim with environment variable

vicmarbev:
As of now, if we are authenticating via an OIDC compliant IDP (I'm using Keycloak) we are not able to set the GroupsClaim, being forced to use claim `"https://app.hoop.dev/groups"` defined here: https://github.com/hoophq/hoop/blob/main/common/proto/const.go#L65.
With this PR you could use an environment variable in case you want to use the standard `groups` claim (or any other you want).

What do you think about this?

### DIFF
--- a/gateway/security/idp/authenticator.go
+++ b/gateway/security/idp/authenticator.go
@@ -159,6 +159,9 @@ func setProviderConfFromEnvs(p *Provider) error {
 	p.Audience = os.Getenv("IDP_AUDIENCE")
 	p.CustomScopes = os.Getenv("IDP_CUSTOM_SCOPES")
 	p.GroupsClaim = proto.CustomClaimGroups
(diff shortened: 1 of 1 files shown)


vicmarbev:
Hi @sandromello , could you provide any feedback on this?

sandromello:
Hey @vicmarbev , thanks for helping to improve Hoop with this pull request. We're currently evaluating how to accept external contributions to this project. There're some legal details that we're still discussing internally.

For that time being, you can configure the groups claim using the environment variable `IDP_URI` with the query string `&groupsclaim=<yourvalue>`. Refer to the Oauth2/OIDC Auth tab in the [environment variable documentation](https://hoop.dev/docs/configure/env-vars) for more details.

vicmarbev:
Ok, I'll go ahead and also close #576 then. Thanks!

PR #578, opened 2024-11-29
[FEAT] Database schema in the Gateway

rogefm:
- Create 2 endpoints to get information from a database connection
  - Endpoint to GET databases;
  - Endpoint to GET schema and indexes;
- Create and improve all queries to get database schema;
- Remove all database schema logic from frontend and start to get the database schema information from Gateway;
- Provide for the connections of the types Mongo and Postgres to interact with all databases available;


gateway/api/connections/helpers.go
```diff
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/gin-gonic/gin"
```
sandromello, line 14: Remove this unused import

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
 	}
 	return conn, nil
 }
+
```
sandromello, line 396: Rewrite every comment to English please.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+			}
+		}
+
+		c.JSON(http.StatusOK, DatabaseListResponse{
+			Databases: databases,
+		})
+	case <-timeoutCtx.Done():
+		client.Close()
+		log.Infof("runexec timeout (50s), it will return async")
+		c.JSON(http.StatusRequestTimeout, gin.H{"message": "Request timed out"})
```
sandromello: The more appropriate status code is bad request for this case, because it failed to return the proper time defined by the server. A 408 status code is usually sent when the connection with the http client is lost which is not the case.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+			var result []map[string]interface{}
+			if err := json.Unmarshal([]byte(outcome.Output), &result); err != nil {
+				c.JSON(http.StatusInternalServerError, gin.H{"message": fmt.Sprintf("failed to parse MongoDB response: %v", err)})
+				return
+			}
+			for _, db := range result {
+				if dbName, ok := db["database_name"].(string); ok {
+					databases = append(databases, dbName)
+				}
+			}
```
sandromello: Since you're using the function `parseCommandOutput`, just move this whole block logic to this function which will be more clean.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+
+	if dbName == "" {
+		connEnvs := conn.Envs
+		switch pb.ToConnectionType(conn.Type, conn.SubType.String) {
+		case pb.ConnectionTypePostgres:
+			dbName = getEnvValue(connEnvs, "envvar:DB")
+		case pb.ConnectionTypeMSSQL:
+			dbName = getEnvValue(connEnvs, "envvar:DB")
+		case pb.ConnectionTypeMySQL:
+			dbName = getEnvValue(connEnvs, "envvar:DB")
```
sandromello: Refactor to use one line instead: 
```go
(...)
case pb.ConnectionTypePostgres, case pb.ConnectionTypeMSSQL, pb.ConnectionTypeMySQL:
  dbName = getEnvValue(connEnvs, "envvar:DB")
```

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+		case pb.ConnectionTypeMongoDB:
+			if connStr := connEnvs["envvar:CONNECTION_STRING"]; connStr != "" {
+				dbName = getMongoDBFromConnectionString(connStr)
+			}
+		case pb.ConnectionTypeOracleDB:
+			dbName = getEnvValue(connEnvs, "envvar:SID")
```
sandromello, line 577: Is this correct (`envvar:SID`) or it's a typo?

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+	}
+
+	if conn.Type != "database" {
+		c.JSON(http.StatusBadRequest, gin.H{"message": "connection is not a database type"})
+		return
+	}
+
+	if dbName == "" {
+		connEnvs := conn.Envs
+		switch pb.ToConnectionType(conn.Type, conn.SubType.String) {
```
sandromello: Add a scoped variable as:
```go
connType := pb.ToConnectionType(conn.Type, conn.SubType.String)
```

And then just re-use it throughout this handler.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+	script := getSchemaQuery(pb.ToConnectionType(conn.Type, conn.SubType.String), dbName)
+	if script == "" {
+		c.JSON(http.StatusBadRequest, gin.H{"message": "unsupported database type"})
+		return
+	}
+
+	userAgent := apiutils.NormalizeUserAgent(c.Request.Header.Values)
+	if userAgent == "webapp.core" {
+		userAgent = "webapp.schema"
+	}
```
sandromello: This is just for analytics or adding logic in the gateway to prevent execution based in the user agent.
The following statement should be enough if there isn't any particular reason to provide a custom user agent.

```go
userAgent := apiutils.NormalizeUserAgent(c.Request.Header.Values)
```

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+	case outcome := <-respCh:
+		if outcome.ExitCode != 0 {
+			c.JSON(http.StatusInternalServerError, gin.H{"message": fmt.Sprintf("failed to get schema: %s", outcome.Output)})
+			return
+		}
+
+		var schema SchemaResponse
+		var err error
+
+		connType := pb.ToConnectionType(conn.Type, conn.SubType.String)
```
sandromello: You're using this function 3 times for the same structure, just use a single time in the top of your handler and then re-use it here.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+			c.JSON(http.StatusInternalServerError, gin.H{"message": fmt.Sprintf("failed to parse schema: %v", err)})
+			return
+		}
+
+		c.JSON(http.StatusOK, schema)
+
+	case <-timeoutCtx.Done():
+		client.Close()
+		log.Infof("runexec timeout (50s), it will return async")
+		c.JSON(http.StatusRequestTimeout, gin.H{"message": "Request timed out"})
```
sandromello: The more appropriate status code is bad request for this case, because it failed to return the proper time defined by the server. A 408 status code is usually sent when the connection with the http client is lost which is not the case.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+		default:
+		}
+	}()
+
+	timeoutCtx, cancelFn := context.WithTimeout(context.Background(), time.Second*50)
+	defer cancelFn()
+	select {
+	case outcome := <-respCh:
+		if outcome.ExitCode != 0 {
+			c.JSON(http.StatusInternalServerError, gin.H{"message": fmt.Sprintf("failed to get schema: %s", outcome.Output)})
```
sandromello, line 620: There's a lot of conditions that may happen here which are not considered an internal error. I recommend to left the http status code as 400 for now and later on improving with more granular ones.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+
+	case <-timeoutCtx.Done():
+		client.Close()
+		log.Infof("runexec timeout (50s), it will return async")
+		c.JSON(http.StatusRequestTimeout, gin.H{"message": "Request timed out"})
+	}
+}
+
+// parseMongoDBSchema process the raw output from the command and organize it into a SchemaResponse structure
+func parseMongoDBSchema(output string) (SchemaResponse, error) {
```
sandromello: There's a lot of things going on here, could you provide some basic unit tests? Having tests will help us to catch errors throughout the lifecycle of this feature.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+	// Convert map to slice
+	for _, schema := range schemaMap {
+		response.Schemas = append(response.Schemas, *schema)
+	}
+
+	return response, nil
+}
+
+// parseSchemaOutput process the raw output from the command and organize it into a SchemaResponse structure
+func parseSQLSchema(output string, connType pb.ConnectionType) (SchemaResponse, error) {
```
sandromello: There's a lot of things going on here, could you provide some basic unit tests? Having tests will help us to catch errors throughout the lifecycle of this feature.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+//	@Produce		json
+//	@Param			nameOrID	path	string	true	"Name or UUID of the connection"
+//	@Param			database	path	string	true	"Name of the database"
+//	@Success		200			{object}	SchemaResponse
+//	@Failure		400,404,500	{object}	openapi.HTTPError
+//	@Router			/connections/{nameOrID}/databases/{database}/schema [get]
+func GetDatabaseSchema(c *gin.Context) {
+	ctx := storagev2.ParseContext(c)
+	connNameOrID := c.Param("nameOrID")
+	dbName := c.Query("database")
```
sandromello: You must validate this input to be a valid database, because it's being used to as the introspection query to avoid any injections.

I think something that validates lower, upper cases, `-`,  `_` and numbers should be enough

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+//
+//	@Summary		Get Database Schema
+//	@Description	Get detailed schema information including tables, views, columns and indexes
+//	@Tags			Core
+//	@Produce		json
+//	@Param			nameOrID	path	string	true	"Name or UUID of the connection"
+//	@Param			database	path	string	true	"Name of the database"
+//	@Success		200			{object}	SchemaResponse
+//	@Failure		400,404,500	{object}	openapi.HTTPError
+//	@Router			/connections/{nameOrID}/databases/{database}/schema [get]
```
sandromello: The route is incorrect here.

gateway/api/connections/connections.go
```diff
@@ -382,3 +391,532 @@ func FetchByName(ctx pgrest.Context, connectionName string) (*models.Connection,
...
+	IsPrimaryKey bool   `json:"is_primary_key"`
+	IsForeignKey bool   `json:"is_foreign_key"`
+}
+
+type Index struct {
+	Name      string   `json:"name"`
+	Columns   []string `json:"columns"`
+	IsUnique  bool     `json:"is_unique"`
+	IsPrimary bool     `json:"is_primary"`
+}
```
sandromello: API types should be in the openapi package: `gateway/api/openapi/openapi.go`. Make sure to document the types properly and give proper examples to them.

gateway/api/server.go
```diff
@@ -248,6 +248,14 @@ func (api *Api) buildRoutes(r *apiroutes.Router) {
 		r.AuthMiddleware,
 		api.TrackRequest(analytics.EventDeleteConnection),
 		apiconnections.Delete)
+	r.GET("/connections/:nameOrID/databases",
+		apiroutes.ReadOnlyAccessRole,
+		r.AuthMiddleware,
+		apiconnections.ListDatabases)
+	r.GET("/connections/:nameOrID/schema",
```
sandromello: Use plural here: `/schemas`

sandromello (changes requested):

gateway/api/connections/connections.go
rogefm: It's correct; that is what I found necessary for using Oracle in the past. We can change it in the future if needed.d.

sandromello (approved):

PR #579, opened 2024-11-29
chore(sentry) add sentry to webapp

mtmr0x:


.github/workflows/release.yml
```diff
@@ -55,6 +55,9 @@ jobs:
       - name: Set Git Tag
         run: echo "GIT_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
 
+      - name: Set webapp version
+        run: export WEBAPP_VERSION=$(awk -F"'" '{printf "%s", $2}' webapp/src/webapp/version.js)
```
sandromello: This step won't work as you expect.
- You must expose the environment variable by appending the key=value in the `$GITHUB_ENV` file. [Reference](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)
- It won't preserve the environment variable across jobs, therefore you need to add this step at the `publish-release` job

You'll need to extract the tarball to obtain the webapp version. I think using the version of the gateway (`GIT_TAG`) would be more appropriate and straightforward.

mtmr0x: Hey @sandromello thanks for raising these points.I will check your references to make this work.

Regarding using `GIT_TAG`, the webapp doesn't have a direct relation with this, we would have to always manually update the webapp version even when it has no updates to be related to the `GIT_TAG`. We might find a solution to this, but at the moment, I didn't. I'm happy to join a call to explain these limitations further and try to get some ideas from you on how to overcome them.

sandromello: So if you want to maintain this behavior, you'll need to extract the artifact and parse the version as you did and append into the file `$GITHUB_ENV`

mtmr0x: seems complicated to explain. I will talk to Rogerio and maybe we do it manually for every release of the webapp then

mtmr0x: oh got it! Yeah, I will do that and also do the same implementation on another line in this file for arm64, I just realized I'm doing it only for amd64.

sandromello (changes requested):
Any of these changes won't work @mtmr0x. The version won't be available in the `publish-release` step which requires this information.

You need to extract the version in the `publish-release` step instead. Let me know if you need help with it.

PR #597, opened 2024-12-10
fix(tracking) adjust identify to work with anon id

mtmr0x:
This adds the anon id to the identify method

### DIFF
--- a/gateway/api/localauth/register.go
+++ b/gateway/api/localauth/register.go
@@ -83,16 +83,17 @@ func Register(c *gin.Context) {
 		Name:   adminGroupName,
 	}
 	trackClient := analytics.New()
-	trackClient.MergeIdentifiedUserTrack(org.ID, user.Email, analytics.EventSingleTenantFirstUserCreated, nil)
 	trackClient.Identify(&types.APIContext{
 		OrgID: org.ID,
 		OrgName: org.Name,
 		UserName: user.Name,
 		UserID: user.Email,
+		UserAnonSubject: org.ID,
 		UserEmail: user.Email,
 		UserGroups: []string{adminGroupName},
 		ApiURL: appconfig.Get().ApiURL(),
 	})
+	trackClient.Track(user.Email, analytics.EventSingleTenantFirstUserCreated, nil)
 
 	err = models.InsertUserGroups([]models.UserGroup{adminUserGroup})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


gateway/api/localauth/register.go
```diff
@@ -83,16 +83,17 @@ func Register(c *gin.Context) {
 		Name:   adminGroupName,
 	}
 	trackClient := analytics.New()
-	trackClient.MergeIdentifiedUserTrack(org.ID, user.Email, analytics.EventSingleTenantFirstUserCreated, nil)
 	trackClient.Identify(&types.APIContext{
 		OrgID: org.ID,
 		OrgName: org.Name,
 		UserName: user.Name,
 		UserID: user.Email,
+		UserAnonSubject: org.ID,
```
sandromello, line 91: You already have this information in the `OrgID` attribute, why are you using this attribute to propagate the organization ID?

mtmr0x: it's intentional in this case because when the default org is created, that's the only identifiable information to later relate to the user signing up. It's weird, but I couldn't find an easier solution. If you have anything, please let me know

gateway/api/login/login.go
```diff
@@ -360,7 +360,17 @@ func syncSingleTenantUser(ctx *pguserauth.Context, uinfo idp.ProviderUserInfo) (
 		// When the first user is created, there's already an
 		// anonymous event tracked with his org id. We need to
 		// merge this anonymous event with the identified user
-		trackClient.MergeIdentifiedUserTrack(org.ID, uinfo.Email, analytics.EventSingleTenantFirstUserCreated, nil)
+		trackClient.Identify(&types.APIContext{
+			OrgID:      org.ID,
+			OrgName: org.Name,
+			UserName: uinfo.Profile,
+			UserID: uinfo.Email,
+			UserAnonSubject: org.ID,
```
sandromello, line 368: You already have this information in the `OrgID` attribute, why are you using this attribute to propagate the organization ID?

mtmr0x: same reason as comment above

gateway/api/localauth/register.go
sandromello: You're not using the `UserAnonSubject` attribute in the `Identify` method and you already have this information in the `OrgID` attribute. Would you mind explaining your intention?

mtmr0x: I'm using `UserAnonSubject` for the `AnonymousId`, I didn't find the `AnonymousId` in `APIContext` type, so that's why. The intention is explained above. Do you think I should extend `APIContext` to have `AnonymousID`?

sandromello (approved):

PR #656, opened 2025-01-24
add charts OCI registry push

thekoma:
This change will also publish the chart as OCI charts in ghcr.io registry. It will enable for gitops usage like FluxCD and/or ArgoCD


sandromello:
Thank you @thekoma for your contribution! I'm closing this PR in favor of #730 which addresses the same issue. We're planning to introduce a Contributor License Agreement (CLA) soon to accept external contributions from the community.

PR #660, opened 2025-01-28
[REFACTOR]: Show Connection string in the access flow credentials when is mongodb

rogefm:
### Gateway
- Added connection type and subtype to proxy manager endpoint;

### web app
- showing connection string to credentials of access flow;
Screenshot 2025-01-28 at 17 05 08
Screenshot 2025-01-28 at 17 05 01
Screenshot 2025-01-28 at 17 05 24
Screenshot 2025-01-28 at 17 05 20


gateway/pgrest/current_state.template.sql
```diff
@@ -291,7 +291,7 @@ $$ stable language sql;
 -- PROXYMANAGER
 --
 CREATE VIEW proxymanager_state AS
-    SELECT id, org_id, status, connection, port, access_duration, metadata, connected_at
+    SELECT id, org_id, status, connection, connection_type, connection_subtype, port, access_duration, metadata, connected_at
```
sandromello: When changing this file, it triggers a state migration in the `private.appstate` when the gateway is starting. We had some problems in the past regarding this process that we should avoid it, because these kind of problems are very hard to diagnose and provide a fix.

Therefore, we should avoid changing this file and refactor it by adding new gorm models to manage the proxymanager state resource.

gateway/api/proxymanager/proxymanager.go
```diff
@@ -43,13 +43,15 @@ func Get(c *gin.Context) {
...
-		RequestConnectionName: obj.RequestConnectionName,
-		RequestPort:           obj.RequestPort,
-		RequestAccessDuration: obj.RequestAccessDuration,
-		ClientMetadata:        obj.ClientMetadata,
-		ConnectedAt:           obj.ConnectedAt.Format(time.RFC3339),
+		ID:                       obj.ID,
+		Status:                   openapi.ClientStatusType(obj.Status),
+		RequestConnectionName:    obj.RequestConnectionName,
+		RequestConnectionType:    obj.RequestConnectionType,
+		RequestConnectionSubType: obj.RequestConnectionSubType,
```
sandromello: The proxy manager state table is a way of passing the state from the HTTP API to the gRPC layer. If the intention of this change is showing the connection type and subtype, storing these attributes are redundant because we're already have such information in the `connections` table.

Another option is propagating this information in the `metadata` attribute in the [proxy manager layer](https://github.com/hoophq/hoop/blob/main/gateway/transport/proxymanager.go#L125-L132). The `pctx` contains the context of the connection (type and subtype).

You just need to fetch the connection from the database and expose it.

sandromello (changes requested):

webapp/src/webapp/connections/views/connection_connect.cljs
```diff
@@ -149,26 +183,29 @@
...
+      [:> Button {:size "2"
+                  :variant "outline"
+                  :color "gray"
+                  :on-click #(js/window.open "https://install.hoop.dev")}
+       [:> Download {:size 16}]
+       " Download Desktop App"]
+      [:> Button {:size "2"
+                  :on-click (fn []
+                              (rf/dispatch [:connections->start-connect {:type (:connection_type connection-data)
+                                                                         :subtype (:connection_subtype connection-data)
```
sandromello: These are only read only fields, you don't need to send them.

gateway/api/openapi/types.go
```diff
@@ -667,6 +667,10 @@ type ProxyManagerResponse struct {
 	Status ClientStatusType `json:"status"`
 	// The requested connection name
 	RequestConnectionName string `json:"connection_name"`
+	// The requested connection type
+	RequestConnectionType string `json:"connection_type"`
+	// The requested connection subtype
+	RequestConnectionSubType string `json:"connection_subtype"`
```
sandromello: You need to mark them as read only with the tag: `readonly:"true"`.
Make sure also to generate the docs when these files are changed:

```sh
make generate-openapi-docs
```

sandromello (changes requested):

webapp/src/webapp/connections/views/connection_connect.cljs
rogefm: done

gateway/api/openapi/types.go
rogefm: done

sandromello (approved):

PR #679, opened 2025-02-06
[FEAT] Add DLP to server info

rogefm:
- Added DLP info to serverinfo endpoint
- Create two types of options for DLP in frontend


gateway/api/serverinfo/serverinfo.go
```diff
@@ -18,14 +18,26 @@ import (
...
+		dlpProvider := os.Getenv("DLP_PROVIDER")
+		switch dlpProvider {
+		case "presidio":
+			return isEnvSet("MSPRESIDIO_ANALYZER_URL") && isEnvSet("MSPRESIDIO_ANONYMIZER_URL")
+		case "gcp":
+			return isEnvSet("GOOGLE_APPLICATION_CREDENTIALS_JSON")
+		default:
+			return false
+		}
+	}()
```
sandromello: The `DLP_PROVIDER` env just make a selection between which provider to use, where gcp is the default option. Therefore, you shouldn't rely that this env is always going to be set. Returning `false` assumes the data masking isn't set, which is a wrong assumption.

For example, you could have data masking working by having this option as empty and `GOOGLE_APPLICATION_CREDENTIALS_JSON` set.

The right implementation based on your logic would be:
- Check if `MSPRESIDIO_` envs are set and DLP_PROVIDER is `mspresidio`
- Check if `GOOGLE_APPLICATION_CREDENTIALS_JSON` env is set

sandromello (changes requested):
- The assumption of your logic is not correct, please see my comment
- Don't use envs directly, rely in the appconfig.Get() implementing your logic in the appconfig package, then, use it in the serverinfo handler.

gateway/api/openapi/types.go
```diff
@@ -788,7 +788,9 @@ type ServerInfo struct {
 	AdminUsername string `json:"admin_username" example:"admin"`
 	// Auth method used by the server
 	AuthMethod string `json:"auth_method" enums:"oidc,local" example:"local"`
-	// Report if GOOGLE_APPLICATION_CREDENTIALS_JSON is set
+	// DLP provider used by the server
+	RedactProvider string `json:"redact_provider" enums:"gcp,presidio" example:"gcp"`
```
sandromello: It's `mspresidio`

gateway/appconfig/appconfig.go
```diff
@@ -146,6 +147,19 @@ func Load() error {
 		return fmt.Errorf("failed loading env TLS_CERT, reason=%v", err)
 	}
 
+	hasRedactCreds := func() bool {
+		dlpProvider := os.Getenv("DLP_PROVIDER")
+		if dlpProvider == "" || dlpProvider == "gcp" {
+			return isEnvSet("GOOGLE_APPLICATION_CREDENTIALS_JSON")
+		}
+
+		if dlpProvider == "presidio" {
```
sandromello: It's `mspresidio`

sandromello (changes requested):

sandromello (approved):

PR #692, opened 2025-02-22
feat: allow to set Docker agent env vars externally

mchristopher:
I'm working on setting up Hoop using AWS ECS (instead of EC2 instances), and I'm running into a problem running the system agent due to not being able to override `gateway:8010` with a valid DNS entry which can resolve.

This change allows you to pass in values for `SCHEME` and `GRPC_URL` if desired. Otherwise, the default behavior remains unchanged.

### DIFF
--- a/deploy/docker-compose/rootfs/usr/local/bin/run-agent.sh
+++ b/deploy/docker-compose/rootfs/usr/local/bin/run-agent.sh
@@ -14,11 +14,16 @@ INSERT INTO agents (id, org_id, name, mode, key_hash, status)
 COMMIT;
 EOF
 
-SCHEME=grpcs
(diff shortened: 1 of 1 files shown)


sandromello:
Hey @mchristopher , this file changes the image `hoophq/hooplabs` which is intended for evaluating Hoop with docker-compose. For production setup we recommend using the image `hoophq/hoopdev`.

The setup of the system agent is also optional and used for convenience. You could easily setup an additional agent.
See this [guide](https://hoop.dev/docs/concepts/agents) for more information.

If you still want to use a default agent with this image, you could use the `hoop-default-agent.sh` script:

```sh
docker run --rm -it --platform linux/amd64 hoophq/hoopdev  hoop-default-agent.sh
```

PR #738, opened 2025-03-26
Add global keyboard shortcut for search (⌘K / Ctrl+K) and update tooltip

andriosrobert:


webapp/src/webapp/core.cljs
```diff
@@ -18,8 +18,19 @@
...
+(defn add-global-keyboard-shortcuts []
+  (.addEventListener js/document "keydown" 
+    (fn [e]
+      (when (and (or (.-metaKey e) (.-ctrlKey e)) 
+                 (= (.-key e) "k"))
+        (.preventDefault e)
+        (let [search-input (.getElementById js/document "header-search")]
+          (when search-input
+            (.focus search-input)))))))
+
```
rogefm, line 30: The ideal would be to place this listener inside the Editor page and have it remove the listener when you leave the editor page. In other words, when you mount the component, you activate the listener; when you unmount it, you remove the listener.
It is not a good practice to leave it open forever.

rogefm (changes requested):

PR #786, opened 2025-04-14
Add bidirectional relationship between GuardRails and Connections

rogefm:
# Backend
This PR enhances the GuardRails API to support bidirectional relationships with Connections. Previously, Connections could reference GuardRails through their `GuardRailRules` array, but GuardRails couldn't directly reference their associated Connections.

## Changes
- Added functionality to associate connections with guardrails directly through the API
- Modified `GetConnectionIDsForGuardRail` to return connection IDs instead of names
- Enhanced `SyncGuardRailConnections` to properly handle both connection IDs and names
- Fixed HTTP status code in POST response to match API documentation (201 Created)
- Improved error handling and added more detailed logging
- Made sure all connections are properly synced when creating or updating a guardrail

## Testing
- Tested creating guardrails with connections (POST)
- Tested updating guardrails with connections (PUT)
- Verified that connection IDs are properly returned in API responses

## Technical Details
- Used the existing `guardrail_rules_connections` table for persistent storage
- Improved the robustness of connection ID resolution
- Enhanced error reporting to make debugging easier

# Frontend 
This PR adds support for associating connections with guardrails in the frontend, allowing users to select which connections a guardrail should be applied to. The feature includes a new component for connection selection and updates to the guardrail creation/editing flow.

### Changes

- Added `connection_ids` field support in guardrail state management
- Created a new component for selecting connections 
- Added API integration to fetch available connections
- Updated create/edit form to include connection selection section
- Added form handler for connection changes
- Ensured proper serialization and deserialization of connection IDs

### Technical Details

- The connection selection uses the existing multiselect component 
- Connection IDs are stored in the guardrail object and sent to the backend
- Selected connections are persisted when editing existing guardrails

### Screenshots

Screenshot 2025-04-14 at 16 49 13

### DIFF
--- a/gateway/api/guardrails/guardrails.go
+++ b/gateway/api/guardrails/guardrails.go
@@ -30,6 +30,9 @@ func Post(c *gin.Context) {
 		return
 	}
 
+	// Filter out empty connection IDs
+	validConnectionIDs := filterEmptyIDs(req.ConnectionIDs)
+
 	rule := &models.GuardRailRules{
 		ID:          uuid.NewString(),
 		OrgID:       ctx.GetOrgID(),
@@ -41,10 +44,10 @@ func Post(c *gin.Context) {
 		UpdatedAt:   time.Now().UTC(),
 	}
 
-	log.Infof("Creating guardrail %s with connections: %v", rule.Name, req.ConnectionIDs)
+	log.Infof("Creating guardrail %s with connections: %v", rule.Name, validConnectionIDs)
 
 	// Create guardrail and associate connections in a single transaction
-	err := models.UpsertGuardRailRuleWithConnections(rule, req.ConnectionIDs, true)
+	err := models.UpsertGuardRailRuleWithConnections(rule, validConnectionIDs, true)
 
 	switch err {
 	case models.ErrAlreadyExists:
@@ -88,6 +91,9 @@ func Put(c *gin.Context) {
 		return
 	}
 
+	// Filter out empty connection IDs
+	validConnectionIDs := filterEmptyIDs(req.ConnectionIDs)
+
 	rule := &models.GuardRailRules{
 		OrgID:       ctx.GetOrgID(),
 		ID:          c.Param("id"),
@@ -98,10 +104,10 @@ func Put(c *gin.Context) {
 		UpdatedAt:   time.Now().UTC(),
 	}
 
-	log.Infof("Updating guardrail %s with connections: %v", rule.ID, req.ConnectionIDs)
+	log.Infof("Updating guardrail %s with connections: %v", rule.ID, validConnectionIDs)
 
 	// Update guardrail and associate connections in a single transaction
-	err := models.UpsertGuardRailRuleWithConnections(rule, req.ConnectionIDs, false)
+	err := models.UpsertGuardRailRuleWithConnections(rule, validConnectionIDs, false)
 
 	switch err {
 	case models.ErrNotFound:
@@ -228,4 +234,15 @@ func parseRequestPayload(c *gin.Context) *openapi.GuardRailRuleRequest {
 		return nil
 	}
 	return &req
+}
+
+// Helper to filter out empty connection IDs
+func filterEmptyIDs(ids []string) []string {
+	result := make([]string, 0, len(ids))
+	for _, id := range ids {
+		if id != "" {
+			result = append(result, id)
+		}
+	}
+	return result
 }
--- a/gateway/api/guardrails/guardrails.go
+++ b/gateway/api/guardrails/guardrails.go
@@ -61,7 +61,7 @@ func Post(c *gin.Context) {
 			}
 
 			if len(validIDs) > 0 {
-				if err := models.SyncGuardRailConnections(ctx.GetOrgID(), rule.ID, validIDs); err != nil {
+				if err := models.SyncGuardRailConnectionAssociations(ctx.GetOrgID(), rule.ID, validIDs); err != nil {
 					log.Errorf("Failed to sync connections for guardrail %s: %v", rule.ID, err)
 					// Continue but we'll warn in the response
 				}
@@ -141,7 +141,7 @@ func Put(c *gin.Context) {
 			}
 
 			if len(validIDs) > 0 {
-				if err := models.SyncGuardRailConnections(ctx.GetOrgID(), rule.ID, validIDs); err != nil {
+				if err := models.SyncGuardRailConnectionAssociations(ctx.GetOrgID(), rule.ID, validIDs); err != nil {
 					log.Errorf("Failed to sync connections for guardrail %s: %v", rule.ID, err)
 					// Continue but we'll warn in the response
 				}
--- a/gateway/api/guardrails/guardrails.go
+++ b/gateway/api/guardrails/guardrails.go
@@ -61,21 +61,26 @@ func Post(c *gin.Context) {
 			}
 
 			if len(validIDs) > 0 {
+				log.Infof("Valid connection IDs to sync: %v", validIDs)
+
 				if err := models.SyncGuardRailConnectionAssociations(ctx.GetOrgID(), rule.ID, validIDs); err != nil {
 					log.Errorf("Failed to sync connections for guardrail %s: %v", rule.ID, err)
-					// Continue but we'll warn in the response
 				}
 			}
 		}
 
-		// Get connection IDs - this should now work properly
-		connectionIDs, err := models.GetConnectionIDsForGuardRail(ctx.GetOrgID(), rule.ID)
+		// Get the updated rule with connection IDs
+		updatedRule, err := models.GetGuardRailRules(ctx.GetOrgID(), rule.ID)
 		if err != nil {
-			log.Errorf("Error fetching connection IDs for guardrail %s: %v", rule.ID, err)
+			log.Errorf("Error fetching updated guardrail %s: %v", rule.ID, err)
+			c.JSON(http.StatusInternalServerError, gin.H{"message": "Failed to retrieve updated guardrail"})
+			return
 		}
 
-		log.Infof("Guardrail %s created successfully with %d connection IDs: %v",
-			rule.ID, len(connectionIDs), connectionIDs)
+		rule = updatedRule // Use the updated rule with connection IDs
+
+		log.Infof("Guardrail %s created successfully with %d connection IDs",
+			rule.ID, len(rule.ConnectionIDs))
 
 		c.JSON(http.StatusCreated, &openapi.GuardRailRuleResponse{
 			ID:            rule.ID,
@@ -83,7 +88,7 @@ func Post(c *gin.Context) {
 			Description:   rule.Description,
 			Input:         rule.Input,
 			Output:        rule.Output,
-			ConnectionIDs: connectionIDs,
+			ConnectionIDs: rule.ConnectionIDs,
 			CreatedAt:     rule.CreatedAt,
 			UpdatedAt:     rule.UpdatedAt,
 		})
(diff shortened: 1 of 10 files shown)


gateway/models/guadrails.go
```diff
@@ -40,6 +52,96 @@ func GetGuardRailRules(orgID, ruleID string) (*GuardRailRules, error) {
...
+	if err != nil {
+		log.Errorf("Error retrieving connection IDs for guardrail %s: %v", ruleID, err)
+		return nil, err
+	}
+
+	return connectionIDs, nil
+}
+
+// SyncGuardRailConnections updates the connections associated with a guardrail
+func SyncGuardRailConnections(orgID, ruleID string, connectionIDs []string) error {
```
sandromello: Change the name of this function `SyncGuardRailConnectionAssociations`.

I think this function has a lot of unnecessary logic, you could achieve the same thing with less code and more efficiently:

- To achieve the sync behavior you could just INSERT an association in the table only if the connection exists, you could achieve this by using a simple SQL statement.
- Don't make validation of parameters in this layer, you're already doing that in the API layer.
- Don't concatenate queries in GO, write the raw query or use gorm functions and pass  the necessary parameters accordingly
- Remove the log statements and make sure to return the error properly instead
- Use gorm transactions and void using raw transactions for rollback: (`DB.Transaction(func(tx *gorm.DB) error {})`)
- See the file `gateway/models/connections.go:178` for an example

gateway/models/guadrails.go
```diff
@@ -40,6 +52,96 @@ func GetGuardRailRules(orgID, ruleID string) (*GuardRailRules, error) {
...
+	var connectionIDs []string
+	err := DB.Raw(`
+		SELECT c.id 
+		FROM private.connections c
+		JOIN private.guardrail_rules_connections grc ON grc.connection_id = c.id
+		WHERE grc.org_id = ? AND grc.rule_id = ?
+	`, orgID, ruleID).Pluck("id", &connectionIDs).Error
+
+	if err != nil {
+		log.Errorf("Error retrieving connection IDs for guardrail %s: %v", ruleID, err)
```
sandromello: remove logging, just return the error and catch it in the next layer

gateway/api/openapi/types.go
```diff
@@ -1115,6 +1118,9 @@ type GuardRailRuleResponse struct {
 	*/
 	Output map[string]any `json:"output"`
 
+	// List of connection IDs or names that this guardrail applies to
+	ConnectionIDs []string `json:"connection_ids" example:"pgdemo,redis-prod"`
```
sandromello: The example is incorrect, it should have uuid instead of names. Make sure to generate the openapi documentation as well

gateway/api/openapi/types.go
```diff
@@ -1068,6 +1068,9 @@ type GuardRailRuleRequest struct {
 		}
 	*/
 	Output map[string]any `json:"output"`
+
+	// List of connection IDs or names that this guardrail applies to
+	ConnectionIDs []string `json:"connection_ids" example:"pgdemo,redis-prod"`
```
sandromello: The example is incorrect, it should has uuid instead of names. Make sure to generate the openapi documentation as well

gateway/api/guardrails/guardrails.go
```diff
@@ -158,14 +229,18 @@ func Get(c *gin.Context) {
 	case models.ErrNotFound:
 		c.JSON(http.StatusNotFound, gin.H{"message": "resource not found"})
 	case nil:
+		// Get connections for this rule
+		connectionIDs, _ := models.GetConnectionIDsForGuardRail(ctx.GetOrgID(), rule.ID)
```
sandromello: Don't hide the error, at least log to the stdout

gateway/api/guardrails/guardrails.go
```diff
@@ -125,16 +191,21 @@ func List(c *gin.Context) {
 		c.JSON(http.StatusInternalServerError, gin.H{"message": err.Error()})
 		return
 	}
+
 	rules := []openapi.GuardRailRuleResponse{}
 	for _, rule := range ruleList {
+		// Get connections for each rule
+		connectionIDs, _ := models.GetConnectionIDsForGuardRail(ctx.GetOrgID(), rule.ID)
```
sandromello: This is very bad because you're opening a connection for each record. I'm going to let this pass for now, but make sure to perform what you want in SQL and just map the data properly to the models.

What I recommend is refactoring the logic of Updating, Creating and Listing to include the connection ids instead of writing a separate function to fetch parts of this resources.

gateway/api/guardrails/guardrails.go
```diff
@@ -86,20 +120,52 @@ func Put(c *gin.Context) {
...
 	}
+
 	err := models.UpdateGuardRailRules(rule)
 	switch err {
 	case models.ErrNotFound:
 		c.JSON(http.StatusNotFound, gin.H{"message": err.Error()})
 		return
 	case nil:
+		// Sync connections
+		if len(req.ConnectionIDs) > 0 {
```
sandromello: The synchronous approach currently used presents a reliability issue since it's not integrated into the core methods (Update, Create, List). When failures occur, it may return an empty list of connection IDs without proper error handling. This inconsistent behavior makes troubleshooting difficult.
I recommend refactoring this endpoint to perform all operations within a single transaction, ensuring atomicity and better error handling. However, I leave the decision to you whether to merge the current implementation with its limitations and address these issues in a future enhancement, or to address them before merging.

sandromello (changes requested):

gateway/models/guadrails.go
```diff
@@ -71,7 +139,110 @@ func DeleteGuardRailRules(orgID, ruleID string) error {
...
+
+// UpsertGuardRailRuleWithConnections creates or updates a guardrail rule and its connections in a single transaction
+func UpsertGuardRailRuleWithConnections(rule *GuardRailRules, connectionIDs []string, isNew bool) error {
+	// Clean empty connection IDs
+	var validConnectionIDs []string
+	for _, id := range connectionIDs {
+		if id != "" {
+			validConnectionIDs = append(validConnectionIDs, id)
+		}
+	}
```
sandromello: Don't perform these kind of checks here, do it in the api layer.

gateway/models/guadrails.go
```diff
@@ -71,7 +139,110 @@ func DeleteGuardRailRules(orgID, ruleID string) error {
...
+			if res.Error != nil {
+				return res.Error
+			}
+			if res.RowsAffected == 0 {
+				return ErrNotFound
+			}
+		}
+
+		// 2. Delete existing connections
+		if err := tx.Exec("DELETE FROM "+tableGuardRailsConnections+" WHERE org_id = ? AND rule_id = ?",
```
sandromello: Don't concatenate SQL queries with string, use the name of the table instead

gateway/models/guadrails.go
```diff
@@ -71,7 +139,110 @@ func DeleteGuardRailRules(orgID, ruleID string) error {
...
+			rule.OrgID, rule.ID).Error; err != nil {
+			return err
+		}
+
+		// 3. Add new connections
+		for _, connNameOrID := range validConnectionIDs {
```
sandromello: You're naming the variables as connection id in every route and now they are names or ids. Does the frontend need them to be names or ids? In case they are only ids, remove any logic related to names.

gateway/models/guadrails.go
```diff
@@ -71,7 +139,110 @@ func DeleteGuardRailRules(orgID, ruleID string) error {
...
+				if errors.Is(err, gorm.ErrRecordNotFound) {
+					// Skip non-existent connections
+					continue
+				}
+				return err
+			}
+
+			// Add the association
+			err = tx.Exec(`
+				INSERT INTO `+tableGuardRailsConnections+` (id, org_id, rule_id, connection_id, created_at)
```
sandromello: Don't concatenate SQL queries with string, use the name of the table instead

gateway/models/guadrails.go
```diff
@@ -71,7 +139,110 @@ func DeleteGuardRailRules(orgID, ruleID string) error {
...
+
+		if err != nil {
+			return err
+		}
+
+		// Populate ConnectionIDs
+		rule.ConnectionIDs = make([]string, len(connections))
+		for i, conn := range connections {
+			rule.ConnectionIDs[i] = conn.ConnectionID
+		}
```
sandromello: You don't need any of this to construct the connection ids. As I said in the previous review: if you're able to save the connection ids without any issue, then you can just use these values as the response

gateway/models/guadrails.go
```diff
@@ -4,28 +4,73 @@ import (
...
+	Name          string         `gorm:"column:name"`
+	Description   string         `gorm:"column:description"`
+	Input         map[string]any `gorm:"column:input;serializer:json"`
+	Output        map[string]any `gorm:"column:output;serializer:json"`
+	CreatedAt     time.Time      `gorm:"column:created_at"`
+	UpdatedAt     time.Time      `gorm:"column:updated_at"`
+	ConnectionIDs []string       `gorm:"-"` // Not stored in DB, populated from join query
+}
+
+type GuardRailConnection struct {
```
sandromello, line 27: Where this type is being used for?

gateway/api/openapi/types.go
```diff
@@ -1068,6 +1068,9 @@ type GuardRailRuleRequest struct {
 		}
 	*/
 	Output map[string]any `json:"output"`
+
+	// List of connection IDs or names that this guardrail applies to
```
sandromello: These are not names, are only connection ids, rename this properly and generate the docs again please?

gateway/api/openapi/types.go
```diff
@@ -1115,6 +1118,9 @@ type GuardRailRuleResponse struct {
 	*/
 	Output map[string]any `json:"output"`
 
+	// List of connection IDs or names that this guardrail applies to
```
sandromello: These are not names, are only connection ids, rename this properly and generate the docs again please?

gateway/api/guardrails/guardrails.go
```diff
@@ -86,23 +97,32 @@ func Put(c *gin.Context) {
...
+
+	// Update guardrail and associate connections in a single transaction
+	err := models.UpsertGuardRailRuleWithConnections(rule, req.ConnectionIDs, false)
+
 	switch err {
 	case models.ErrNotFound:
 		c.JSON(http.StatusNotFound, gin.H{"message": err.Error()})
 		return
 	case nil:
+		log.Infof("Guardrail %s updated successfully with %d connection IDs",
```
sandromello: Remove this log statement please, it's not necessary since we're already auditing this endpoint in a middleware

gateway/api/guardrails/guardrails.go
```diff
@@ -86,23 +97,32 @@ func Put(c *gin.Context) {
 		Output:      req.Output,
 		UpdatedAt:   time.Now().UTC(),
 	}
-	err := models.UpdateGuardRailRules(rule)
+
+	log.Infof("Updating guardrail %s with connections: %v", rule.ID, req.ConnectionIDs)
```
sandromello: Remove this log statement please, it's not necessary since we're already auditing this endpoint in a middleware

gateway/api/guardrails/guardrails.go
```diff
@@ -40,22 +40,32 @@ func Post(c *gin.Context) {
...
 	case nil:
-		c.JSON(http.StatusOK, &openapi.GuardRailRuleResponse{
-			ID:          rule.ID,
-			Name:        rule.Name,
-			Description: rule.Description,
-			Input:       rule.Input,
-			Output:      rule.Output,
-			CreatedAt:   rule.CreatedAt,
-			UpdatedAt:   rule.UpdatedAt,
+		log.Infof("Guardrail %s created successfully with %d connection IDs",
```
sandromello: Remove this log statement please, it's not necessary since we're already auditing this endpoint in a middleware

gateway/api/guardrails/guardrails.go
```diff
@@ -40,22 +40,32 @@ func Post(c *gin.Context) {
 		CreatedAt:   time.Now().UTC(),
 		UpdatedAt:   time.Now().UTC(),
 	}
-	err := models.CreateGuardRailRules(rule)
+
+	log.Infof("Creating guardrail %s with connections: %v", rule.Name, req.ConnectionIDs)
```
sandromello: Remove this log statement please, it's not necessary since we're already auditing this endpoint in a middleware

sandromello (changes requested):

gateway/models/guadrails.go
rogefm: The GuardRailConnection type represents the many-to-many relationship between guardrail rules and connections in the database. It's used for creating and deleting the association records in the private.guardrail_rules_connections table during the UpsertGuardRailRuleWithConnections operation.

gateway/api/openapi/autogen/docs.go
```diff
@@ -4647,6 +4647,17 @@ const docTemplate = `{
                 "name"
             ],
             "properties": {
+                "connection_ids": {
+                    "description": "List of connection IDs or names that this guardrail applies to",
```
sandromello: You need to regenerate the docs to update this

sandromello (approved):

PR #795, opened 2025-04-22
Implementation of complete CRUD for user groups

rogefm:
Implements complete CRUD operations for the user groups entity. Previously, only the LIST operation was available, and groups could only be created/updated indirectly through user updates. Now, groups can be managed directly via a dedicated API.

## Changes
- Added CREATE operation: Creates a new group (`POST /api/users/groups`)
- Added UPDATE operation: Updates an existing group (`PUT /api/users/groups/{name}`)
- Added DELETE operation: Removes an existing group (`DELETE /api/users/groups/{name}`)
- Implemented `UserGroup` model in OpenAPI for standardization
- Implemented function to create a group without associating it with a user

## Motivation
Allow direct management of user groups via API, facilitating:
- Automation of group creation/modification
- Pre-creation of groups before assignment to users
- Group cleanup/maintenance operations

## How to test
Curl examples to test the API:

```bash
# List all groups
curl -X GET 'http://localhost:8009/api/users/groups' \
  -H "Authorization: Bearer $TOKEN"

# Create a new group
curl -X POST 'http://localhost:8009/api/users/groups' \
  -H "Authorization: Bearer $TOKEN" \
  -H 'Content-Type: application/json' \
  -d '{"name": "engineering"}'

# Update an existing group
curl -X PUT 'http://localhost:8009/api/users/groups/engineering' \
  -H "Authorization: Bearer $TOKEN" \
  -H 'Content-Type: application/json' \
  -d '{"name": "engineering-v2"}'

# Delete a group
curl -X DELETE 'http://localhost:8009/api/users/groups/engineering-v2' \
  -H "Authorization: Bearer $TOKEN"
```

## Technical details
- Added UserGroup model in the `gateway/api/openapi/types.go` file
- Implemented new controller functions: `CreateGroup`, `UpdateGroup`, and `DeleteGroup`
- Added new model methods: `CreateUserGroupWithoutUser`, `UpdateUserGroupName`, and `DeleteUserGroup`
- Added new routes in the `gateway/api/server.go` file
- All operations include OpenAPI documentation for Swagger


gateway/api/user/user.go
```diff
@@ -581,6 +581,177 @@ func ListAllGroups(c *gin.Context) {
...
+		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed getting org groups"})
+		return
+	}
+
+	for _, ug := range userGroups {
+		if ug.Name == req.Name {
+			c.JSON(http.StatusConflict, gin.H{"message": fmt.Sprintf("group %s already exists", req.Name)})
+			return
+		}
+	}
```
sandromello: You're obtaining all user groups to validate if a single group exists. You could make this more simpler by removing the constraint in the underline function,  create the group and catch the unique constraint error gorm.ErrDuplicatedKey. See this function for more details: `UpsertGuardRailRuleWithConnections`

gateway/api/user/user.go
```diff
@@ -581,6 +581,177 @@ func ListAllGroups(c *gin.Context) {
...
+//	@Produce		json
+//	@Param			name	path	string	true	"The name of the group to update"
+//	@Param			request	body	openapi.UserGroup	true	"Group object with new name"
+//	@Success		200		{object}	openapi.UserGroup	"{"name": "string"}"
+//	@Failure		400		{object}	openapi.HTTPError	"{"message": "error message"}"
+//	@Failure		404		{object}	openapi.HTTPError	"{"message": "group not found"}"
+//	@Failure		409		{object}	openapi.HTTPError	"{"message": "group already exists"}"
+//	@Failure		500		{object}	openapi.HTTPError	"{"message": "server error"}"
+//	@Router			/users/groups/{name} [put]
+func UpdateGroup(c *gin.Context) {
```
sandromello: This route will perform a delete and insert operation which doesn't comply with the REST method PUT.
The `name` attribute has a unique constraint meaning it's used as a unique identifier and you're permitting to change this unique constraint. Therefore, in practice you're going to remove the record and add a new one.

I think it will be better to remove this route and allow clients to perform a DELETE and POST operation in this case which is more explicit.

gateway/api/user/user.go
```diff
@@ -581,6 +581,177 @@ func ListAllGroups(c *gin.Context) {
...
+		}
+	}
+
+	if !groupExists {
+		c.JSON(http.StatusNotFound, gin.H{"message": fmt.Sprintf("group %s not found", name)})
+		return
+	}
+
+	// Delete all instances of this group
+	if err := models.DeleteUserGroup(ctx.OrgID, name); err != nil {
```
sandromello, line 644: This function could just return models.ErrNotFound in case it didn't returned any rows in the delete operation

gateway/api/user/user.go
```diff
@@ -581,6 +581,177 @@ func ListAllGroups(c *gin.Context) {
...
+//	@Failure		404		{object}	openapi.HTTPError	"{"message": "group not found"}"
+//	@Failure		422		{object}	openapi.HTTPError	"{"message": "cannot delete built-in group"}"
+//	@Failure		500		{object}	openapi.HTTPError	"{"message": "server error"}"
+//	@Router			/users/groups/{name} [delete]
+func DeleteGroup(c *gin.Context) {
+	ctx := storagev2.ParseContext(c)
+	name := c.Param("name")
+
+	// Prevent deletion of built-in groups
+	if name == types.GroupAdmin || name == types.GroupAuditor {
```
sandromello: I think you could allow the user to delete the auditor group, there's no issue on allowing users to remove this group/role. 

The admin is indeed a problem, because users would lose access to the system when deleting the administrator role.

sandromello (changes requested):

gateway/api/user/user.go
rogefm: I implemented the suggested improvement to make the code more efficient.

It wasn't possible to use the same duplicate key error pattern (gorm.ErrDuplicatedKey) as in UpsertGuardRailRuleWithConnections, because the user_groups table doesn't have a unique constraint for the (org_id, name) combination. However, the implemented approach is equally efficient and has the same behavior.

sandromello (approved):

PR #800, opened 2025-04-25
Add Support for do_not_track Flag

rogefm:
This PR supports the `do_not_track` flag from the serverinfo API, allowing complete disabling of all tracking and analytics tools when enabled.

### Core Implementation
- Added subscription for `do_not_track` flag from serverinfo API response
- Removed hardcoded tracking scripts from `index.html`
- Created a dynamic tracking initialization system that only loads scripts when tracking is allowed
- Implemented a centralized mechanism for tracking management

### Tracking Tools Integration
- **Segment**: Conditionally loads and initializes based on `do_not_track` setting
- **Clarity**: Added safe verification before calls and respects the tracking preference
- **Intercom**: Updated initialization to check tracking preferences
- **Canny**: Modified user identification to happen only when tracking is allowed
- **Sentry**: Delayed initialization until tracking preferences are known

## Testing
The changes have been tested with both `do_not_track: true` and `do_not_track: false` settings, confirming:
- No tracking scripts are loaded when disabled
- Network requests to tracking services are properly prevented
- Application functions normally in both modes

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -67,6 +67,6 @@ POSTGRES_DB_URI=postgres://hoopdevuser:1a2b3c4d@host.docker.internal:5449/hoopde
 DISABLE_SESSIONS_DOWNLOAD=false
 
 # Tracking control
-# Set to 'true' to enable analytics and tracking scripts on the frontend (default)
-# Set to 'false' to disable all analytics and tracking scripts
-TRACKING=true
+# Set to 'true' to disable all analytics and tracking scripts on the frontend
+# Set to 'false' to allow analytics and tracking scripts (default)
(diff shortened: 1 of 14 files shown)


sandromello (changes requested):
Hey @rogefm , as we spoke before. Let's use a better naming for the environment variable and use enums.

Suggestions:
- `ANALYTICS_TRACKING=disabled`
- `ANALYTICS=disabled`

You must also add this value to the helm chart, see `./deploy/helm-chart/chart/gateway/templates/secret-configs.yaml`

sandromello (approved):

PR #819, opened 2025-05-08
Runbook Editor Improvements: Enhanced UI Responsiveness and Connection Management

rogefm:
This PR refactors the Runbooks functionality to improve user experience when working with connections that have specific access mode restrictions. It focuses on better UI responsiveness, proper permission handling, and enhanced layout management.

### Connection Management and Permission Handling
- Added permission checks for determining when connections should be available for web terminal or native client access
- Implemented connection filtering logic to exclude specific subtypes (`tcp` and `httpproxy`) and validate access modes
- Added functionality to automatically select the appropriate interface (runbooks editor or SQL editor) based on connection permissions

### UI Responsiveness Improvements
- Implemented proper button state management in header to indicate when actions are unavailable
- Added `cursor-not-allowed` styling for disabled elements to improve visual feedback
- Created loading overlay for runbook transitions to provide better feedback during operations

### Layout Enhancements
- Refactored side panel component to use Allotment for improved layout and responsiveness
- Improved the directory tree structure in runbooks list for better navigation
- Enhanced panel transitions and management to reduce UI flickering during state changes

### Code Optimization
- Removed unnecessary code in panel.cljs to improve performance
- Simplified schema processing for improved editor performance
- Optimized how runbooks are loaded and displayed when connection permissions change

## Testing

The changes have been tested across different connection types and permission configurations to ensure a consistent user experience:
- Connections with only runbooks enabled
- Connections with only exec enabled
- Connections with both runbooks and exec enabled
- Empty connection states

## Notes for Reviewers

The main focus of this PR is to ensure a seamless experience when working with connections that have restricted access modes, particularly focusing on the "runbooks only" mode. Pay special attention to how the UI responds when switching between different connection types.

### DIFF
--- a/webapp/src/webapp/connections/views/connection_list.cljs
+++ b/webapp/src/webapp/connections/views/connection_list.cljs
@@ -79,6 +79,30 @@
          (when (= selected-resource value)
            [:> Check {:size 16}])])]]]])
 
+(defn can-connect? [connection]
+  (not (and (= "disabled" (:access_mode_runbooks connection))
+            (= "disabled" (:access_mode_exec connection))
+            (= "disabled" (:access_mode_connect connection)))))
+
+(defn can-open-web-terminal? [connection]
+  (if-not (#{"tcp" "httpproxy"} (:subtype connection))
+
+    (if (or (= "enabled" (:access_mode_runbooks connection))
+            (= "enabled" (:access_mode_exec connection)))
+      true
+      false)
+
+    false))
+
+(defn can-open-native-client? [connection]
+  (if (#{"database" "application"} (:type connection))
+
+    (if (= "enabled" (:access_mode_connect connection))
+      true
+      false)
+
+    false))
+
 (defn panel [_]
   (let [connections (rf/subscribe [:connections])
         user (rf/subscribe [:users->current-user])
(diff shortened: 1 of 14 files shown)


webapp/src/webapp/connections/views/connection_list.cljs
```diff
@@ -79,6 +79,30 @@
...
+    false))
+
+(defn can-open-native-client? [connection]
+  (if (#{"database" "application"} (:type connection))
+
+    (if (= "enabled" (:access_mode_connect connection))
+      true
+      false)
+
+    false))
```
sandromello: Hey @rogefm we have native SSH which currently doesn't allow access to the web console.
I try to maintain the list of current features updated as best as I can in the Connection [Quick Start documentation](https://hoop.dev/docs/quickstart/ssh). It could be a good place to consult before implementing these conditions.

Another option that we could do is allowing the server to inform the available connection types and not rely in this static information.

sandromello (approved):

webapp/src/webapp/connections/views/connection_list.cljs
rogefm: It would be great! Avoiding business rules on the front end is always a good idea.
And I will fix this rule right now

PR #845, opened 2025-05-19
Remove unused notification block from Helm values

lekspyl:
A search over the Helm chart shows that the`.Values.config.notification` map is declared but not used anywhere in the templates. It can be removed to avoid confusion.


lekspyl:

PR #869, opened 2025-06-02
feat: Allow setting arbitrary max JIT session duration

lekspyl:
Currently max sesstion duration is hardcoded to 48 hours. Users may want to set custom value.
I'm adding a new `MAX_ACCESS_DURATION_HOURS` environment variable and using it in the `review.go` plugin. I'm not sure whether the `review.oss.go` file is used, so I've kept it intact.


mtmr0x:
hey @lekspyl sorry for hanging you here for so long. I will take this to the team and we will provide you with some feedback

rogefm:
We've developed a solution for this; please update to the new version to access this feature :)

PR #987, opened 2025-08-14
[CHORE] the roadblock removal initiative

mtmr0x:
## 📝 Description

This is a work in progress/draft of the simplification work the engineering team will work on. The changes here are not final but follow some principles:

- Gateway is a web service; it shouldn’t be run from inside the CLI.
- When setting the gateway, no one should need a preloaded agent to be able to move forward in the first steps of the UI.
- It should just run if it can find a hoop database with user and pass `postgres` for simplicity's sake;
- It should only break if it can’t find the database; otherwise, it should run without any issues or lack of config anywhere
- Temporary files configurations should be amicable for the development environment and set differently for the production environment. Users don’t need to worry about OS user permissions when executing a session anywhere, unless they choose to (by explicitly configuring the path for it).
- These changes might affect our infrastructure files, at least a little bit (hopefully it will simplify them by a lot), we should review them as well.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [x] 🔧 Build configuration change
- [x] 🧹 Chore

## 📋 Changes Made

- WIP
- 
- 

### Test Configuration:
- **Browser(s)**: 
- **OS**:

### Tests performed:
- [ ] Unit tests pass
- [ ] Integration tests pass

## ✅ Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have checked my code and corrected any misspellings

## 📄 Additional Notes

---

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -12,7 +12,7 @@ This is the Hoop Gateway folder. Below are the steps to run it in development. F
 
 ## Dependencies
 
-- Install Golang 1.20 or later: [https://go.dev/doc/install](https://go.dev/doc/install)
+- Install Golang 1.23 or later: [https://go.dev/doc/install](https://go.dev/doc/install)
 - A PostgreSQL anywhere you want.
 
 ## Setup


mtmr0x:

webapp/resources/public/images/command-line-dark.svg
rogefm, line 1: You don't need to include these images here; we already have them in /icons.

webapp/src/webapp/onboarding/setup.cljs
```diff
@@ -82,7 +82,8 @@
    [:> Box {:class "max-w-[600px] text-center space-y-6"}
     [:img {:src (str config/webapp-url "/images/hoop-branding/PNG/hoop-symbol_black@4x.png")
            :class "w-16 mx-auto py-4"}]
-    [:> Spinner {:class "justify-self-center"}]
+    [:div {:class "flex justify-center items-center"}
```
rogefm:
```suggestion
    [:> Flex {:align "center" :justify "center"}
```

webapp/src/webapp/onboarding/setup_agent.cljs
```diff
@@ -0,0 +1,162 @@
...
+                  :subtitle "Define basic identification properties to create your new Agent."
+                  :value "agent-information"
+                  :disabled false
+                  :avatar-icon [:> Info {:size 16}]
+                  :show-icon? (= (:status @agent-key) :ready)
+                  :content [:> Grid {:columns "7" :gap "7"}
+                            [:> Box {:gridColumn "span 2 / span 2"}
+                             [:> Heading {:size "4" :weight "medium" :as "h3"}
+                              "Set an Agent name"]
+                             [:p {:class "text-sm text-gray-500"}
```
rogefm:
```suggestion
                             [:> Text {:as "p" :size "2" :class "text-[--gray-7]"}
```

webapp/src/webapp/onboarding/setup_agent.cljs
```diff
@@ -0,0 +1,162 @@
...
+                  :value "installation-method"
+                  :avatar-icon [:> ListOrdered {:size 16}]
+                  :disabled (not (= (:status @agent-key) :ready))
+                  :subtitle "Get Agent deployment details for your preferred method."
+                  :content [:> Flex {:direction "column" :gap "8"}
+                            [:> Grid {:columns "7" :gap "7"}
+                             [:> Box {:gridColumn "span 2 / span 2"}
+                              [:> Heading {:size "4" :weight "medium" :as "h3"}
+                               "Installation method"]
+                              [:p {:class "text-sm text-gray-500"}
```
rogefm:
```suggestion
                              [:> Text {:as "p" :size "2" :class "text-[--gray-7]"}
```

webapp/src/webapp/onboarding/setup_agent.cljs
```diff
@@ -0,0 +1,162 @@
...
+                               "Select the type of environment to setup the agent in your infrastructure."]]
+                             [:> Box {:class "space-y-radix-7" :grid-column "span 5 / span 5"}
+                              [:> Flex {:direction "column" :gap "3"}
+                               (doall
+                                 (for [method installation-methods]
+                                   [:div {:key (:title method)
```
rogefm:
```suggestion
                                   [:> Box {:key (:title method)
```

webapp/resources/public/images/command-line-dark.svg
mtmr0x: these specific weren't, plus they need to follow the same identity of Docker's and Kubernete's icons, so I had to make them specifically for this view

gateway/main.go
```diff
@@ -45,7 +46,7 @@ func Run() {
 		log.Fatalf("failed loading gateway configuration, reason=%v", err)
 	}
 	if err := webappjs.ConfigureServerURL(); err != nil {
-		log.Fatal(err)
+		log.Warnf("failed configuring webappjs server URL, running gateway without it, reason=%v", err)
```
sandromello, line 49: We should not let errors pass silently. Troubleshooting this kind of errors in production or in proof of concepts is very hard. A container crashing is better than trying to figure out why a blank screen is appearing when accessing the Webapp.

Since the goal is allowing this error pass for development environments, you could use the value unknown from the common version package, which means the build is not generated from a build pipeline. Example:

```go
import "github.com/hoophq/hoop/common/version"
(...)

if isProductionBuild := version.Get().Version != "unknown"; isProductionBuild {
  if err := webappjs.ConfigureServerURL(); err != nil {
    log.Fatal(err)
  }
}
```

gateway/README.md
```diff
@@ -0,0 +1,48 @@
...
+<p align="center"> 🔒 Secure infrastructure access without complexity or cost 
+<br /> <br />
+
+## Hellow world
+
+This is the Hoop Gateway folder. Below are the steps to run it in development. For installation and usage instructions, please refer to the [Hoop documentation](https://hoop.dev/docs).
+
+## Dependencies
+
+- Install Golang 1.20 or later: [https://go.dev/doc/install](https://go.dev/doc/install)
```
sandromello: Recommend GO 1.23 and later

gateway/appconfig/appconfig.go
```diff
@@ -242,6 +243,9 @@ func hasIdpEnvs() bool {
 
 func loadPostgresCredentials() (*pgCredentials, error) {
 	pgConnectionURI := os.Getenv("POSTGRES_DB_URI")
+	if pgConnectionURI == "" {
+		pgConnectionURI = "postgres://postgres:postgres@localhost:5432/hoop?sslmode=disable"
+	}
```
sandromello, line 248: Since the goal of this PR is having an environment for development, I recommend using a check to validate if it's a development build. This will only add this default for these environments.

> Having a default value for production may lead to unwanted behaviors that are hard to troubleshoot on proof of concepts or production environments.

Example:

```go
import "github.com/hoophq/hoop/common/version"
(...)

isDevBuild := version.Get().Version == "unknown"
if isDevBuild && pgConnectionURI == "" {
  pgConnectionURI = "postgres://postgres:postgres@localhost:5432/hoop?sslmode=disable"
}

sandromello (changes requested):

gateway/main.go
mtmr0x: `isProd`, `isDev` kinda env vars lead to snowflake behavior instead of consistent configuration. We shouldn't do it. The same applies to the other comment regarding the `pgConnectionURI` value. Our recipes, images, and deployment builds should make sure they have the webapp or the configs that correspond to the proper environment, and the app shouldn't have an opinion on configuration

sandromello: I want to make sure we're aligned on the trade-offs here. This change could potentially increase our MTTR for troubleshooting issues in both POCs and production environments.

sandromello (approved):
This change will increase MTTR due to these production defaults. Please proceed if you're comfortable with this tradeoff.

PR #991, opened 2025-08-20
feat: Add new commands for start, stop, remove and logs for a service daemon

matheusfrancisco:
## 📝 Description
This PR implements new commands for the client. 
This PR will introduce the hoop agent commands with the subcommands start, stop, remove, and logs. The start command creates a daemon service for the agent and start the agent, the stop subcommand will stop the daemon, the remove will stop and delete the deamon, and the logs will block the command and will append in the stdot the logs from the daemon.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [x] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- Added the command hoop agent start: creates a agent daemon in the linux and starts
- Added the command hoop agent stop: stops the agent daemon
- Added the commnad hoop agent remove: remove the agent daemon
- Added the command hoop agent logs: return the logs for the agent daemon
- 

### Test Configuration:
- **Browser(s)**: 
- **OS**: Linux

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

image

## ✅ Checklist

## 📄 Additional Notes

---


client/cmd/install.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	opts := systemd.Options{
+		ServiceName: "hoop-agent",
+		ExecPath:    exe,
+		ExecArgs:    " start agent",
+		Env: map[string]string{
+			"HOOP_KEY": cfg.Token,
+		},
+		UserMode: true,
```
matheusfrancisco: idk if I get this from flags, for example if the user wants to put the systemd service in the etc/ instead of .config/systemd/user 
So the command would be `hoop install systemd agent --user-mode root` and then this needs to be running with sudo, the default would be user

client/cmd/install.go
```diff
@@ -0,0 +1,114 @@
...
+		WantedBy: "multi-user.target",
```
matheusfrancisco: same for this, we have the dafault.target and multi-user.target, would the user be able to select this config as well?

 multi-user.target
- 	Service starts for the whole machine after boot.

default.target
- Service starts when the user logs in (or after logout, too, if linger is enabled).

client/cmd/install.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return fmt.Errorf("get executable path: %w", err)
+	}
+	exe, err = filepath.EvalSymlinks(exe)
+	if err != nil {
+		return fmt.Errorf("resolve executable symlinks: %w", err)
+	}
+
+	opts := systemd.Options{
+		ServiceName: "hoop-agent",
```
matheusfrancisco: Do we want to allow the user to pass the service name? `hoop install systemd agent --service-name hoop`

client/cmd/systemd/systemd.go
```diff
@@ -0,0 +1,176 @@
...
+	ServiceName string
+	ExecPath    string
+	ExecArgs    string
+	Env         map[string]string
+	UserMode    bool
+	UnitPath    string
+	WantedBy    string
+}
+
+func Install(opts Options) error {
```
matheusfrancisco: another point to have in mind is whether we want this to be installed and start, or if the user will have to run
`hoop install systemd agent` and then `hoop start systemd` 

Splitting this into two commands makes each command have only one responsibility. On the other hand, this will add the complexity of running two commands, which is not too bad

client/cmd/systemd/systemd.go
```diff
@@ -0,0 +1,176 @@
...
+
+[Service]
+Type=simple
+%sExecStart=%s%s
+Restart=on-failure
+RestartSec=3
+StandardOutput=journal
+StandardError=journal
+# In user units, don't set User/Group. In system units, prefer a dedicated user.
+# ProtectHome/read-only may block reading configs in $HOME; enable only if safe.
```
matheusfrancisco:
```suggestion
```

client/cmd/install.go
```diff
@@ -0,0 +1,114 @@
...
+		ExecPath:    exe,
+		ExecArgs:    " start agent",
+		Env: map[string]string{
+			"HOOP_KEY": cfg.Token,
```
matheusfrancisco: Currently, if the user does not have the environment variable HOOP_KEY exported, we will fail with a missing HOOP_KEY error. As we discussed, we could block and wait for the user to pass the HOOP_KEY. Is that the expected behavior for this flow, or do we want to keep it as the other flow?

client/cmd/install.go
matheusfrancisco: ans: user level only

client/cmd/install.go
matheusfrancisco: ans: default

client/cmd/systemd/systemd.go
matheusfrancisco: ans: only one command

client/cmd/install.go
```diff
@@ -0,0 +1,114 @@
...
+
+	agentconfig "github.com/hoophq/hoop/agent/config"
+	"github.com/hoophq/hoop/client/cmd/systemd"
+	"github.com/hoophq/hoop/common/log"
+	"github.com/hoophq/hoop/common/version"
+	"github.com/spf13/cobra"
+)
+
+var exampleInstallCmd = `
+hoop install systemd agent
```
matheusfrancisco: - [x] move the commands for:

hoop agent start
hoop agent stop
hoop agent remove

client/cmd/systemd/linux.go
```diff
@@ -0,0 +1,193 @@
...
+	if err := execRunner.Logs("journalctl", args...); err != nil {
+		return fmt.Errorf("journalctl --user -u %s failed: %v\n", serviceName, err)
+	}
+	return nil
+}
+func StartLinuxAgent() error {
+	cfg, err := agentconfig.Load()
+
+	if err != nil {
+		log.With("version", vi.Version).Fatal(err)
```
sandromello: It's best to return the error to the caller instead of dying here.

client/cmd/systemd/linux.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return nil
+}
+
+func remove(serviceName string) error {
+	unitPath := userPaths(Options{ServiceName: serviceName})
+
+	if out, err := execRunner.Run("systemctl", "--user", "disable", "--now", serviceName); err != nil {
+		_ = out
+	}
```
sandromello: I think you forget to return the error here. You could omit the out if you're not using it as well

```go
if _, err := execRunner.Run(...); err != nil {
  return err
}
```

client/cmd/systemd/linux.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if out, err := execRunner.Run("systemctl", "--user", "daemon-reload"); err != nil {
+		return fmt.Errorf("systemctl daemon-reload failed: %v\n%s", err, out)
+	}
+	return nil
+}
+
+func userPaths(opts Options) (unitPath string) {
+	home, _ := os.UserHomeDir()
```
sandromello: Be aware that this could fail and you'll have a wrong unit path

client/cmd/agent.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/hoophq/hoop/client/cmd/systemd"
+	"github.com/spf13/cobra"
+)
+
+var exampleAgentStartCmd = `
+hoop agent install 
+`
+var exampleRemoveCmd = `
+hoop remove systemd agent
+`
```
sandromello: Is this example correct? I think it's `hoop agent remove` right?

matheusfrancisco: 🤦🏽 I forgot to change

mtmr0x (approved):

PR #1004, opened 2025-09-02
Add SSH proxy implementation on gateway

sandromello:
## 📝 Description

Introduces SSH proxy in the gateway

- Client authenticates with temporary credentials over Keyboard-interactive Authentication
- Allows connecting into multiple remote hosts from a single port
- The feature is turned off by default and enabled once an administrator enable it via API
    - It allows configuring the default encryption key. In case none is provide a new one is generated
    - Disabling and enabling the server will impact on clients receiving known hosts identification changed in case the hosts key is lost or changed.
    - It allows configuring the listen address of the proxy

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- **BREAKING CHANGE:** Refactor dbaccess endpoint and resources to connection credentials
- Rename table connection_dbaccess to connection_credentials
- Allow starting and stopping the SSH server proxy via API
- Accept SSH encryption key via API

## 🧪 Testing

- Tested navigating via terminal session using `ssh`
- Tested using `scp` utility
- Tested with port forwarding


gateway/api/connectioncredentials/connection_credentials.go
```diff
@@ -0,0 +1,179 @@
+package apiconnectioncredentials
```
mtmr0x: if this belongs to the connections, why don't we let this file inside the `apiconnections` package in  `credentials.go` file?

common/keys/keys.go
```diff
@@ -115,6 +117,27 @@ func Base64DecodeEd25519PublicKey(encodedKey string) (ed25519.PublicKey, error)
...
+	if err != nil {
+		return nil, err
+	}
+	ed25519Key, ok := privateKey.(*ed25519.PrivateKey)
+	if !ok {
+		return nil, fmt.Errorf("key is not Ed25519, found=%T", privateKey)
+	}
+	return *ed25519Key, nil
+}
+
```
mtmr0x: if this is not being used by the agent, I would discourage adding new things to this `common`, specially because this is being used only in one place. It looks like an abstraction being added too soon

gateway/api/connectioncredentials/connection_credentials.go
```diff
@@ -0,0 +1,179 @@
...
+
+type credentialsInfo struct {
+	serverHostname   string
+	serverPort       string
+	username         string
+	secretKey        string
+	secretKeyHash    string
+	databaseName     *string
+	connectionString string
+}
```
mtmr0x: I see this is being used for both postgres and ssh, I would go with a map-like config using some generics for each type of connection so this can evolve sanely

sandromello: This attribute is being used as a utility struct only. It will be used to propagate the information to the openapi type.

Using maps or generics are more difficult to deal with than using a known type. I think we could left this struct as is and refactor if it's needed in the future. All database protocols will have the same information, therefore I don't see how this could evolve beyond these attributes.

mtmr0x: I'd say the difficulty pays off in clarity, but I understand. Thanks

gateway/api/serverconfig/misc.go
```diff
@@ -62,10 +65,19 @@ func GetServerMisc(c *gin.Context) {
 		}
 	}
```
EmanuelJr, line 70: I would put some comments just to split the `ssh` and the `postgres` logic.

gateway/api/serverconfig/misc.go
```diff
@@ -196,3 +256,63 @@ func forbiddenOnMultiTenantSetups(c *gin.Context) (forbidden bool) {
...
+	if newState != nil && newState.SSHServerConfig != nil {
+		newConf = *newState.SSHServerConfig
+	}
+
+	switch {
+	// stop instance when new configuration is empty
+	case newConf.ListenAddress == "":
+		return newConf, "stop"
+	// restart on configuration drift
+	case currentConf.ListenAddress != newConf.ListenAddress || currentConf.HostsKey != newConf.HostsKey:
```
EmanuelJr:
```
case currentConf.ListenAddress != newConf.ListenAddress, currentConf.HostsKey != newConf.HostsKey:
```
or
```
case currentConf.ListenAddress != newConf.ListenAddress:
case currentConf.HostsKey != newConf.HostsKey:
```
 looks a bit more like `switch/case` style

EmanuelJr (commented):
Just some small changes.
I'll review the `gateway/proxyproto/sshproxy/sshproxy.go` with more care yet.

PR #1031, opened 2025-09-15
Feat RDP - Initial implementation of RDP agent

matheusfrancisco:
## 📝 Description

We are creating an RDP proxy agent in Rust to communicate with our gateway via WebSocket

## How to run
Enable the rdp proxy
```curl -s -H "Authorization: Bearer $(hoop config view token)" http://localhost:8009/api/serverconfig/misc -XPUT -d '{"product_analytics": "active", "postgres_server_config": {"listen_address":"0.0.0.0:15432"}, "ssh_server_config": null, "rdp_server_config": {"listen_address": "0.0.0.0:3389"}}' |jq```

After enabling the proxy RDP, you can create an RDP connection. `./scripts/create_rdp_connection.sh`

`
RDP_HOST="10.211.55.6:3389" RDP_USERNAME="realuser" RDP_PASSWORD="realpassword" ./scripts/create_rdp_connection.sh -n "connection-name"
`

after the connection allows you to create a credential for it
`./scripts/create_rdp_credentials.sh -n "connection-name"
`
This will generate an xfreerdp command you can copy and execute in your terminal.

About the make build dependencies:

We are adding the hoop_rs inside the tar, so that the rust agents will be built and added 
image

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- Add the first agent for RDP in Rust, handle sessions and connections
- Add websocket layers into the agent and the gateway
- 

### Test Configuration:
- **Browser(s)**: 
- **OS**:

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

## ✅ Checklist

## 📄 Additional Notes

---

### CODE REVIEW

File: rootfs/app/migrations/000045_rdp_server_config.down.sql [hunk 1]
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE private.serverconfig DROP COLUMN rdp_server_config;
```
matheusfrancisco: add drop enum connection type as well

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		log.Infof("Timeout waiting for RDP started response for session %s", sessionID)
+		// Clean up session on timeout
+		BrokerInstance.sessions.Delete(sessionID)
+		return nil, fmt.Errorf("timeout waiting for RDP started response")
+	}
+	//try to do the handshake here
+
+}
+
+func CreateAgent(agentId string, conn any) error {
```
matheusfrancisco: I think when we create the agent, it could be "something like, registering agent for this weboscket RDP connection" so we can be more generical, in the future when ws registre more type of agents like ssh, https proxy, ..

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		conn, _ := s.Consumer.Connection.(net.Conn)
+		if _, err := conn.Write(data); err != nil {
+			log.Infof("TCP write error: %v", err)
+			break
+		}
+	}
+
+}
+
+func CreateRDPSession(
```
matheusfrancisco: this functoin is to specific for rdp protocol, we can think in a way to create a abstraction to remove trom this layer

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// forward data from agent to tcp
+func (s *Session) ForwardToTCP(data []byte) {
+	log.Infof("Forwarded %d bytes to TCP session %s", len(data), s.ID)
+	s.dataChannel <- data
+}
+
+// this will spam data from tcp to agent wsconn
+func (s *Session) StartingForwardind(data []byte) error {
```
matheusfrancisco: This whole session file is a couple of TCP/WS/RDP protocols. Now we can think in a way to abstract like a session has the type and its type has its own instance that handles the protocol internals

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
+package broker
```
matheusfrancisco, line 1: need to talk a bit more about this "file" with @EmanuelJr  and @sandromello

agentrs/src/ws/client.rs
```diff
@@ -0,0 +1,183 @@
...
+impl WebSocket {
+    pub fn new() -> anyhow::Result<Self> {
+        let config_manager =
+            conf::ConfigHandleManager::init().context("Failed to init config manager")?;
+
+        let gateway_url = std::env::var("GATEWAY_URL")
+            .unwrap_or_else(|_| "ws://localhost:8009/api/ws".to_string());
+        let mut request = gateway_url.into_client_request().unwrap();
+
+        // Insert a custom header
```
matheusfrancisco: TODO: create some retry mechanism, for example, if the agent starts first and the gateway is not up yet

agentrs/src/ws/session.rs
```diff
@@ -0,0 +1,25 @@
+use std::sync::Arc;
+
+use futures::stream::SplitSink;
+use tokio_tungstenite::WebSocketStream;
+use uuid::Uuid;
+
+#[derive(Clone, Debug)]
```
matheusfrancisco, line 7: TODO: improve this to not be rdp specific

agentrs/src/ws/rdp_message_processor.rs
```diff
@@ -0,0 +1,285 @@
...
+                    "> Session {} started, processing connection info...",
+                    header.sid
+                );
+                self.handle_session_started(header, message).await
+            }
+            _ => {
+                info!("> Unknown message type: {}", message_type);
+                Ok(())
+            }
+        }
```
matheusfrancisco, line 142: TODO: improve this abstraction so that it is not RDP specific, so we could add more protocols easily

agentrs/src/ws/rdp_message_processor.rs
```diff
@@ -0,0 +1,285 @@
+use crate::ws::proxy::start_rdp_proxy_session;
```
matheusfrancisco, line 5: todo: abstract this into some rdp/protocol handle session

matheusfrancisco:
overall: I need to improve the webscoket layer to be more agnostic to protocols, and add some tests

gateway/transport/websocket.go
```diff
@@ -0,0 +1,89 @@
...
+	// Try to decode as header + JSON message
+	if header, headerLen, err := broker.DecodeHeader(data); err == nil && headerLen <= len(data) {
+		session := broker.GetSession(header.SID)
+
+		if session == nil {
+			log.Printf("No session found for SID: %s", header.SID)
+			return
+		}
+
+		jsonData := data[headerLen:]
```
EmanuelJr: I recommend isolating this packet parse in another function, like `DecodePacket(data []byte) (sessionId string, data byte[], err error)` doing all the checks about length, etc.

Moreover, `jsonData` and `rdpData` are the same thing.

agentrs/src/ws/client.rs
```diff
@@ -0,0 +1,183 @@
...
+                        "> Connection failed (attempt {}/{}): {}",
+                        attempts, self.max_reconnection_attempts, e
+                    );
+
+                    tokio::time::sleep(self.reconnect_interval).await;
+                    continue;
+                }
+            }
+        }
+    }
```
sandromello: The current agent implementation keeps trying to establish a connection until a connection is established with the gateway. It will only hang if there's an issue parsing the HOOP_KEY. It uses a backoff mechanism to avoid sending overloading the server

Not a blocker though

agentrs/src/ws/proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+                "> Forwarding {} bytes from RDP proxy to WebSocket",
+                data.len()
+            );
+
+            // Frame the RDP response data with a Header
+            let header = Header {
+                sid: session_id,
+                len: data.len() as u32,
+            };
+            let mut framed_data = Vec::with_capacity(20 + data.len());
```
sandromello: Don't leave magic numbers without documentation, add a constant for the number `20`.

agentrs/src/ws/client.rs
matheusfrancisco: Ok, so I will do the same here.

agentrs/src/ws/proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+        .server_stream(server_stream)
+        .config(config)
+        .creds(session_info.proxy_user.clone())
+        .username(session_info.username.clone())
+        .password(session_info.password.clone())
+        .client_address(
+            session_info
+                .client_address
+                .parse()
+                .unwrap_or_else(|_| "127.0.0.1:0".parse().unwrap()),
```
sandromello, line 111: Should we fail in case it failed to parse the client address? Passing `127.0.0.1:0` as default will be more difficult to troubleshoot for such cases.

agentrs/src/conf.rs
```diff
@@ -0,0 +1,439 @@
...
+
+fn get_default_path() -> Utf8PathBuf {
+    let path = get_data_dir();
+    path.join("gateway.json")
+}
+
+fn get_path() -> Utf8PathBuf {
+    std::env::var("HOOP_PATH")
+        .map(Utf8PathBuf::from)
+        .unwrap_or_else(|_| get_default_path())
```
sandromello: You could always assume the hoop path for the agent is:
- `$HOME/.hoop`

The GO agent assumes this path exists, if it doesn't, it creates this folder. You could use the same logic for the rust agent. 

> There's also a configuration toml file that you could reuse as well (`$HOME/.hoop/config.toml`), but there's no problem if you want to use a different configuration file.

agentrs/src/conf.rs
```diff
@@ -0,0 +1,439 @@
...
+                error!("Failed to generate certificate: {e}");
+                None
+            }
+        };
+        if tls.is_none() {
+            Err(anyhow::anyhow!("TLS error generating certificates."))?;
+        }
+
+        let conf = Conf {
+            hostname: "gateway.hoop".to_string(),
```
sandromello, line 58: This value is hard-coded

agentrs/src/conf.rs
```diff
@@ -0,0 +1,439 @@
...
+
+fn get_data_dir() -> Utf8PathBuf {
+    std::env::var("HOOP_DATA_DIR")
+        .map(Utf8PathBuf::from)
+        .unwrap_or_else(|_| {
+            default_data_dir()
+                .context("couldn't determine data directory")
+                .unwrap()
+        })
+}
```
sandromello, line 248: You could assume any configuration is derived from `$HOME/.hoop/`, in case this directory doesn't exists you could create it with any defaults that you need to run it

gateway/api/apiroutes/auth.go
```diff
@@ -182,6 +183,42 @@ func (r *Router) validateTokenWithUserInfo(tokenVerifier idp.UserInfoTokenVerifi
 	return tokenVerifier.VerifyAccessTokenWithUserInfo(accessToken)
 }
 
+// validate hoop key from coonectoin websocket
+func (r *Router) VerifyWebsocketToken(c *gin.Context) {
```
sandromello: I think you could move this logic to transport and handle the authentication in a unexported function, it will make more simpler and easier to read. If we plan to add more websocket routes it will make sense to create a middleware.

gateway/broker/headers.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Decode decodes a header from bytes
+func DecodeHeader(data []byte) (*Header, int, error) {
+	if len(data) < 20 {
+		return nil, 0, fmt.Errorf("insufficient data for header")
+	}
+
+	var sid uuid.UUID
+	copy(sid[:], data[:16])
```
sandromello: You could use `uuid.ParseBytes(data[:16])`. It will ensure it's a valid uuid as well

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+	case "tcp":
+		if tcpConn, ok := s.Agent.Connection.(net.Conn); ok {
+			_, err := tcpConn.Write(data)
+			return err
+		}
+	default:
+		return nil
+	}
+	return nil
```
sandromello, line 50: If it's not a known type, you should return an error indicating that.

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// this will forward data from agent to tcp
+func (s *Session) SendAgentToTCP() {
+	for data := range s.dataChannel {
+		log.Infof("Agent -> TCP: %d bytes for session %s", len(data), s.ID)
+
+		conn, _ := s.Consumer.Connection.(net.Conn)
+		if _, err := conn.Write(data); err != nil {
```
sandromello: It will panic if you're not be able to cast `conn` .

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+
+func CreateRDPSession(
+	connTcp *Connection,
+	connectionInfo models.Connection,
+	proxyuser string,
+	clientAddr string) (*Session, error) {
+
+	sessionID := uuid.New()
+	fmt.Println("Creating session with ID:", sessionID)
+	fmt.Println("Connection Info:", connectionInfo.AgentName)
```
sandromello: Don't use fmt, use our logging library (common package)

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	log.Infof("Sent handshake to agent for session %s: %d bytes", sessionID, len(handshakeData))
+
+	// Wait for RDP started response
+	select {
+	case <-credentialsReceived:
+		log.Infof("Received RDP started response for session %s", sessionID)
+		return session, nil
+	case <-time.After(5 * time.Second):
```
sandromello: Don't use this, it could cause memory leaks.
Use this instead:
```go
timeoutCtx, cancelFn := context.WithTimeout(context.Background(), time.Second*5)
(...)
case <-timeoutCtx.Done():
```

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+	if sess, ok := BrokerInstance.sessions.Load(sessionId); ok {
+		if session, valid := sess.(*Session); valid {
+			return session
+		}
+	}
+	return nil
+}
+
+func GetSessions() map[uuid.UUID]*Session {
+	sessions := make(map[uuid.UUID]*Session)
```
sandromello:
```go
sessions := map[uuid.UUID]*Session{}
```

gateway/rdp/parser.go
```diff
@@ -0,0 +1,197 @@
+package rdp
```
sandromello, line 1: I think this part is important to have some unit tests

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	secretKeyHash, err := keys.Hash256Key(extractedCreds)
+	log.Printf("Extracted credentials: %s", extractedCreds)
+	dba, err := models.GetValidConnectionCredentialsBySecretKey(
+		pb.ConnectionTypeRDP.String(),
+		secretKeyHash)
+	if err != nil {
+		if err == models.ErrNotFound {
+			log.Errorf("invalid secret access key credentials")
```
sandromello: The user is passing invalid credentials, don't use error here, just a infof or warnf

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+			return
+		}
+		log.Errorf("failed obtaining secret access key, reason=%v", err)
+		return
+	}
+	fmt.Println("obtained db access by id, id=", dba.ID, ", subject=", dba.UserSubject, ", connection=", dba.ConnectionName)
```
sandromello: Don't use fmt, use log

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+
+			go proxy.handleRDPClient(conn, conn.RemoteAddr())
+		}
+	}()
+
+	return proxy, nil
+}
+
+func (r *RDPProxy) handleRDPClient(conn net.Conn, peerAddr net.Addr) {
```
sandromello, line 135: It's important to respond to the RDP protocol that an error has occurred, it will be more easy to diagnose and troubleshoot issues from clients. Is it possible to return a fatal error to the RDP protocol to inform the user of the problem?

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+	userCtx, err := models.GetUserContext(dba.UserSubject)
+	if err != nil {
+		log.Errorf("failed fetching user context, reason=%v", err)
+		return
+	}
+	connectionModel, err := models.GetConnectionByNameOrID(userCtx, dba.ConnectionName)
+	if err != nil {
+		log.Errorf("failed fetching connection by name or id, reason=%v", err)
+		return
+	}
```
sandromello: There's no need to fetch the user context, just create a new organization context:
```go
conn, err := models.GetConnectionByNameOrID(
  storagev2.NewOrganizationContext(dba.OrgID), 
  dba.ConnectionName)
```

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+	BrokerInstance.sessions.Store(sessionID, session)
+
+	// Decode base64 env variables
+	secrets := make(map[string]string)
+	for k, v := range connectionInfo.Envs {
+		value, _ := base64.StdEncoding.DecodeString(v)
+		secrets[k] = string(value)
+
+	}
+	fmt.Println("Decoded secrets:", secrets)
```
sandromello: Don't print Secrets!

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		"client_address": clientAddr,
+		"username":       secrets["envvar:USER"],
+		"password":       secrets["envvar:PASS"],
+		"target_address": secrets["envvar:HOST"],
+		"proxy_user":     proxyuser,
+		"message_type":   "session_started",
+		"protocol":       "rdp",
+	}
+
+	fmt.Println("Handshake Info:", handshakeInfo)
```
sandromello: Don't print secrets

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (r *RDPProxy) handleRDPClient(conn net.Conn, peerAddr net.Addr) {
+	defer conn.Close()
+
+	log.Printf("TCP connection from %s", peerAddr)
+
+	// Generate session ID
+	connection := &broker.Connection{
+		ID:         "connection-id",
```
sandromello: It's hard-coded

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Extract credentials
+	extractedCreds, err := ExtractCredentialsFromRDP(firstRDPData)
+	if err != nil {
+		log.Printf("Failed to extract credentials: %v", err)
+		return
+	}
+
+	secretKeyHash, err := keys.Hash256Key(extractedCreds)
+	log.Printf("Extracted credentials: %s", extractedCreds)
```
sandromello: Do not print the credentials

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+	handshakeData, err := json.Marshal(handshakeInfo)
+
+	if err != nil {
+		log.Infof("Failed to marshal handshake info: %v", err)
```
sandromello: Just log in the caller function and pass the context of the error instead.

```go
return nil, fmt.Errorf("failed to marshal handshake infom reason=%v", err)
```

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+
+	header := &Header{
+		SID: sessionID,
+		Len: uint32(len(handshakeData)),
+	}
+
+	framedData := append(header.Encode(), handshakeData...)
+
+	if err := session.SendToAgent(framedData); err != nil {
+		log.Infof("Failed to send handshake to agent: %v", err)
```
sandromello: Just log in the caller function and pass the context of the error instead.

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		return nil, err
+	}
+
+	log.Infof("Sent handshake to agent for session %s: %d bytes", sessionID, len(handshakeData))
+
+	// Wait for RDP started response
+	select {
+	case <-credentialsReceived:
+		log.Infof("Received RDP started response for session %s", sessionID)
```
sandromello: Log in the caller function instead

gateway/rdp/rdp.go
```diff
@@ -0,0 +1,177 @@
...
+	store.Set(instanceKey, server)
+	return nil
+}
+
+func (r *RDPProxy) Stop() error {
+	if server, ok := store.Pop(instanceKey).(*RDPProxy); ok {
+
+		for _, session := range broker.GetSessions() {
+			if session != nil {
+				log.Infof("closing session %v", session.ID)
```
sandromello: Don't need to log this for every session close

gateway/transport/websocket.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Handle incoming messages
+	client, found := broker.GetAgent(agentName)
+	if !found || client == nil {
+		log.Printf("Agent not found or nil")
+		return
+	}
+
+	for {
+		messageType, data, err := client.Connection.(*websocket.Conn).ReadMessage()
```
sandromello: This will panic. Make sure to proper cast this

agentrs/src/session.rs
```diff
@@ -0,0 +1,58 @@
...
+// a 20 bytes header with a UUID and a length field,
+// followed by the payload of the specified length.
+#[derive(Debug, Clone, Copy)]
+pub struct Header {
+    pub sid: Uuid,
+    pub len: u32,
+}
+
+const UUID_LEN: usize = 16;
+const LEN_LEN: usize = size_of::<u32>();
```
EmanuelJr: Maybe `DATA_SIZE_LEN` makes it a bit clearer.

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+	}
+	BrokerInstance.sessions.Delete(s.ID)
+	log.Infof("Session %s closed", s.ID)
+}
+
+// forward data from agent to tcp
+func (s *Session) ForwardToTCP(data []byte) {
+	log.Infof("Forwarded %d bytes to TCP session %s", len(data), s.ID)
+	s.dataChannel <- data
```
sandromello: Warning: This **may** block the main thread of the websocket handler in case it takes too long to write when reading it.

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+type Session struct {
+	ID uuid.UUID
+	//change this consumer for something else
+	Consumer            *Connection
+	Agent               *Connection
+	Connection          models.Connection
+	clientAddr          string
+	dataChannel         chan []byte
+	credentialsReceived chan bool
+}
```
sandromello: The Connection is already a core concept of the product. Make sure to use more explicit names. It's very confuse to read the code and understand which kind of type is being referenced.

Examples:

- Connection: Core Resource of Hoop
- TCPConnection: A tcp connection
- AgentTcpConnection: A tcp connection from the agent (if it's the case)
- RDPConnection: A rdp connection

gateway/broker/session.go
```diff
@@ -0,0 +1,304 @@
...
+		clientAddr:          clientAddr,
+		dataChannel:         dataChannel,
+		credentialsReceived: credentialsReceived,
+	}
+
+	// Store session immediately so it can be found by WebSocket handler
+	BrokerInstance.sessions.Store(sessionID, session)
+
+	// Decode base64 env variables
+	secrets := make(map[string]string)
```
sandromello:
```go
secrets := map[string]string{}
```

gateway/transport/websocket.go
```diff
@@ -0,0 +1,89 @@
...
+	log.Println("WebSocket connection closed")
+
+}
+
+// handleWebSocketMessage handles incoming WebSocket messages
+func handleWebSocketMessage(data []byte) {
+	log.Printf("Received %d bytes from agent", len(data))
+
+	// Try to decode as header + JSON message
+	if header, headerLen, err := broker.DecodeHeader(data); err == nil && headerLen <= len(data) {
```
sandromello: I recommend at least using a generic type to allow other kind of messages to implement other protocols. Something similar to our protobuf specification. Example:

```go
type WebSocketMessage struct {
  // the type of message indicate how to parse it and who should handle it
  Type string `json:"type"`
  // contains any arbitrary information (session id, credentials, etc)
  Metadata map[string]string`json:"metadata"`
  // contains encoded base64 binary data, it could contain another JSON or any format
  Payload []byte `json:"payload"`
}

(...)
wsMessage, err := broker.DecodeMessage(data)
if err != nil {
  panic("fail to decode message")
}

switch wsMessage.Type {
case "RdpStartSession":
case ...
}
```

Using JSON may not be ideal, but we could think of other kind of serializers such as msgpack to avoid having to encode and decode the payload via base64.

sandromello (changes requested):

gateway/rdp/parser.go
matheusfrancisco: TODO 
- [x] implement unit test

gateway/rdp/rdp.go
matheusfrancisco: I will check this

agentrs/Cargo.toml
```diff
@@ -0,0 +1,75 @@
+[package]
```
EmanuelJr, line 1: Clean the unused dependencies.

agentrs/src/session.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    pub fn decode(buf: &[u8]) -> Option<(Header, usize)> {
+        if buf.len() < HEADER_LEN {
+            //#TODO should we return an error here
+            return None;
+        }
+
+        let uuid_bytes = match buf.get(..UUID_LEN) {
+            Some(b) => Some(b),
```
EmanuelJr: `buf.get(..UUID_LEN)` already returns an `Option`, and you can just `.unwrap()`

agentrs/src/proxy/proxy.rs
```diff
@@ -0,0 +1,109 @@
...
+                Either::Left((res, _)) => res.map(|_| ()),
+                Either::Right(_) => Ok(()),
+            }
+        } else {
+            let forward_fut = tokio::io::copy_bidirectional(&mut transport_a, &mut transport_b);
+            match futures::future::select(pin!(forward_fut), pin!(kill_notified)).await {
+                Either::Left((res, _)) => res.map(|_| ()),
+                Either::Right(_) => Ok(()),
+            }
+        };
```
EmanuelJr: Remove this `Notification` since it's not been used.
Change the if to something like:
```rust
        let res = if let Some(buffer_size) = self.buffer_size {
            copy_bidirectional(&mut transport_a, &mut transport_b, buffer_size, buffer_size).await
        } else {
            tokio::io::copy_bidirectional(&mut transport_a, &mut transport_b).await
        };
```

agentrs/src/proxy.rs
```diff
@@ -0,0 +1,109 @@
...
+
+                let error = anyhow::Error::new(error);
+
+                if really_an_error {
+                    Err(error.context("forward"))
+                } else {
+                    Ok(())
+                }
+            }
+        }
```
EmanuelJr, line 36: Just some ideas:
```rust
match res {
    Ok(()) => Ok(()),
    Err(error) if is_error(&error) => Err(anyhow::Error::new(error).context("forward")),
    Err(_) => Ok(()),
}
```
or:
```rust
if let Err(error) = res {
    if is_error(&error) {
        Err(anyhow::Error::new(error).context("forward"))
    } else {
        Ok(())
    }
} else {
    Ok(())
}
```

agentrs/src/rdp_proxy.rs
EmanuelJr, line 1: Maybe move this to the `proxy` folder, renaming it to `rdp.rs`

agentrs/src/transport/transport.rs
EmanuelJr, line 1: Try to update the `tokio` version and use the `copy_bidirectional_with_sizes`

agentrs/Cargo.toml
EmanuelJr, line 1: - Remove unused dependencies.
- Try to upgrade dependencies (especially tokio`).
- Normalize the patch version, remove patch versions where they are not needed, or add patch versions in all dependencies.

agentrs/src/tls.rs
EmanuelJr, line 1: Some cleanup is needed, like `DEFAULT_CIPHER_SUITES`, `sanity`

agentrs/src/run.rs
```diff
@@ -0,0 +1,41 @@
...
+    use tokio::signal::unix::{SignalKind, signal};
+
+    let mut terminate_signal =
+        signal(SignalKind::terminate()).context("failed to create terminate signal stream")?;
+    let mut quit_signal =
+        signal(SignalKind::quit()).context("failed to create quit signal stream failed")?;
+    let mut interrupt_signal = signal(SignalKind::interrupt())
+        .context("failed to create interrupt signal stream failed")?;
+
+    futures::future::select_all(vec![
```
EmanuelJr: I would recommend using `tokio::select!`

agentrs/src/run.rs
```diff
@@ -0,0 +1,41 @@
...
+    info!("Starting Agent Service...");
+
+    let mut s = Service::new();
+    s.start()?;
+
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_io()
+        .build()
+        .context("failed to build the async runtime")?;
+    rt.block_on(build_signals_fut())?;
```
EmanuelJr: Not sure; you could remove this and only call the `build_signals_fut()?`

agentrs/src/session.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl Header {
+    pub fn encode(self) -> [u8; 20] {
+        let mut buf = [0u8; 20];
+        buf[..16].copy_from_slice(self.sid.as_bytes());
+        buf[16..].copy_from_slice(&self.len.to_be_bytes());
+        buf
+    }
+
+    pub fn decode(buf: &[u8]) -> Option<(Header, usize)> {
```
EmanuelJr: I guess it should be a `Result` or even `Result<Option>>`.

agentrs/src/session.rs
```diff
@@ -0,0 +1,58 @@
+use uuid::Uuid;
+
+// Binary package format:
+// [ sid(16 bytes) | len(4 bytes) | payload(len bytes) ]
+// a 20 bytes header with a UUID and a length field,
+// followed by the payload of the specified length.
+#[derive(Debug, Clone, Copy)]
+pub struct Header {
```
EmanuelJr, line 8: Add a `dataSize` field instead of returning in a tuple; it's still header information.

agentrs/src/ws/message.rs
```diff
@@ -0,0 +1,122 @@
...
+        Some((Header { sid, len }, 20))
+    }
+}
+
+// Protocol types
+pub const PROTOCOL_RDP: &str = "rdp";
+
+// Message types
+pub const MESSAGE_TYPE_SESSION_STARTED: &str = "session_started";
+pub const MESSAGE_TYPE_DATA: &str = "data";
```
EmanuelJr: Use `enum` and maybe implement the traits `ToString` and `FromStr` to make the life easier

agentrs/src/ws/message.rs
```diff
@@ -0,0 +1,122 @@
...
+    base64::engine::general_purpose::STANDARD
+        .decode(s)
+        .map_err(serde::de::Error::custom)
+}
+
+#[derive(Debug, Clone)]
+pub struct Header {
+    pub sid: Uuid,
+    pub len: u32,
+}
```
EmanuelJr: A copy from `agentrs/src/session.rs`?

agentrs/src/run.rs
```diff
@@ -0,0 +1,41 @@
+use tracing::info;
+
+use crate::listener::Service;
+use anyhow::Context;
+
+#[cfg(unix)]
+async fn build_signals_fut() -> anyhow::Result<()> {
```
EmanuelJr: Maybe rename to `wait_shutdown`

agentrs/src/session.rs
EmanuelJr, line 1: It's more about our transport protocol instead of session, which makes a bit more sense to add the encoding and decoding stuff here, since it's agnostic to the transport layer.

agentrs/src/ws/message.rs
EmanuelJr, line 1: It's also related to the `agentrs/src/session.rs` comment in the file related to protocol.

EmanuelJr (changes requested):

agentrs/src/transport/transport.rs
matheusfrancisco: okay

agentrs/src/run.rs
matheusfrancisco: i dont think I can do that I would maybe have tho change some stuffs

matheusfrancisco: 🤔

agentrs/src/session.rs
matheusfrancisco: we can do this in the next pr if you dont mind

agentrs/src/ws/message.rs
matheusfrancisco: we can do this in the next pr if you dont mind

Makefile
```diff
@@ -12,6 +12,8 @@ OS := $(shell echo "$(GOOS)" | awk '{print toupper(substr($$0, 1, 1)) tolower(su
 SYMLINK_ARCH := $(if $(filter $(GOARCH),amd64),x86_64,$(if $(filter $(GOARCH),arm64),aarch64,$(ARCH)))
 POSTREST_ARCH_SUFFIX := $(if $(filter $(GOARCH),amd64),linux-static-x64.tar.xz,$(if $(filter $(GOARCH),arm64),ubuntu-aarch64.tar.xz,$(ARCH)))
 
+RUST_TARGET := $(if $(filter $(GOOS),linux),$(if $(filter $(GOARCH),amd64),x86_64-unknown-linux-gnu,$(if $(filter $(GOARCH),arm64),aarch64-unknown-linux-gnu)),$(if $(filter $(GOOS),darwin),$(if $(filter $(GOARCH),amd64),x86_64-apple-darwin,$(if $(filter $(GOARCH),arm64),aarch64-apple-darwin))))
```
matheusfrancisco: What it does:
It maps Go's GOOS and GOARCH variables to Rust target triplets:
If GOOS=linux:
```
GOARCH=amd64 → x86_64-unknown-linux-gnu
GOARCH=arm64 → aarch64-unknown-linux-gnu
```

If GOOS=darwin (macOS):
```
GOARCH=amd64 → x86_64-apple-darwin
GOARCH=arm64 → aarch64-apple-darwin
```
Why it's needed:
Go uses simple names like linux/amd64
Rust uses more specific target triplets like x86_64-unknown-linux-gnu
This variable converts between the two naming conventions
Example:
When you run make build GOOS=linux GOARCH=amd64
RUST_TARGET becomes x86_64-unknown-linux-gnu
The Rust build command becomes: cross build --release --target x86_64-unknown-linux-gnu
This ensures the Rust binary is built for the same platform as the Go binary.

agentrs/src/ws/message.rs
```diff
@@ -89,14 +56,16 @@ impl WebSocketMessage {
         // Serialize the message to JSON
         let json_data = serde_json::to_vec(self)?;
 
+        let data_size_header = 20;
```
EmanuelJr: Make it a `const`

agentrs/src/conf.rs
```diff
@@ -0,0 +1,425 @@
...
+                };
+                Some(Tls::init(cert_source, true)?)
+            }
+            Err(e) => {
+                error!("Failed to generate certificate: {e}");
+                None
+            }
+        };
+        if tls.is_none() {
+            Err(anyhow::anyhow!("TLS error generating certificates."))?;
```
EmanuelJr: You could use `Result` instead of `Option` and at the end of the `match` just use `?`, you can use `map` and `map_err` too.

agentrs/src/conf.rs
```diff
@@ -0,0 +1,425 @@
...
+        return Ok(conf_manager);
+    }
+
+    pub fn init_from_files() -> anyhow::Result<Self> {
+        let hoop_key = get_token();
+        if hoop_key.is_none() {
+            Err(anyhow::anyhow!(
+                "HOOP_KEY environment variable is not set. This may lead to insecure configurations."
+            ))?;
+        }
```
EmanuelJr, line 87: You can use `ok_or_else` or even `.context` from `anyhow`.

agentrs/src/conf.rs
```diff
@@ -0,0 +1,425 @@
...
+                ConfFile {
+                    hostname: None,
+                    token: hoop_key,
+                    tls_certificate_source: Some(CertSource::External),
+                    tls_certificate_file: None,
+                    tls_private_key_file: None,
+                    tls_verify_strict: Some(true),
+                }
+            }
+        };
```
EmanuelJr: I guess something like this would be better:
```rust
let c = conf.ok_or_else(|| anyhow::anyhow!(
    "No config file found at {path}. Please create a configuration file or set HOOP_CERT=true to auto-generate certificates."
))?;

let conf_file = ConfFile {
    token: hoop_key,
    ..c
};
```

agentrs/src/ws/client.rs
```diff
@@ -0,0 +1,190 @@
...
+        let gateway_url = std::env::var("GATEWAY_URL")
+            .unwrap_or_else(|_| "ws://localhost:8009/api/ws".to_string());
+        let mut request = gateway_url.into_client_request().unwrap();
+
+        // Insert a custom header
+        let token = config_manager.conf.token.clone().unwrap();
+        let token: &'static str = Box::leak(token.into_boxed_str());
+        request
+            .headers_mut()
+            .insert("HOOP_KEY", HeaderValue::from_static(token));
```
EmanuelJr: Maybe something like this, avoids too much allocation on the heap.
```suggestion
        let token = config_manager.conf.token.clone().unwrap();
        request
            .headers_mut()
            .insert("HOOP_KEY", HeaderValue::from_str(token.as_str()).unwrap());
```

agentrs/src/ws/client.rs
```diff
@@ -0,0 +1,190 @@
...
+                    Ok(Ok(())) => {
+                        info!("> Message processor completed normally");
+                        // This is a graceful closure, return Ok to exit reconnection loop
+                        self.cleanup_resources(sessions, active_proxies, session_channels).await;
+                        Ok(())
+                    }
+                    Ok(Err(e)) => {
+                        error!("> Message processor error: {}", e);
+                        // This is a connection error, return it to trigger reconnection
+                        self.cleanup_resources(sessions, active_proxies, session_channels).await;
```
EmanuelJr: You can move all those `cleanup_resources` after the `select!`

agentrs/src/ws/client.rs
```diff
@@ -0,0 +1,190 @@
...
+        let processor_task =
+            tokio::spawn(async move { message_processor.process_messages(ws_receiver).await });
+
+        // Start heartbeat task in case connection stucked or deadlock
+        let heartbeat_task = self.spawn_heartbeat_task(ws_sender.clone());
+
+        let result = tokio::select! {
+            result = processor_task => {
+                match result {
+                    Ok(Ok(())) => {
```
EmanuelJr, line 118: Just an idea, maybe I would make a `match` inside the `Ok` case

agentrs/src/ws/stream.rs
```diff
@@ -0,0 +1,125 @@
...
+            Ok(data) => {
+                debug!(
+                    "> ChannelWebSocketStream: Received {} bytes from channel",
+                    data.len()
+                );
+                let to_read = std::cmp::min(buf.remaining(), data.len());
+                buf.put_slice(&data[..to_read]);
+                if data.len() > to_read {
+                    this.read_buffer.extend_from_slice(&data[to_read..]);
+                }
```
EmanuelJr: I would check before the first copy.
```suggestion
                if data.len() > to_read {
                    this.read_buffer.extend_from_slice(&data);
                } else {
                    buf.put_slice(&data[..to_read]);
                }
```

Makefile
```diff
@@ -3,6 +3,9 @@ VERSION ?= $(or ${GIT_TAG},${GIT_TAG},v0)
 GITCOMMIT ?= $(shell git rev-parse HEAD)
 DIST_FOLDER ?= ./dist
 
+# Sanitize VERSION by replacing forward slashes with underscores
+SANITIZED_VERSION := $(shell echo "${VERSION}" | sed 's|/|_|g')
```
matheusfrancisco: added this because my branch had / and was considering folder

agentrs/src/ws/stream.rs
matheusfrancisco: will do this later

agentrs/src/ws/client.rs
matheusfrancisco: same i will do this later

.github/workflows/release.yml
```diff
@@ -195,6 +195,16 @@ jobs:
           go-version: ">=1.23.8"
           cache-dependency-path: "**/go.sum"
 
+      - name: Setup Rust
+        uses: actions-rs/toolchain@v1
```
EmanuelJr: This action is archived.
I've seen some repositories using [this](https://github.com/dtolnay/rust-toolchain) one.

.gitignore
```diff
@@ -4,6 +4,7 @@
 *.dll
 *.so
 *.dylib
+.bacon-locations
```
EmanuelJr: Move to the `# IDE related` area.

agentrs/src/listener.rs
```diff
@@ -0,0 +1,120 @@
...
+                    Err(error) => error!("Something went very wrong with a task {:?}", error),
+                }
+
+                if rest.is_empty() {
+                    break;
+                } else {
+                    join_all = futures::future::select_all(rest);
+                }
+            }
+        });
```
EmanuelJr: Try something like this:
```suggestion
        let mut join_set = JoinSet::new();
        for child in tasks.inner {
            join_set.spawn(child.join());
        }

        runtime.spawn(async move {
            while let Some(result) = join_set.join_next().await {
                match result {
                    Ok(Ok(_)) => {
                        info!("A task terminated gracefully");
                    }
                    Ok(Err(error)) => {
                        error!("A task failed {:?}", error);
                    }
                    Err(error) => {
                        error!("Something went very wrong with a task {:?}", error);
                    }
                }
            }
        });
```

agentrs/src/listener.rs
```diff
@@ -0,0 +1,120 @@
...
+                                }
+                            }
+                        }
+                    }
+                });
+
+                // Wait for 1 more second before forcefully shutting down the runtime
+                runtime.shutdown_timeout(Duration::from_secs(1));
+
+                self.state = State::Stopped;
```
EmanuelJr: I guess we don't need it.

agentrs/src/listener.rs
```diff
@@ -0,0 +1,120 @@
...
+
+                                if count >= MAX_COUNT {
+                                    error!("Some tasks are not terminating, forcing shutdown");
+                                    break;
+                                } else {
+                                    error!("Waiting for tasks to terminate... (attempt {}/{})", count, MAX_COUNT);
+                                }
+                            }
+                        }
+                    }
```
EmanuelJr: Just an idea...
```rust
for attempt in 1..=MAX_COUNT {
    tokio::select! {
        _ = shutdown_handle.all_closed() => {
            info!("All tasks have terminated gracefully");
            return; // or break ¯\_(ツ)_/¯
        }
        _ = tokio::time::sleep(Duration::from_secs(10)) => {
            if attempt == MAX_COUNT {
                error!("Some tasks did not terminate, forcing shutdown");
            } else {
                warn!("Waiting for tasks to terminate... (attempt {}/{})", attempt, MAX_COUNT);
            }
        }
    }
}
```

agentrs/src/logio.rs
EmanuelJr, line 1: Just remove this file.

EmanuelJr (changes requested):
I'll be back

agentrs/src/listener.rs
matheusfrancisco: Adding join set will need to change the shutdown 
i think will need to coordinate the shutdown with the join set
everything inside a for loop 
```
for .. max_count {
select {
   maybe = join_set.join_next().. 
     your match...

   all_closed
   time to tick // sleep 
}
}
```

matheusfrancisco: JoinSet is not Sync; it should be owned and polled by a single task—the one coordinating shutdown.

matheusfrancisco: i might wrong, could be I could change it easily

agentrs/src/ws/client.rs
EmanuelJr: It should be really fast. You can ask the copilot or something to do.

rootfs/app/migrations/000045_rdp_server_config.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE private.serverconfig ADD COLUMN rdp_server_config JSONB NULL;
+ALTER TYPE private.enum_connection_type ADD VALUE IF NOT EXISTS 'rdp';
```
EmanuelJr: No need, since the `subtype` that is `rdp`.

common/proto/types.go
```diff
@@ -182,6 +182,8 @@ func ToConnectionType(connectionType, subtype string) ConnectionType {
 			return ConnectionType(ConnectionTypeDynamoDB)
 		case "cloudwatch":
 			return ConnectionType(ConnectionTypeCloudWatch)
+		case "rdp":
```
EmanuelJr: It should be in the `application` instead of `custom`.

scripts/create_rdp_connection.sh
```diff
@@ -0,0 +1,132 @@
...
+echo -e "${YELLOW}Host: $RDP_HOST${NC}"
+echo -e "${YELLOW}Username: $RDP_USERNAME${NC}"
+
+RESPONSE=$(curl -s -w "\n%{http_code}" \
+    -X POST \
+    -H "Content-Type: application/json" \
+    -H "Authorization: Bearer $TOKEN" \
+    -d "{
+        \"name\": \"$CONNECTION_NAME\",
+        \"type\": \"custom\",
```
EmanuelJr: `application` instead of `custom`.

gateway/broker/communicators.go
EmanuelJr, line 1: Try to separate (in the same file) the `AgentCommunicator` and `ClientCommunicator` contexts.
Like:
```go
type AgentCommunicator struct{ conn *websocket.Conn }
func NewAgentCommunicator(conn net.Conn)
func (a *AgentCommunicator) Send
...

type ClientCommunicator struct{ conn net.Conn }
func NewClientCommunicator(conn net.Conn)
...
```
Also, no need to return the `interface` instead of the struct, I guess.
The `Read` function could follow the same idea of the `io.Reader` interface.

gateway/broker/communicators.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type ConnectionCommunicator interface {
+	Send(data []byte) error
+	Read() (int, []byte, error)
+	Close()
+}
+
+type AgentCommunicator struct{ conn *websocket.Conn }
+type ClientCommunicator struct{ conn net.Conn }
```
racerxdl: Since you're returning the interface, following the go patterns you would make these structs private.
Usually the interface is the public part and the implementation is hidden.

gateway/broker/communicators.go
EmanuelJr: It also would be possible to have after that:
(just an idea)
```go
type ConnectionCommunicator interface {
	io.Writer
	io.Reader
	io.Closer
}
```

gateway/broker/communicators.go
matheusfrancisco: great catch

gateway/broker/communicators.go
matheusfrancisco: 👍🏽

agentrs/examples/generate_cert.rs
```diff
@@ -0,0 +1,49 @@
...
+    println!("  Private Key: {}", key_path);
+
+    // Example 4: Convert to rustls types
+    let (cert_der, key_der) = cert_key_pair.to_rustls();
+    println!("\nCertificate converted to rustls-compatible types");
+    println!("  Certificate DER length: {} bytes", cert_der.len());
+    println!("  Private Key type: {:?}", std::mem::discriminant(&key_der));
+
+    Ok(())
+}
```
racerxdl, line 49: I got the feeling that these `examples` folder are sort "left here" (I see that the go version does not have it).

The generate_cert, might be a good functionality to have inside the agent itself if its needed to generate a self-signed certificate to make it work (imagine that if user runs with docker/manual its easier if its a single binary that can provision the "default" config).

agentrs/src/tls.rs
```diff
@@ -0,0 +1,263 @@
...
+// > Making one of these can be expensive, and should be once per process rather than once per connection.
+//
+// source: https://docs.rs/rustls/0.21.1/rustls/client/struct.ClientConfig.html
+//
+// We’ll reuse the same TLS client config for all proxy-based TLS connections.
+// (TlsConnector is just a wrapper around the config providing the `connect` method.)
+static TLS_CONNECTOR: LazyLock<tokio_rustls::TlsConnector> = LazyLock::new(|| {
+    let mut config = rustls::client::ClientConfig::builder()
+        .dangerous()
+        .with_custom_certificate_verifier(Arc::new(danger::NoCertificateVerification))
```
racerxdl, line 20: Should this be really hardcoded? That should probably be a config arg (PCI compliance envs require full check of certificates).

agentrs/examples/generate_cert.rs
matheusfrancisco: It is already being generated as a self-signed certificate; it will not be generated if the user provides one in the config.toml.

agentrs/src/tls.rs
matheusfrancisco: I don't know why I had trouble with this; I don't remember. We're not validating because the client is accepting the self-signed certificate we're generating, and we're acting like a man-in-the-middle after the TLS upgrade between the client and RDP server. I had a few issues, but I also know that wasn't your question, and we can add it in the environment config. 


```rust
let verifier = if config.verify_certificates {
    Arc::new(rustls::client::WebPkiVerifier::new(
        rustls::RootCertStore::empty(), 
        None
    ))
} else {
    Arc::new(danger::NoCertificateVerification)
};
```
I will be back in this area when we start testing with different configurations on the RDP server, if you don't mind

gateway/broker/headers.go
EmanuelJr, line 1: Those `20` put in a `const` like `HEADER_LENGTH` pls

gateway/broker/headers.go
```diff
@@ -0,0 +1,112 @@
...
+	Payload  []byte            `json:"payload"`
+}
+
+// Message types
+const (
+	MessageTypeSessionStarted = "session_started"
+	MessageTypeData           = "data"
+)
+
+const HEADER_SIZE = 20 // 16 bytes for UUID + 4 bytes for length
```
EmanuelJr: Put `headerSize` (or `HeaderSize` if it'll be used outside this package) just because of the GoLang standard 🫠

matheusfrancisco: 🤷🏽

gateway/broker/session.go
```diff
@@ -0,0 +1,199 @@
...
+				log.Infof("Failed to send RDP data to agent: %v", err)
+				break
+			}
+		}
+	}
+	return nil
+}
+
+// this will forward data from agent to tcp
+func (s *Session) SendAgentToTCP() {
```
EmanuelJr: `ForwardToClient` makes more sense.

gateway/broker/session.go
```diff
@@ -0,0 +1,199 @@
...
+	case s.dataChannel <- data:
+		// Successfully sent
+	case <-s.ctx.Done():
+		// Session is being closed, don't send
+		log.Infof("Session %s closed, dropping data", s.ID)
+	}
+}
+
+// this will spam data from tcp to agent wsconn
+func (s *Session) StartingForwardind(data []byte) error {
```
EmanuelJr: Also change to `ForwardToAgent` (or something)

EmanuelJr (changes requested):

agentrs/src/tls.rs
racerxdl: It makes sense for normal environments: usually the certs are self-signed even on windows and you need to accept them. Some enterprise envs use ones that has the CA installed in the OS. And its fine to have it default to ignore validation (ideally, we would have a way to pin them, like passing the hash or something like). 
I just think it might be a good idea to have an configuration to not ignore validations.

rootfs/app/migrations/000045_rdp_server_config.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE private.serverconfig ADD COLUMN rdp_server_config JSONB NULL;
```
EmanuelJr, line 3: Set a default configuration.

scripts/dev/run.sh
```diff
@@ -56,8 +67,10 @@ docker stop hoopdev &> /dev/null || true
 docker rm hoopdev &> /dev/null || true
 
 docker run --rm --name hoopdev \
+  --network host \
```
racerxdl, line 70: If it runs on host network, ports doesnt need to be forwared :thinking:

matheusfrancisco: yep, i forgot

rootfs/app/migrations/000045_rdp_server_config.down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE private.serverconfig ADD COLUMN rdp_server_config JSONB DEFAULT '{"listen_address":"0.0.0.0:3389"}';
```
EmanuelJr, line 3: I would recommend a port like `11320` or something like this, some unused port.

gateway/broker/session.go
```diff
@@ -0,0 +1,199 @@
...
+	if s.AgentCommunicator != nil {
+		s.AgentCommunicator.Close()
+	}
+
+	// Remove from sessions map
+	BrokerInstance.sessions.Delete(s.ID)
+}
+
+// forward data from agent to tcp
+func (s *Session) ForwardToTCP(data []byte) {
```
EmanuelJr, line 90: Well, there is the function `SendToClient` already; it does not make much sense, maybe.

gateway/broker/headers.go
```diff
@@ -0,0 +1,112 @@
...
+	// Combine header + JSON data
+	result := make([]byte, HeaderSize+len(jsonData))
+	copy(result[:HeaderSize], header.Encode())
+	copy(result[HeaderSize:], jsonData)
+
+	return result, nil
+}
+
+// DecodeWebSocketMessage decodes bytes to WebSocketMessage
+func DecodeWebSocketMessage(data []byte) (uuid.UUID, *WebSocketMessage, error) {
```
EmanuelJr, line 94: Maybe... `DecodeJsonMessage`, and make another to `DecodeBinaryMessage`

EmanuelJr (dismissed):

sandromello (approved):

PR #1166, opened 2025-11-28
fix: Fix group confusion on multiple-group approve list

racerxdl:
## 📝 Description

Previously when there were multiple groups that could approve a request, if a user that belongs to only one hit the button of the second one, the slack bot would answer as approved even though it was not approved.

Example:

1. Lucas is from group A
2. Resource X requires approve from either A or B
3. A user requests the resource and see the slack message
4. User clicks on message of B group
5. Slack confirms that it was approved, but it was actually not.

I added a check now, that will return a message to the user telling it is not in the approvers group and keep the approve button there.

## 🔗 Related Issue

Fixes SUP-14

## 🚀 Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- Added a check if the collected groups contains the group that came from the message.
- Added a check to not send a user feedback in slack in case status is still PENDING.

## 🧪 Testing

1. Create a resource that can be used upon approval by two groups
2. Be a user in one of the groups only
3. Configure slack integration
4. Make a query to that resource so it will trigger the slack approval message
5. Try to click the `Approve` button of the group you dont belong
6. It should show a private message "You do not belong to group XXXXX"

Before:
image

After:
image

### DIFF
--- a/gateway/transport/plugins/slack/events.go
+++ b/gateway/transport/plugins/slack/events.go
@@ -1,6 +1,9 @@
 package slack
 
 import (
+	"fmt"
+	"slices"
+
 	"github.com/aws/smithy-go/ptr"
 	"github.com/hoophq/hoop/common/log"
 	reviewapi "github.com/hoophq/hoop/gateway/api/review"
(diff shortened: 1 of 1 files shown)


EmanuelJr (approved):

matheusfrancisco (approved):

gateway/transport/plugins/slack/events.go
```diff
@@ -84,7 +94,12 @@ func (p *slackPlugin) performReview(ev *event, ctx *storagev2.Context, status mo
 		msg = "You're not eligible to approve/reject this review"
 	case nil:
 		isApproved := rev.Status == models.ReviewStatusApproved
-		err = ev.ss.UpdateMessage(ev.msg, isApproved)
+		isStillPending := rev.Status == models.ReviewStatusPending
+		if isStillPending {
+			err = fmt.Errorf("user cannot approve/reject this review")
```
sandromello: The user was able to approve the review resource, however the resource is still pending and need more users to approve it. Change the error message because it will confuse us.

A more aligned message would be:
"user was able to approve it, but the resource is still pending"

racerxdl: Done ^^

PR #1176, opened 2025-12-04
Made agent controller process packets async

racerxdl:
## 📝 Description

This makes processing packets async. This is safe because we're already using a mutexClient so concurrent calls to `.Send` are properly sync.

I also notice this improves latency on DBeaver for postgres, even in a single connection (it seens dbeaver makes several ops in parallel, which was hanging before).

## 🔗 Related Issue

Partially fixes ENG-103

## 🚀 Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- Agent Controller will now receive a packet, and spawn a go routine to process it.

## 🧪 Testing

There are several ways to test that. The ways I tested it:

- Have one SSH and one Postgres connection:
1. Put a very screen updating intesive on SSH side, like `htop` or `cat /dev/urandom` or `watch -n0.1 cat /proc/cpuinfo`
2. Make a slow query in postgres (that will hang the connection) which can be a simple `select pg_sleep(100)`.

In our previous scenario, they would interfer in eachother. Now it should almost not interfere on eachother (it will still be using the same connection so there is a cpu overhead for that)

### Test Configuration:
- **Browser(s)**: 
- **OS**:

### Tests performed:
- [ ] Unit tests pass
- [ ] Integration tests pass

## ✅ Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have checked my code and corrected any misspellings

## 📄 Additional Notes

---

### DIFF
--- a/rootfs/app/migrations/000052_user_groups_name_size.up.sql
+++ b/rootfs/app/migrations/000052_user_groups_name_size.up.sql
@@ -1,1 +1,8 @@
+-- very old legacy views cleanup - versions before 1.35.22
+-- the migration will fail if those view exists
+DROP VIEW IF EXISTS public.user_groups CASCADE;
+DROP VIEW IF EXISTS public_a.user_groups CASCADE;
+DROP VIEW IF EXISTS public_b.user_groups CASCADE;
+
+-- modify column size
 ALTER TABLE private.user_groups ALTER COLUMN name TYPE VARCHAR(180);


agent/controller/agent.go
```diff
@@ -98,82 +98,94 @@ func (a *Agent) Close(cause error) {
...
+			break
+		}
 
-		case pbsystem.RunbookHookRequestType:
-			runbookhook.ProcessRequest(a.client, pkt)
+		select {
+		case <-a.shutdownCtx.Done():
+			break
+		default:
 		}
```
sandromello, line 179: This function must be processed first to avoid blocking on recv for any reason in case of a shutdown
Make sure to return the proper shutdown context as well. Break will return from the select block not from the loop

```go
return context.Cause(a.shutdownCtx)
```

agent/controller/agent.go
```diff
@@ -98,82 +98,94 @@ func (a *Agent) Close(cause error) {
...
-		// system
-		case pbsystem.ProvisionDBRolesRequest:
-			dbprovisioner.ProcessDBProvisionerRequest(a.client, pkt)
+	for {
+		pkt, err := a.client.Recv()
+		if err != nil {
+			if !errors.Is(err, io.EOF) {
+				log.Errorf("receiving channel error: %v", err)
+			}
+			break
```
sandromello: Just make sure to only return , the caller will handle of printing properly the context of the error

sandromello (changes requested):

sandromello:
✅ Build Completed with Success, Version=1176.0.0-24a1c54

PR #1196, opened 2025-12-15
Set Default Runbook for new orgs and orgs that does not have runbook

matheusfrancisco:
## 📝 Description

-  Setting a new organization, that organization should have a Runbooks repo set with our open source Runbooks repository;
- If an organization already is using hoop, but don't have Runbooks set, we should set our open source Runbooks repository
- If an organization already is using hoop and have Runbooks set, we should not set anything for this case, we leave it as it 

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [x] 🧹 Chore

## 📋 Changes Made

- add: default runbook for organization without runbook and new orgs created
- 
- 

### Test Configuration:
- **Browser(s)**: 
- **OS**:

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

## ✅ Checklist

## 📄 Additional Notes

---

### DIFF
--- a/gateway/models/runbooks.go
+++ b/gateway/models/runbooks.go
@@ -194,7 +194,7 @@ func BuildCommonConfig(config *RunbookRepositoryConfig) (*commonRunbooks.Config,
 
 func CreateDefaultRunbookConfiguration(db *gorm.DB, orgID string) (*Runbooks, error) {
 	const defaultRepoURI = "https://github.com/hoophq/demo-runbooks"
-	const defaultRepoName = "demo-runbooks"
+	const defaultRepoName = "github.com/hoophq/demo-runbooks"
 
 	// Check if runbook configuration already exists for this org
 	existing, _ := GetRunbookConfigurationByOrgID(db, orgID)


gateway/api/signup/signup.go
```diff
@@ -74,6 +74,14 @@ func Post(c *gin.Context) {
 		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed creating organization"})
 		return
 	}
+
+	_, err = models.CreateDefaultRunbookConfiguration(models.DB, org.ID)
+	if err != nil {
+		log.Errorf("failed creating default runbook configuration, err=%v", err)
+		c.JSON(http.StatusInternalServerError, gin.H{"message": "failed creating default runbook configuration"})
+		return
+	}
```
matheusfrancisco: idk if we should fail the signup if the runbook, for some reason, fails 🤔 ? @EmanuelJr WDYT?

matheusfrancisco: also @sandromello wdyt?

gateway/main.go
```diff
@@ -82,6 +82,11 @@ func Run() {
 			log.Fatal(err)
 		}
 
+		_, err = models.CreateDefaultRunbookConfiguration(models.DB, org.ID)
+		if err != nil {
+			log.Fatal(err)
+		}
+
```
matheusfrancisco, line 89: Also, if it fails creating, should I not run?  @EmanuelJr @sandromello

gateway/api/signup/signup.go
mtmr0x: agree! Even tho it's a edge case, this should be optimistic and if the Runbook wasn't created by some reason, failing to create the account shouldn't be the outcome of this

EmanuelJr: You could call this `CreateDefaultRunbookConfiguration` on the `CreateOrganization`.
There is the logic about the plugins already, so it would not be a big problem.

gateway/main.go
sandromello: I think it should be a best effort scenario

rootfs/app/migrations/000054_default_runbook.up.sql
```diff
@@ -0,0 +1,15 @@
...
+SELECT 
+    o.id,
+    '{"demo-runbooks": {"git_url": "https://github.com/hoophq/demo-runbooks", "git_branch": "main"}}'::JSONB,
+    NOW(),
+    NOW()
+FROM orgs o
+LEFT JOIN runbooks r ON r.org_id = o.id
+WHERE r.id IS NULL;
+
+COMMIT;
```
sandromello: The organization creation runs after the migration process, therefore it won't work for setups with clean database state

matheusfrancisco: I test it here,  I think this clause LEFT JOIN runbooks r ON r.org_id = o.id will get only r.org_id = o.id righit it ? @sandromello

gateway/api/signup/signup.go
matheusfrancisco: Ok I will call only in one place and not stop the account

rootfs/app/migrations/000054_default_runbook.up.sql
matheusfrancisco: i will do this on the application, removing this migration

gateway/models/runbooks.go
```diff
@@ -190,3 +191,37 @@ func BuildCommonConfig(config *RunbookRepositoryConfig) (*commonRunbooks.Config,
 
 	return commonRunbooks.NewConfigV2(configInput)
 }
+
+func CreateDefaultRunbookConfiguration(db *gorm.DB, orgID string) (*Runbooks, error) {
+	const defaultRepoURI = "https://github.com/hoophq/demo-runbooks"
+	const defaultRepoName = "demo-runbooks"
```
EmanuelJr: `repoName` must be `github.com/hoophq/demo-runbooks`, same applies for the migrations, etc.

PR #1264, opened 2026-02-03
[ENG 231] Promote HTTPProxy to a connectionType

matheusfrancisco:
## 📝 Description

HTTP proxy was promoted from a **subtype** under `application`/`custom` to a **first-class connection type** `httpproxy`. A migration (`000061_add_httpproxy_connection_type`) adds the new enum value so new connections can be stored as `type = 'httpproxy'`. Existing connections may still be stored as `type = 'application'` or `type = 'custom'` with `subtype = 'httpproxy'`.

Without these changes, existing HTTP proxy connections would break in the UI (e.g. credentials tab, setup flow, connection method inference), and new connections would not be supported everywhere.

### 1. `resource_catalog/helpers.cljs`

- **Change**: Map for `"httpproxy"` was updated from `{:type "application" :subtype "httpproxy"}` to `{:type "httpproxy" :subtype "httpproxy"}` so **new** resources use the new type.
- **Backwards compatibility**: Subtype is kept as `"httpproxy"` because it is still used elsewhere (e.g. setup/resource screen). Existing connections that remain `application`/`custom` + `httpproxy` continue to work because the rest of the frontend now handles both shapes.

### 2. `setup/events/process_form.cljs`

- **Change**: All logic that previously treated HTTP proxy only as `(application + subtype httpproxy)` was extended to **also** treat `type = "httpproxy"` the same way:
  - **Credentials**: Added `http-credentials-v1` for `connection-type "httpproxy"` and merged it with existing `http-credentials` so both old and new types contribute to `:network-credentials` and connection method inference.
  - **Env vars**: `http-env-vars` is now applied when either `(application + http proxy subtype)` **or** `type = "httpproxy"`, so REMOTE_URL/INSECURE and related env handling work for both.
  - **Connection info**: Added `http-proxy-connection-info` for `connection-type "httpproxy"` and included it in `inferred-connection-info` and in the branch that sets `:connection-env-vars` so the setup flow works for the new type as well as the old one.

This keeps **backwards compatibility**: existing connections (application/custom + httpproxy) keep their behavior, while new connections (type httpproxy) get the same behavior.

### 3. `configure_role/credentials_tab.cljs`

- **Change**: The credentials step dispatcher now has an explicit branch for `"httpproxy"` that renders `[network/http-credentials-form]`, in addition to the existing `"application"` branch that handles SSH/git/github and (via shared logic) HTTP proxy subtype.
- **Backwards compatibility**: Connections with `type = "application"` and `subtype = "httpproxy"` still use the `"application"` branch and existing behavior. Connections with `type = "httpproxy"` use the new branch and see the same HTTP proxy credentials form. So both old and new connection shapes render correctly in the credentials tab.

## Backend alignment (proto/types.go)

So that the **backend** also treats legacy rows as HTTP proxy, `ToConnectionType` was updated to map `application`/`custom` + subtype `"httpproxy"` to `ConnectionTypeHttpProxy`. That way:

- Existing DB rows with `type = 'application'` or `type = 'custom'` and `subtype = 'httpproxy'` still resolve to HttpProxy (CLI, gateway, agents).
- New rows with `type = 'httpproxy'` also resolve to HttpProxy.

Together, the **frontend** (render both shapes) and **backend** (resolve both shapes to HttpProxy) keep connection backwards compatibility while supporting the new `httpproxy` connection type.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [x] ♻️ Code refactor
- [ ] ✅ Test update
- [x] 🧹 Chore

## 📋 Changes Made

- add: httpproxy as a connectionType
- edit: frontend to keep the backward compatibility

### Test Configuration:
- **Browser(s)**: 
- **OS**:

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

## ✅ Checklist

## 📄 Additional Notes

---

### DIFF
--- a/webapp/src/webapp/connections/views/setup/events/process_form.cljs
+++ b/webapp/src/webapp/connections/views/setup/events/process_form.cljs
@@ -399,12 +399,14 @@
         network-credentials (when (and (= connection-type "application")
                                        (= connection-subtype "tcp"))
                               (extract-network-credentials credentials))
-        http-credentials (when (and (= connection-type "application")
-                                    is-http-proxy-subtype?)
-                           (extract-http-credentials credentials))
+        ;; Deprecated: we are moving to httpproxy being its own connection type, not a subtype
+        ;; of application. This is needed to render older forms saved in that format.
+        http-credentials-deprecated (when (and (= connection-type "application")
+                                               is-http-proxy-subtype?)
+                                      (extract-http-credentials credentials))
 
-        http-credentials-v1 (when (= connection-type "httpproxy")
-                              (extract-http-credentials credentials))
+        http-credentials (when (= connection-type "httpproxy")
+                           (extract-http-credentials credentials))
         ssh-credentials (when (and (= connection-type "application")
                                    (or (= connection-subtype "ssh")
                                        (= connection-subtype "git")
@@ -518,7 +520,7 @@
 
         http-connection-info (when (and (= connection-type "application")
                                         is-http-proxy-subtype?
-                                        (or (seq http-credentials) (seq http-env-vars)))
+                                        (or (seq http-credentials-deprecated) (seq http-env-vars)))
                                (let [env-vars-map (when (seq http-env-vars)
                                                     (reduce (fn [acc {:keys [key value]}]
                                                               (if (map? value)
@@ -526,13 +528,13 @@
                                                                 (assoc acc (keyword key) {:value (str value) :source "manual-input"})))
                                                             {}
                                                             http-env-vars))
-                                     combined-credentials (merge http-credentials env-vars-map)]
+                                     combined-credentials (merge http-credentials-deprecated env-vars-map)]
                                  (when (seq combined-credentials)
                                    (connection-method/infer-connection-method combined-credentials))))
 
         http-proxy-connection-info (when (and (= connection-type "httpproxy")
-                                              (seq http-credentials-v1))
-                                     (connection-method/infer-connection-method http-credentials-v1))
+                                              (seq http-credentials))
+                                     (connection-method/infer-connection-method http-credentials))
 
         inferred-connection-info (cond
                                    (seq normalized-credentials)
@@ -578,7 +580,7 @@
                           "manual-input")
      :secrets-manager-provider (when inferred-connection-info
                                  (:secrets-manager-provider inferred-connection-info))
-     :network-credentials (or network-credentials http-credentials http-credentials-v1)
+     :network-credentials (or network-credentials http-credentials http-credentials-deprecated)
      :ssh-credentials ssh-credentials
      :kubernetes-token kubernetes-token
      :ssh-auth-method (or ssh-auth-method "password")


webapp/src/webapp/connections/views/setup/events/process_form.cljs
```diff
@@ -402,6 +402,9 @@
         http-credentials (when (and (= connection-type "application")
                                     is-http-proxy-subtype?)
                            (extract-http-credentials credentials))
+
+        http-credentials-v1 (when (= connection-type "httpproxy")
```
mtmr0x: why `v1`?

matheusfrancisco: because there is this name already here http-credentials 
could be http-credentials and change the older http-credentials-deprecated.

Basically, this is needed to keep every connection as (custom,*) (application, * ) that was treated like httpproxy. If we remove this from the frontend, those connections will not work in the edit forms.

matheusfrancisco: just couldn't think of a name.

mtmr0x: oh, naming is hard, I feel you haha. Tho, we need to either name it properly so the next engineer without any context can know what's happening in here. And even with better naming, add a comment saying why things are happening like this so we have future reference

matheusfrancisco: okay

common/proto/types.go
```diff
@@ -213,22 +213,21 @@ func IsInList(item string, items []string) bool {
...
-			return ConnectionType(ConnectionTypeHttpProxy)
-		case "kibana":
-			return ConnectionType(ConnectionTypeHttpProxy)
 		case "ssh":
 			return ConnectionType(ConnectionTypeSSH)
 		case "git":
 			return ConnectionType(ConnectionTypeSSH)
 		case "github":
 			return ConnectionType(ConnectionTypeSSH)
+		// TODO(chico): needs to keep it for backward compatibility
```
matheusfrancisco:
```suggestion
		// TODO(chico): remove this case in the future, for now we need it to keep it for backward compatibility
```

common/proto/types.go
```diff
@@ -244,8 +243,8 @@ func ToConnectionType(connectionType, subtype string) ConnectionType {
 			return ConnectionType(ConnectionTypeSSM)
 		case "kubernetes", "kubernetes-eks":
 			return ConnectionType(ConnectionTypeKubernetes)
-		// TODO(san): deprecate it in flavor of kubernetes type
-		case "kubernetes-token":
+		// TODO(chico): needs to keep it for backward compatibility
```
matheusfrancisco:
```suggestion
		// TODO(chico): remove this case in the future, for now we need it to keep it for backward compatibility
```

sandromello:
✅ Build Completed with Success, Version=1264.0.0-cfd0814

webapp/src/webapp/connections/views/setup/events/process_form.cljs
rogefm: To avoid this name, you can add an "OR" in the conditional above.

PR #1317, opened 2026-03-11
feat: session creation and JIT review gate for native client credentials

rogefm:
## 📝 Description

Native client credentials already existed in Hoop, but every request happened without an audit trail and without any review gate. Sessions were only created later, when the native client actually connected to the proxy — meaning credentials could be handed out before any approval had a chance to intervene.

This PR moves session creation to the moment of the credential request, and wires the existing review/JIT system into the credentials flow. If a connection has reviewers or a JIT access request rule, the endpoint now returns a pending review instead of credentials. Once approved, a new resume endpoint releases the credentials tied to the original session.

All native client connections using the same credential now share a single session in the audit trail, regardless of how many TCP connections are opened.

## 🔗 Related Issue

Fixes ENG-266

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- **Session created upfront** — `POST /credentials` now creates a session record immediately, before any credential is issued, enabling review gating and a complete audit trail from request to expiry
- **JIT review gate** — `checkConnectionRequiresReview()` evaluates both OSS reviewers and enterprise `AccessRequestRule` (JIT type); if matched, returns HTTP 202 with a pending review and no credentials
- **Resume endpoint** — new `POST /credentials/:sessionID` releases credentials after reviewer approval, preserving the original expiry timer across multiple calls
- **`session_id` on credentials** — all five proxy daemons (postgres, ssh, http-proxy, rdp, ssm) now read `session_id` from the credential and reuse the pre-existing session instead of generating a new one
- **Audit plugin** — `closeSession()` skips closing sessions that have an active credential linked to them; the credential expiry owns the session lifecycle, not the client disconnect
- **Migration** — adds `session_id` column to `connection_credentials` with a unique constraint per org
- **Frontend: JIT notice** — when a connection has a JIT rule, the duration picker is replaced by a callout showing the fixed access window and a "Request Access" button
- **Frontend: review flow** — on a `has_review` response, the native client modal is replaced by the session details modal, giving the user visibility into the pending review and a "Connect" button once approved
- **`jit_access_duration_sec` on `GET /connections/:id`** — exposes the JIT rule's fixed duration so the frontend can render the correct UI before the credentials request is made

### Test Configuration:
- **OS**: macOS, Linux

### Tests performed:
- [ ] Unit tests pass
- [ ] Integration tests pass

**Scenario 1 — No review:** request credentials on a connection without reviewers or JIT rules → credentials returned immediately, session visible in audit trail, proxy connections grouped under the same session.

**Scenario 2 — OSS reviewers:** request credentials on a connection with `reviewers` configured → session details modal opens with PENDING status, reviewer approves, "Connect" button appears, resume endpoint returns credentials with the original expiry timestamp.

**Scenario 3 — JIT rule:** same as scenario 2, but the duration picker is replaced by the fixed-window callout before the request is even made.

**Scenario 4 — Multiple connections:** connect with the same credential from two clients simultaneously → both connections appear under the same session in audit, disconnecting one does not close the session.

**Scenario 5 — Expiry:** wait for credentials to expire → session closes automatically on the next request (lazy cleanup), expired credentials are rejected by the proxy.

## Support material

https://www.notion.so/hoopdotdev/POC-Native-Client-Credentials-Session-JIT-Review-31f9704c5a7880a3864fefe78aa61fbf

## ✅ Checklist

## 📄 Additional Notes

- Server-side enforcement of `access_max_duration` from the JIT rule (currently surfaced to the frontend but not validated against the request)
- Real-time update in the session details modal after reviewer approval (currently requires the user to keep the modal open)
- The JIT callout only appears for enterprise rules; connections with OSS reviewers still show the duration picker

### DIFF
--- a/gateway/api/connections/connection_credentials.go
+++ b/gateway/api/connections/connection_credentials.go
@@ -41,7 +41,10 @@ func CreateConnectionCredentials(c *gin.Context) {
 	ctx := storagev2.ParseContext(c)
 
 	// Lazy cleanup of expired credential sessions
-	_ = models.CloseExpiredCredentialSessions()
+	err := models.CloseExpiredCredentialSessions()
+	if err != nil {
+		log.Errorf("failed to close expired credential sessions, err=%v", err)
+	}
 
 	var req openapi.ConnectionCredentialsRequest
 	if err := c.ShouldBindJSON(&req); err != nil {
@@ -191,7 +194,10 @@ func ResumeConnectionCredentials(c *gin.Context) {
 	ctx := storagev2.ParseContext(c)
 
 	// Lazy cleanup of expired credential sessions
-	_ = models.CloseExpiredCredentialSessions()
+	err := models.CloseExpiredCredentialSessions()
+	if err != nil {
+		log.Errorf("failed to close expired credential sessions, err=%v", err)
+	}
 
 	var req openapi.ConnectionCredentialsRequest
 	if err := c.ShouldBindJSON(&req); err != nil {
--- a/gateway/api/session/session.go
+++ b/gateway/api/session/session.go
@@ -381,10 +381,13 @@ func CoerceMetadataFields(metadata map[string]any) error {
 //	@Router			/sessions [get]
 func List(c *gin.Context) {
 	ctx := storagev2.ParseContext(c)
-	
+
 	// Lazy cleanup of expired credential sessions
-	_ = models.CloseExpiredCredentialSessions()
-	
+	err := models.CloseExpiredCredentialSessions()
+	if err != nil {
+		log.Errorf("failed to close expired credential sessions, err=%v", err)
+	}
+
 	option := models.NewSessionOption()
 	for _, optKey := range openapi.AvailableSessionOptions {
 		if queryOptVal, ok := c.GetQuery(string(optKey)); ok {
@@ -476,7 +479,10 @@ func Get(c *gin.Context) {
 	ctx := storagev2.ParseContext(c)
 
 	// Lazy cleanup of expired credential sessions
-	_ = models.CloseExpiredCredentialSessions()
+	err := models.CloseExpiredCredentialSessions()
+	if err != nil {
+		log.Errorf("failed to close expired credential sessions, err=%v", err)
+	}
 
 	sessionID := c.Param("session_id")
 	apiroutes.SetSidSpanAttr(c, sessionID)


gateway/api/connections/connection_credentials.go
```diff
@@ -79,8 +86,53 @@ func CreateConnectionCredentials(c *gin.Context) {
...
+		Verb:              proto.ClientVerbConnect,
+		Status:            string(openapi.SessionStatusOpen),
+		CreatedAt:         time.Now().UTC(),
+	}
+
+	// Check if connection requires review/JIT approval
+	requiresReview, accessRule := checkConnectionRequiresReview(ctx, conn)
+	
+	// Persist session
+	if err := services.UpsertSession(c, newSession, *conn); err != nil {
```
p3rotto: The services.UpsertSession validates missing mandatory metadata and the session here is not filling the Metadata field.

If we want to keep the mandatory metadata validation then we need to start sending Metadata here. The other option is to use the models.UpsertSession directly to ignore missing mandatory metadata

rogefm: I've initialized Metadata as an empty map to maintain consistency with the rest of the codebase (see gateway/api/session/session.go:131) and preserve the mandatory metadata validation. Also added the same ErrMissingMetadata error handling pattern used in the session API.

p3rotto: But do we need to validate metadata on this endpoint too?

If we need then we have to change the `openapi.ConnectionCredentialsRequest` to receive metadata in the request body.

If we don't need to validate mandatory metadata on this endpoint then we should use the models.UpsertSession function

gateway/transport/plugins/audit/audit.go
```diff
@@ -233,6 +233,17 @@ func (p *auditPlugin) OnDisconnect(pctx plugintypes.Context, err error) error {
...
+	// Check if this session is owned by a connection credential
+	// If so, don't close it - the session lifecycle is managed by credential expiry
+	isCredentialSession := p.isCredentialOwnedSession(pctx.SID, pctx.OrgID)
+	
+	if isCredentialSession {
+		log.With("sid", pctx.SID, "origin", pctx.ClientOrigin, "verb", pctx.ClientVerb).
+			Infof("skipping session close - session is owned by connection credentials")
+		memorySessionStore.Del(pctx.SID)
+		return
+	}
```
matheusfrancisco: We cannot return here because it does not flush the session logs. The logs are flushed in the after this 
if isCredentialSession {..}

matheusfrancisco: You can remove the code; it will keep working.

webapp/src/webapp/audit/views/session_details.cljs
```diff
@@ -158,6 +158,17 @@
...
+              ;; credentials_expire_at is stored in session metadata by the backend
+              ;; when connection credentials are issued (see SetSessionCredentialsExpireAt)
+              credentials-expire-at (get-in session [:metadata :credentials_expire_at])
+              credentials-expired? (when credentials-expire-at
+                                     (<= (.getTime (js/Date. credentials-expire-at))
+                                         (.getTime (js/Date.))))
+              credentials-remaining-min (when (and credentials-expire-at (not credentials-expired?))
+                                          (max 1 (js/Math.ceil (/ (- (.getTime (js/Date. credentials-expire-at))
+                                                                     (.getTime (js/Date.)))
+                                                                  60000))))
```
matheusfrancisco, line 171:
```suggestion
              credentials-expire-at (get-in session [:metadata :credentials_expire_at])
              expires-ms (some-> credentials-expire-at js/Date. .getTime)
              now-ms  (.getTime (js/Date.))
              ms-remaning (when expire-ms (- expire-ms now-ms))
              credentials-expired? (when ms-remaning (<= ms-remaning 0))
              
              credentials-remaining-min (when (and ms-remaning (not credentials-expired?))
                                                                       (max 1 (js/Math.ceil (/ ms-remaning 60000))))

```

gateway/transport/plugins/audit/audit.go
rogefm: Without this return, in the first interaction with the resource, the session will be closed, and we can't do that.
If you need more context about the flow, let me know.

rogefm: Same here

gateway/models/connection_credentials.go
```diff
@@ -60,3 +61,51 @@ func GetConnectionByTypeAndID(connectionType, id string) (*ConnectionCredentials
...
+
+// UpdateConnectionCredentialsSecretKey updates the secret key hash of an existing credential
+func UpdateConnectionCredentialsSecretKey(id, secretKeyHash string) error {
+	return DB.Table("private.connection_credentials").
+		Where("id = ?", id).
+		Update("secret_key_hash", secretKeyHash).Error
+}
+
+// CloseExpiredCredentialSessions closes sessions for expired connection credentials
+// This is called lazily when accessing credentials or sessions
```
matheusfrancisco, line 86:
```suggestion
```

PR #1339, opened 2026-03-27
add: CLI agent-friendly UX enhancements (exec, login, connect)

andriosrobert:
## Summary
- **`exec --output json`**: Emits structured NDJSON events for each state transition (`waiting_approval`, `approved`, `running`, `completed`). Implies `--auto-approve`. On review-required connections, returns immediately with `session_id` and `review_url` instead of blocking.
- **`exec --session <id>`**: Executes a previously approved reviewed session via API. Enables the full non-blocking agent workflow: submit → poll → execute.
- **`login --token <key>`**: Non-interactive authentication for agents using API keys or pre-existing tokens.
- **`connect --output json`**: Structured JSON events for approval wait and proxy connection info (host/port).
- **`connect --duration` fix**: Only sends JIT timeout when `--duration` is explicitly set, fixing "exceeds max duration" errors on connections with lower limits.

## Agent workflow (exec with review)
```bash
# 1. Submit — returns immediately
hoop exec db-write -i 'SELECT 1' --output json
# {"status":"waiting_approval","data":{"review_url":"...","session_id":"abc-123"}}

# 2. Poll — check review.status
hoop admin get sessions abc-123 -o json

# 3. Execute approved session
hoop exec --session abc-123 --output json
# {"status":"completed","exit_code":0,"data":{"stdout":"..."}}
```

## Test plan
- [x] `go test ./client/cmd/ -v` — all 12 tests pass
- [x] `go build ./client/...` — compiles clean
- [x] `exec --help` shows `--output` and `--session` flags
- [x] `login --help` shows `--token` flag
- [x] `connect --help` shows `--output` flag
- [x] Tested exec JSON flow end-to-end: submit → poll PENDING → approve → poll APPROVED → exec --session
- [x] Tested connect without --duration no longer fails with max duration error

### DIFF
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -145,11 +145,13 @@ func runConnect(args []string, clientEnvVars map[string]string, durationFlagChan
 		case pbclient.SessionOpenWaitingApproval:
 			if jsonMode {
 				reviewURL := string(pkt.Payload)
+				sessionID := string(pkt.Spec[pb.SpecGatewaySessionID])
 				emitJSONEvent(os.Stdout, JSONEvent{
 					Status:  "waiting_approval",
-					Message: "poll status with: hoop admin get reviews " + string(pkt.Spec[pb.SpecGatewaySessionID]) + " -o json",
+					Message: "poll status with: hoop admin get sessions " + sessionID + " -o json | check review.status field",
 					Data: map[string]string{
-						"review_url": reviewURL,
+						"review_url":  reviewURL,
+						"session_id":  sessionID,
 					},
 				})
 				os.Exit(0)
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -147,11 +147,12 @@ func runConnect(args []string, clientEnvVars map[string]string, durationFlagChan
 				reviewURL := string(pkt.Payload)
 				emitJSONEvent(os.Stdout, JSONEvent{
 					Status:  "waiting_approval",
-					Message: "waiting task to be approved",
+					Message: "waiting task to be approved, poll GET /api/reviews/{session_id} for status",
 					Data: map[string]string{
 						"review_url": reviewURL,
 					},
 				})
+				os.Exit(0)
 			} else {
 				loader.Color("yellow")
 				if !loader.Active() {
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -83,6 +83,7 @@ func init() {
 	connectCmd.Flags().StringSliceVarP(&inputEnvVars, "env", "e", nil, "Input environment variables to send")
 	connectCmd.Flags().StringVarP(&connectFlags.duration, "duration", "d", "30m", "The amount of time that the session will last. Valid time units are 's', 'm', 'h'")
 	connectCmd.Flags().BoolVarP(&silentMode, "silent", "s", false, "Silent mode")
+	connectCmd.Flags().StringVarP(&outputFlag, "output", "o", "", "Output format. One of: (json)")
 	rootCmd.AddCommand(connectCmd)
 }
 
@@ -96,10 +97,13 @@ type connect struct {
 }
 
 func runConnect(args []string, clientEnvVars map[string]string) {
+	jsonMode := outputFlag == "json"
 	config := clientconfig.GetClientConfigOrDie()
 	loader := spinner.New(spinner.CharSets[11], 70*time.Millisecond)
 	loader.Color("green")
-	loader.Start()
+	if !jsonMode {
+		loader.Start()
+	}
 	defer func() { loader.Stop() }()
 	ossig := &osInterrupt{shutdownFn: func() { loader.Stop() }}
 	ossig.handleOsInterrupt()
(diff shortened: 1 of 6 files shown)


client/cmd/connect.go
```diff
@@ -156,6 +195,17 @@ func runConnect(args []string, clientEnvVars map[string]string) {
...
 				fmt.Println("ready to accept connections!")
+				if jsonMode {
+					emitJSONEvent(os.Stdout, JSONEvent{
+						Status:  "ready",
+						Message: "ready to accept connections",
+						Data: map[string]string{
+							"host": srv.Host().Host,
+							"port": srv.Host().Port,
+							"user": "noop",
+						},
```
matheusfrancisco, line 206: It looks like the emitJSONEvent is a duplicate. This could be a function 

for e.g you could pass the status and duplicate info 

```go
func emitReadyEvent(host, port string, extra map[string]string) {
    data := map[string]string{"host": host, "port": port}
    for k, v := range extra { data[k] = v }
    emitJSONEvent(os.Stdout, JSONEvent{Status: "ready", Message: "ready to accept connections", Data: data})
}
```

client/cmd/exec.go
```diff
@@ -196,40 +278,171 @@ func runExec(args []string, clientEnvVars map[string]string) {
...
+			Status:   status,
+			ExitCode: &exitCode,
+			Data:     data,
+		})
+		os.Exit(exitCode)
+	}
+
+	// Non-JSON mode: print output directly
+	fmt.Print(string(body))
+}
```
matheusfrancisco, line 448: Extract error handling into a helper to eliminate repetition
Extract request building into a helper or use a shared client

Avoid map[string]any for known API response shapes
Using map[string]any with type-asserting .(float64) and .(string) is fragile. 
Define a struct: This makes zero-value behavior explicit and removes the float64→int cast.
```go
type execSessionResponse struct {
    ExitCode     int    `json:"exit_code"`
    Output       string `json:"output"`
    OutputStatus string `json:"output_status"`
}
```
Separate concerns: HTTP call vs. output rendering

```go

type execSessionResponse struct {
	ExitCode     int    `json:"exit_code"`
	Output       string `json:"output"`
	OutputStatus string `json:"output_status"`
}

func doExecSession(config *clientconfig.ClientConfig, sessionID string) (*http.Response, []byte, error) {
	url := fmt.Sprintf("%s/api/sessions/%s/exec", config.ApiURL, sessionID)
	req, err := http.NewRequest(http.MethodPost, url, nil)
	if err != nil {
		return nil, nil, err
	}

	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", config.Token))
	if config.IsApiKey() {
		req.Header.Set("Api-Key", config.Token)
	}
	req.Header.Set("User-Agent", fmt.Sprintf("hoopcli/%v", version.Get().Version))

	resp, err := httpclient.NewHttpClient(config.TlsCA()).Do(req)
	if err != nil {
		return nil, nil, err
	}
	defer resp.Body.Close()

	body, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, nil, fmt.Errorf("reading response body: %w", err)
	}

	return resp, body, nil
}

func handleExecErr(jsonMode bool, format string, args ...any) {
	msg := fmt.Sprintf(format, args...)
	if jsonMode {
		exitCode := 1
		emitJSONEvent(os.Stdout, JSONEvent{Status: "error", Message: msg, ExitCode: &exitCode})
	} else {
		styles.PrintErrorAndExit(msg)
	}
	os.Exit(1)
}

// runExecSession executes an already-approved reviewed session via the API.
// Used by agents to trigger execution after polling for approval.
func runExecSession(sessionID string) {
	jsonMode := outputFlag == "json"
	config := clientconfig.GetClientConfigOrDie()

	resp, body, err := doExecSession(config, sessionID)
	if err != nil {
		handleExecErr(jsonMode, err.Error())
	}

	if resp.StatusCode >= 400 {
		handleExecErr(jsonMode, "failed executing session: %s", string(body))
	}

	if !jsonMode {
		fmt.Print(string(body))
		return
	}

	var apiResp execSessionResponse
	if err := json.Unmarshal(body, &apiResp); err != nil {
		handleExecErr(jsonMode, "parsing response: %v", err)
	}

	status := "completed"
	if resp.StatusCode == http.StatusAccepted {
		status = "running"
	}

	data := map[string]string{"session_id": sessionID}
	if apiResp.Output != "" {
		data["stdout"] = apiResp.Output
	}
	if apiResp.OutputStatus != "" {
		data["output_status"] = apiResp.OutputStatus
	}

	emitJSONEvent(os.Stdout, JSONEvent{
		Status:   status,
		ExitCode: &apiResp.ExitCode,
		Data:     data,
	})
	os.Exit(apiResp.ExitCode)
}

```

client/cmd/login.go
```diff
@@ -42,6 +43,41 @@ var loginCmd = &cobra.Command{
...
+			if err != nil {
+				printErrorAndExit(err.Error())
+			}
+			if saved {
+				fmt.Println("Login succeeded")
+			} else {
+				fmt.Println(conf.Token)
+			}
+			return
+		}
```
matheusfrancisco, line 80: Split into a function 

```go

func runTokenLogin(token string) {
    if err := validateLoginToken(token); err != nil {
        printErrorAndExit(err.Error())
    }

    conf, err := proxyconfig.Load()
    switch {
    case err == proxyconfig.ErrEmpty || (err == nil && !conf.IsValid()):
        configureHostsPrompt(conf)
    case err != nil:
        printErrorAndExit(err.Error())
    }

    conf.Token = token

    if conf.GrpcURL == "" {
        si, err := fetchServerInfo(conf.ApiURL, conf.Token, conf.TlsCA())
        if err != nil {
            printErrorAndExit(err.Error())
        }
        conf.GrpcURL = si.GrpcURL
    }

    saved, err := conf.Save()
    if err != nil {
        printErrorAndExit(err.Error())
    }

    if saved {
        fmt.Println("Login succeeded")
    } else {
        fmt.Println(conf.Token)
    }
}
```

client/cmd/login.go
```diff
@@ -42,6 +43,41 @@ var loginCmd = &cobra.Command{
...
+			saved, err := conf.Save()
+			if err != nil {
+				printErrorAndExit(err.Error())
+			}
+			if saved {
+				fmt.Println("Login succeeded")
+			} else {
+				fmt.Println(conf.Token)
+			}
+			return
```
matheusfrancisco, line 79:
```go
 if loginToken != "" {
        runTokenLogin(loginToken)
        return
    }
```

PR #1362, opened 2026-04-07
feat(client): add batch connection creation via YAML file

feraio:
## 📝 Description

Add `hoop admin create connections --from-file` command that allows admins to create multiple connections from a single YAML file, reducing friction during environment onboarding.

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- New `hoop admin create connections --from-file FILE` command (plural `connections`, aliased as `conns`)
- YAML file format with top-level defaults (`agent`, `type`, `tags`, `access_modes`, `overwrite`) and per-connection overrides
- Connection type is required — no silent fallback to `custom`, preventing misconfigured resources
- Supports same env value formats as single create: raw, `base64://`, `file:///`
- Progress output with `[n/total]` prefix and error summary on completion
- Unit tests for YAML parsing, env var resolution, and validation
- Example YAML file at `client/cmd/admin/examples/batch-connections.yaml`
- Registered new command in `create.go` — no gateway/API changes needed

### Test Configuration:
- **OS**: macOS (darwin/arm64)

### Tests performed:
- [ ] Unit tests pass
- [ ] Integration tests pass

> **Note**: Go toolchain was not available in the authoring environment. Unit tests are written but need to be compiled and run. Manual testing with a live Hoop instance is pending.

### Example YAML file

```yaml
agent: default
type: database/postgres
tags:
  env: production
connections:
  - name: users-db
    env:
      HOST: 10.0.1.10
      USER: admin
      PASS: file:///secrets/users-db.txt
      PORT: "5432"
  - name: orders-db
    env:
      HOST: 10.0.1.11
      USER: admin
      PASS: base64://c2VjcmV0
      PORT: "5432"
```

**Expected output:**
```creating 2 connection(s) from connections.yaml

[1/2] connection users-db created
[2/2] connection orders-db created

done: 2 succeeded, 0 failed out of 2 total
```

## ✅ Checklist

- [ ] I have run `make generate-openapi-docs` to update the OpenAPI docs

### Scope
This is a **client-side only** change. No gateway or API modifications. The new command iterates over YAML entries and calls the existing `POST /api/connections` endpoint for each one.

### Out of scope (considered for next version)
- **Plugin support** — YAML format does not support a `plugins` field per connection
- **Dry-run mode** — `--dry-run` flag to validate the file without creating resources
- **Parallelism** — creation is sequential; concurrent execution for large batches may be added later
- **Additional flags** — `reviewers`, `guardrail-rules`, `redact-types` are not yet exposed in the YAML format
- **Move to `resources` subcommand** — the `admin` subcommand is experimental and not fully aligned with the webapp experience. A future version should introduce `hoop resources create --from-file` (or similar) to match the webapp's resource model and move away from `admin`
- **Resource Roles support** — the webapp treats Roles as a first-class concept (named, multiple per resource, with different credential methods). The current YAML format maps to flat env vars. A future version should support defining multiple roles per connection, aligning CLI and webapp experiences

### DIFF
--- a/client/cmd/admin/create-conn-batch.go
+++ b/client/cmd/admin/create-conn-batch.go
@@ -1,0 +1,343 @@
+package admin
+
+import (
+	"encoding/base64"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/hoophq/hoop/client/cmd/styles"
+	clientconfig "github.com/hoophq/hoop/client/config"
+	"github.com/hoophq/hoop/common/log"
+	pb "github.com/hoophq/hoop/common/proto"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+// BatchConnectionFile represents the YAML file format for batch connection creation.
+//
+// Example file:
+//
+//	agent: default
+//	type: database/postgres
+//	overwrite: false
+//	tags:
+//	  env: production
+//	  team: platform
+//	access_modes:
+//	  - connect
+//	  - exec
+//	  - runbooks
+//	connections:
+//	  - name: users-db
+//	    env:
+//	      HOST: 10.0.1.10
+//	      USER: admin
+//	      PASS: file:///secrets/users-db.txt
+//	      PORT: "5432"
+//	  - name: orders-db
+//	    env:
+//	      HOST: 10.0.1.11
+//	      USER: admin
+//	      PASS: base64://c2VjcmV0
+//	      PORT: "5432"
+type BatchConnectionFile struct {
+	Agent       string            `yaml:"agent"`
+	Type        string            `yaml:"type"`
+	Overwrite   bool              `yaml:"overwrite"`
+	Tags        map[string]string `yaml:"tags"`
+	AccessModes []string          `yaml:"access_modes"`
+	Schema      string            `yaml:"schema"`
+	Connections []BatchConnection `yaml:"connections"`
+}
+
+// BatchConnection represents a single connection entry in the batch file.
+// Each entry can override the top-level agent, type, tags, and access_modes.
+type BatchConnection struct {
+	Name        string            `yaml:"name"`
+	Agent       string            `yaml:"agent"`
+	Type        string            `yaml:"type"`
+	Env         map[string]string `yaml:"env"`
+	Command     []string          `yaml:"command"`
+	Tags        map[string]string `yaml:"tags"`
+	AccessModes []string          `yaml:"access_modes"`
+	Schema      string            `yaml:"schema"`
+	Overwrite   *bool             `yaml:"overwrite"`
+}
+
+var fromFileFlag string
+
+func init() {
+	createConnectionBatchCmd.Flags().StringVar(&fromFileFlag, "from-file", "", "Path to a YAML file defining multiple connections")
+	createConnectionBatchCmd.MarkFlagRequired("from-file")
+}
+
+var createConnectionBatchExamplesDesc = `
+  # Create multiple postgres connections from a YAML file
+  hoop admin create connections --from-file connections.yaml
+
+  # Example YAML file (connections.yaml):
+  #
+  # agent: default
+  # type: database/postgres
+  # connections:
+  #   - name: users-db
+  #     env:
+  #       HOST: 10.0.1.10
+  #       USER: admin
+  #       PASS: secret
+  #       PORT: "5432"
+  #   - name: orders-db
+  #     env:
+  #       HOST: 10.0.1.11
+  #       USER: admin
+  #       PASS: file:///secrets/orders-db.txt
+  #       PORT: "5432"
+`
+
+var createConnectionBatchCmd = &cobra.Command{
+	Use:     "connections --from-file FILE",
+	Aliases: []string{"conns"},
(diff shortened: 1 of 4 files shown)


client/cmd/admin/examples/batch-connections.yaml
```diff
@@ -0,0 +1,73 @@
...
+  team: platform
+
+# Optional: shared access modes (default: connect, exec, runbooks)
+access_modes:
+  - connect
+  - exec
+  - runbooks
+
+# List of connections to create
+connections:
```
rogefm, line 21: We should use resource roles instead of connections

client/cmd/admin/examples/batch-connections.yaml
```diff
@@ -0,0 +1,73 @@
+# Example: Batch connection creation for hoop CLI
+# Usage: hoop admin create connections --from-file batch-connections.yaml
+
+# Top-level defaults (applied to all connections unless overridden)
+agent: default
+type: database/postgres
```
rogefm, line 6: I think we should use only Postgres as a type, database/application/http-proxy/custom is something internal

client/cmd/admin/create-conn-batch.go
```diff
@@ -0,0 +1,343 @@
...
+		return fmt.Errorf("no agent specified (set at top-level or per-connection)")
+	}
+
+	connTypeStr := batch.Type
+	if conn.Type != "" {
+		connTypeStr = conn.Type
+	}
+	if connTypeStr == "" {
+		connTypeStr = "custom"
+	}
```
matheusfrancisco, line 168: If there is no type, it should fail and issue a warning to the user. Adding it as a custom could lead to an error, for example, a database type will be added as a custom and will not work, no?

mtmr0x: agreed

client/cmd/admin/examples/batch-connections.yaml
feraio: this is how we create a connection today, the idea is not changing anything our default method currently has. it can be done in an update for both connection creation flows in a further step.

client/cmd/admin/examples/batch-connections.yaml
feraio: same as the resource type, we're keeping the same standards from the current experience, once we decide if it will have to be updated we do it for both flows.

sandromello:
✅ Build Completed with Success, Version=1362.0.0-a7d5c69

client/cmd/admin/create-conn-batch.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+var fromFileFlag string
+
+func init() {
+	createConnectionBatchCmd.Flags().StringVar(&fromFileFlag, "from-file", "", "Path to a YAML file defining multiple connections")
+	createConnectionBatchCmd.MarkFlagRequired("from-file")
+}
+
+var createConnectionBatchExamplesDesc = `
```
racerxdl, line 75: `fromFileFlag` is a package-level var. If another subcommand in the `admin` package ever uses a `--from-file` flag, they'll collide. Consider scoping this to the command's `PersistentPreRun` or using `cmd.Flags().GetString("from-file")` inside the `Run` function instead.

Automated by MisterMal 🤖

racerxdl:
## Code Review — PR #1362: Batch Connection Creation CLI

Clean, well-structured client-side feature. Good use of top-level defaults with per-connection overrides, and the YAML format is intuitive. Tests cover parsing, env var resolution, and validation. A few inline comments on the diff — mostly around shared package-level flags (`skipStrictValidation`, `outputFlag`, `fromFileFlag`) that are bound to the single-create command but referenced by the batch command without registering their own flags. Nothing blocking.

Automated by MisterMal 🤖

racerxdl (approved):

p3rotto (approved):

luanlorenzo (approved):

client/cmd/admin/create-conn-batch.go
feraio: @rogefm can you take a look into this?

matheusfrancisco (approved):

client/cmd/admin/examples/batch-connections.yaml
rogefm: This is not how we create today :(
We use type and subtype separately, and doing so further consists of break-changes, which is bad

client/cmd/admin/create-conn-batch.go
rogefm: The pattern of package-level flag vars bound in init() is consistent across the entire admin package (~25+ flag vars across create-conn.go, create-user.go, create-plugin.go, etc.). Since fromFileFlag is a local flag (.Flags(), not .PersistentFlags()), Cobra isolates parsing per command — no runtime collision is possible today. The naming is specific enough to avoid accidental reuse. If we want to address this pattern, it would need to be a package-wide refactor, not a one-off fix here.

rogefm (approved):

PR #1364, opened 2026-04-07
feat: add refresh token support for OIDC webapp sessions

andriosrobert:
## Summary

When OIDC access tokens expire during a webapp session, users are currently forced to re-login. This PR adds transparent token refresh so the webapp can continue operating without interruption.

The scope is intentionally limited to the **HTTP REST API path only** (webapp). gRPC and proxy server auth paths are unchanged and will be addressed in follow-up PRs.

### DB migration (`000069`)
- Adds a nullable `refresh_token TEXT` column to `private.user_tokens`.
- The column is nullable so existing rows are unaffected — no data backfill needed.

### OIDC provider (`gateway/idp/oidc/oidc.go`)
- Requests `offline_access` scope and `AccessTypeOffline` so the IdP returns a refresh token alongside the access token.
- `RefreshAccessToken()` — exchanges a refresh token for a new access/refresh token pair via the standard OAuth2 token endpoint.
- `VerifyExpiredTokenSubject()` — uses `jwt.Parse` with `jwt.WithoutClaimsValidation()` to **validate the JWKS signature** while skipping expiry checks, then extracts the `sub` claim. This is the defense-in-depth measure that ensures forged tokens can never trigger the refresh flow.

### Login callbacks
- **OIDC** (`gateway/api/login/oidc/login.go`): stores the refresh token when present.
- **Local** and **SAML**: pass `nil` as the refresh token (signature change only, no behavioral change).

### Model (`gateway/models/user_tokens.go`)
- `UserToken` struct gains an optional `RefreshToken *string` field.
- `UpsertUserToken` accepts an optional refresh token. When `nil`, the upsert only updates the access token column — **existing refresh tokens are never overwritten accidentally**.

### IDP helpers (`gateway/idp/core.go`)
- `TokenRefresher` interface — optional capability for providers that support refresh.
- `ExpiredTokenSubjectVerifier` interface — optional capability for providers that can extract a subject from an expired JWT with signature validation.
- `TryRefreshExpiredToken()` — single entry point that validates the expired JWT signature (via JWKS), extracts the subject, looks up the stored refresh token, refreshes, and persists the new token pair.

### Webapp API middleware (`gateway/api/apiroutes/auth.go`)
- On `"token is expired"` error, attempts to refresh via `tryRefreshFromRequest()`.
- On success, returns the new access token to the client through:
  - `X-New-Access-Token` response header
  - Updated `hoop_access_token` cookie (same attributes as the login callback)
- On failure, falls through to the existing 401 response — **identical to current behavior**.

## Security considerations

**Can an attacker forge a token to trigger the refresh flow?**

No. The flow has two layers of protection:

1. **`jwt.Parse` validation order (jwt/v5)**: signature is validated *before* claims (expiry). A forged token produces `ErrTokenSignatureInvalid`, which does not match the `"token is expired"` check — the refresh path is never entered.

2. **`VerifyExpiredTokenSubject` (defense-in-depth)**: even if the code path changes in the future, the subject extraction itself re-validates the JWT signature via JWKS using `jwt.WithoutClaimsValidation()`. A forged token fails signature verification and returns an error before any subject is extracted or any refresh is attempted.

Only a legitimately-signed-but-expired JWT (issued by the configured IdP) can trigger a refresh — which is the intended behavior.

## Why this has no impact on existing behavior

1. **Additive DB change**: the new column is nullable with no default, so the migration is a no-op for existing data and fully reversible (`DROP COLUMN IF EXISTS`).
2. **Non-OIDC providers are unaffected**: Local and SAML pass `nil` for the refresh token. `ExpiredTokenSubjectVerifier` and `TokenRefresher` are checked at runtime via type assertion — providers that don't implement them simply skip refresh and fall through to the existing 401.
3. **Opaque tokens are unaffected**: `VerifyExpiredTokenSubject` only works on JWTs verifiable via JWKS. Opaque/bearer tokens fail and fall through to the normal error path.
4. **gRPC and proxy paths are untouched**: `gateway/transport/auth.go` and `gateway/transport/interceptors/auth/auth.go` have zero diff against `main`.
5. **Refresh failure is safe**: every error path in the refresh flow falls through to the same 401 response the user would have received before this change.
6. **`offline_access` scope is safe**: IdPs that don't support it simply ignore the scope and don't return a refresh token. The code handles this (refresh token stored as `nil`, refresh path skipped).

## Test plan

- [ ] OIDC login stores a refresh token in `private.user_tokens` (verify via DB query)
- [ ] Local login still works (refresh token is `nil` in DB)
- [ ] SAML login still works (refresh token is `nil` in DB)
- [ ] Webapp API call with a valid token works as before
- [ ] Webapp API call with an expired OIDC token triggers refresh, returns `X-New-Access-Token` header and updated cookie
- [ ] Webapp API call with an expired token and no refresh token returns 401
- [ ] Webapp API call with an expired token and an invalid refresh token returns 401
- [ ] Webapp API call with a **forged** JWT returns 401 (signature check blocks refresh)
- [ ] Migration up/down works cleanly (`000069`)

https://claude.ai/code/session_01Qhwe8X4fLPCwAdYaVA1N1W


sandromello:
✅ Build Completed with Success, Version=1364.0.0-18d7873

p3rotto (approved):

gateway/idp/oidc/oidc.go
```diff
@@ -206,8 +206,39 @@ func downloadJWKS(jwksURL string) (*keyfunc.JWKS, error) {
...
+	subject, ok := claims["sub"].(string)
+	if !ok || subject == "" {
+		return "", fmt.Errorf("'sub' not found or empty in token")
+	}
+	return subject, nil
+}
+
 func (p *Provider) GetAudience() string { return p.Audience }
 func (p *Provider) GetAuthCodeURL(state string, opts ...oauth2.AuthCodeOption) string {
+	opts = append(opts, oauth2.AccessTypeOffline)
```
sandromello: The offline scope is still being referenced here. It should be included if the user has included only

gateway/api/apiroutes/auth.go
```diff
@@ -96,11 +96,17 @@ func (r *Router) AuthMiddleware(c *gin.Context) {
...
+		if strings.Contains(err.Error(), "token is expired") {
+			subject, err = r.tryRefreshFromRequest(tokenVerifier, c)
+		}
+		if err != nil {
+			tokenHeader := c.GetHeader("authorization")
+			log.Infof("failed authenticating, %v, length=%v, reason=%v, url-path=%v",
+				parseHeaderForDebug(tokenHeader), len(tokenHeader), err, c.Request.URL.Path)
+			c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"message": "access denied"})
+			return
+		}
```
sandromello, line 109: This code should only execute if the user provided the offline scope.

sandromello (commented):
It looks good to me with one caveat:

The offline access scope could break our current implementation, usually when adding a new scope the user needs to consent this new access, therefore I recommend not adding this attribute by default and allowing customers to include in the OIDC provider

sandromello (approved):

sandromello:
✅ Build Completed with Success, Version=1364.0.0-825669d

PR #1392, opened 2026-04-18
Self-service org migration for multi-tenant users

rogefm:
## 📝 Description                                                                                                  
                                                                                                                   
In multi-tenant environments (e.g., `use.hoop.dev`), users who sign in via SSO before being explicitly invited to  
their organization are automatically placed in a newly created org. When they later receive an invitation to the   
correct org, the database ends up with two rows for the same email: one active user in the wrong org and one       
invited user in the right org. Resolving this requires manual database intervention, generating recurring support  
tickets.

This PR introduces a self-service org migration flow. On the next login, users with a pending invitation see a     
dialog offering them to migrate to the correct organization. Accepting atomically promotes the invited record,     
binds the IDP subject to it, removes the old record, and cleans up the now-empty org. 

## 🚀 Type of Change                                                                                               
 
- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update                                                                                           
- [ ] ♻️  Code refactor                                    
- [ ] ✅ Test update                                      
- [ ] 🧹 Chore                                                                                                     
 
## 📋 Changes Made                                                                                                 
                                                          
**Backend**                                                                                                        
- Improved OIDC login user lookup: resolves by IDP subject first, falls back to email — ensures invited users (with
 placeholder subject) are matched correctly on first login                                                         
- `GET /userinfo` now returns `pending_org_invitations` for multi-tenant users, listing orgs they were invited to
but haven't joined                                                                                                 
- `POST /userinfo/accept-org-invitation`: atomically promotes the invited user record, transfers the IDP subject,
deletes the old record, and removes the orphaned org if empty                                                      
- `DELETE /userinfo/pending-org-invitation`: dismisses a pending invitation without migrating (dialog won't show
again)                                                                                                             
- Added model helpers: `ListUsersByEmail`, `GetInvitedUserByEmailAndOrg`, `DeleteOrganizationIfEmpty`
                                                                                                                   
**Frontend**                                              
- New `OrgMigrationDialog` component: shown globally on login when a pending invitation exists, built with Radix UI
 + Lucide icons                                                                                                    
- Re-frame events and subscriptions to handle accept/decline flows with proper loading and error states
- On accept, reloads the app after a short delay so the new org session takes effect cleanly                       
                                                                                                                   
## 🧪 Testing                                                                                                      
                                                                                                                   
### Test Configuration:                                   
- **Browser(s)**: Chrome
- **OS**: macOS                                                                                                    
 
### Tests performed:                                                                                               
- [ ] Unit tests pass                                                                                              
- [ ] Integration tests pass
                                                                                                                   
**Scenario tested:**                                      
1. User signs in via SSO before being invited → lands in auto-created org
2. Admin invites that same email to the correct org                                                                
3. User logs in again → migration dialog appears                                                                   
4. User accepts → migrated to correct org, old org deleted, SSO works normally                                     
5. User declines → dialog dismissed, user stays in current org                                                     
                                                                                                                   
## ✅ Checklist                                                                                                    
                                                                                                                   
- [ ] I have run `make generate-openapi-docs` to update the OpenAPI docs
                                                                                                                   
## 📄 Additional Notes
                                                                                                                   
The migration is intentionally scoped to multi-tenant environments only (`isOrgMultiTenant` gate on the backend).  
The org cleanup after migration is best-effort — if the old org still has other resources (connections, sessions,
agents), deletion is skipped with a warning log rather than failing the migration.

### DIFF
--- a/gateway/api/login/oidc/login.go
+++ b/gateway/api/login/oidc/login.go
@@ -161,34 +161,39 @@ func (h *handler) LoginCallback(c *gin.Context) {
 		return
 	}
 	uinfo.Subject = subject
-	// get the user by its email to get the actual subject of that user. This is necessary
-	// due to the user subject when it's created inside hoop is changed after that user
-	// logs in with the IDP. The email should always come from the IDP as a design of how
-	// we handle users in hoop.
-	dbUser, err := models.GetUserByEmail(uinfo.Email)
-	if err != nil {
-		login.Outcome = fmt.Sprintf("failed fetching user by email=%s, reason=%v", uinfo.Email, err)
-		log.Error(login.Outcome)
-		sentry.CaptureException(err)
-		c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
-		return
-	}
-
-	// if the user doesn't exist in the database, we should use the subject from the IDP
-	// to allow the user to login. This user will be a new user and will be created at
-	// the end of this method.
-	if dbUser == nil {
-		subject = uinfo.Subject
-	} else {
-		subject = dbUser.Subject
-	}
-	ctx, err := models.GetUserContext(subject)
+	// Try to find user by IDP subject first (unambiguous). Fall back to email lookup only
+	// when no user is found by subject, to handle users whose stored subject is still a
+	// placeholder (e.g. invited users who have not yet completed first login).
+	ctx, err := models.GetUserContext(uinfo.Subject)
 	if err != nil {
 		login.Outcome = fmt.Sprintf("failed fetching user subject=%s, email=%s, reason=%v", uinfo.Subject, uinfo.Email, err)
 		log.Error(login.Outcome)
 		sentry.CaptureException(err)
 		c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
 		return
+	}
+	if ctx.IsEmpty() {
+		dbUser, err := models.GetUserByEmail(uinfo.Email)
+		if err != nil {
+			login.Outcome = fmt.Sprintf("failed fetching user by email=%s, reason=%v", uinfo.Email, err)
+			log.Error(login.Outcome)
+			sentry.CaptureException(err)
+			c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
+			return
+		}
+		if dbUser != nil {
+			ctx, err = models.GetUserContext(dbUser.Subject)
+			if err != nil {
+				login.Outcome = fmt.Sprintf("failed fetching user subject=%s, email=%s, reason=%v", dbUser.Subject, uinfo.Email, err)
+				log.Error(login.Outcome)
+				sentry.CaptureException(err)
+				c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
+				return
+			}
+		}
+	}
+	if !ctx.IsEmpty() {
+		subject = ctx.UserSubject
 	}
 
 	err = models.UpsertUserToken(models.DB, subject, token.AccessToken)
--- a/gateway/api/login/oidc/login.go
+++ b/gateway/api/login/oidc/login.go
@@ -161,39 +161,34 @@ func (h *handler) LoginCallback(c *gin.Context) {
 		return
 	}
 	uinfo.Subject = subject
-	// Try to find user by IDP subject first (unambiguous). Fall back to email lookup only
-	// when no user is found by subject, to handle users whose stored subject is still a
-	// placeholder (e.g. invited users who have not yet completed first login).
-	ctx, err := models.GetUserContext(uinfo.Subject)
+	// get the user by its email to get the actual subject of that user. This is necessary
+	// due to the user subject when it's created inside hoop is changed after that user
+	// logs in with the IDP. The email should always come from the IDP as a design of how
+	// we handle users in hoop.
+	dbUser, err := models.GetUserByEmail(uinfo.Email)
+	if err != nil {
+		login.Outcome = fmt.Sprintf("failed fetching user by email=%s, reason=%v", uinfo.Email, err)
+		log.Error(login.Outcome)
+		sentry.CaptureException(err)
+		c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
+		return
+	}
+
+	// if the user doesn't exist in the database, we should use the subject from the IDP
+	// to allow the user to login. This user will be a new user and will be created at
+	// the end of this method.
+	if dbUser == nil {
+		subject = uinfo.Subject
+	} else {
+		subject = dbUser.Subject
+	}
+	ctx, err := models.GetUserContext(subject)
 	if err != nil {
 		login.Outcome = fmt.Sprintf("failed fetching user subject=%s, email=%s, reason=%v", uinfo.Subject, uinfo.Email, err)
 		log.Error(login.Outcome)
 		sentry.CaptureException(err)
 		c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
 		return
-	}
-	if ctx.IsEmpty() {
-		dbUser, err := models.GetUserByEmail(uinfo.Email)
-		if err != nil {
-			login.Outcome = fmt.Sprintf("failed fetching user by email=%s, reason=%v", uinfo.Email, err)
-			log.Error(login.Outcome)
-			sentry.CaptureException(err)
-			c.Redirect(http.StatusTemporaryRedirect, redirectErrorURL)
-			return
-		}
-		if dbUser != nil {
-			ctx, err = models.GetUserContext(dbUser.Subject)
-			if err != nil {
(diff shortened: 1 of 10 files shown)


sandromello:
✅ Build Completed with Success, Version=1392.0.0-0aefa1c

gateway/api/login/oidc/login.go
```diff
@@ -161,34 +161,39 @@ func (h *handler) LoginCallback(c *gin.Context) {
...
 	uinfo.Subject = subject
-	// get the user by its email to get the actual subject of that user. This is necessary
-	// due to the user subject when it's created inside hoop is changed after that user
-	// logs in with the IDP. The email should always come from the IDP as a design of how
-	// we handle users in hoop.
-	dbUser, err := models.GetUserByEmail(uinfo.Email)
+	// Try to find user by IDP subject first (unambiguous). Fall back to email lookup only
+	// when no user is found by subject, to handle users whose stored subject is still a
+	// placeholder (e.g. invited users who have not yet completed first login).
+	ctx, err := models.GetUserContext(uinfo.Subject)
```
sandromello, line 167: Could you explain why you need to change this file in order to introduce this change?
By what you're trying to accomplish it doesn't need to touch in this part of the code.

gateway/api/user/user.go
```diff
@@ -475,9 +475,187 @@ func GetUserInfo(c *gin.Context) {
...
+					log.Warnf("failed fetching org %s for pending invitation, err=%v", u.OrgID, err)
+					continue
+				}
+				userInfoData.PendingOrgInvitations = append(
+					userInfoData.PendingOrgInvitations,
+					openapi.PendingOrgInvitation{OrgID: org.ID, OrgName: org.Name},
+				)
+			}
+		}
+	}
```
sandromello, line 497: You're leaking organization context information. You should never expose information from other organizations

gateway/api/user/user.go
```diff
@@ -475,9 +475,187 @@ func GetUserInfo(c *gin.Context) {
...
+	name := invitedUser.Name
+	if ctx.UserName != "" {
+		name = ctx.UserName
+	}
+	picture := invitedUser.Picture
+	if ctx.UserPicture != "" {
+		picture = ctx.UserPicture
+	}
+
+	tx := models.DB.Begin()
```
sandromello, line 569: You should never access the transaction directly on the API layer. Make sure to wrap everything in a function in the models package

gateway/api/user/user.go
```diff
@@ -475,9 +475,187 @@ func GetUserInfo(c *gin.Context) {
...
+	// avoiding false matches when multiple records share the same email in an org
+	currentUser, err := models.GetUserBySubjectAndOrg(idpSubject, ctx.OrgID)
+	if err != nil {
+		httputils.AbortWithErr(c, http.StatusInternalServerError, err, "failed fetching current user")
+		return
+	}
+	if currentUser == nil {
+		c.JSON(http.StatusNotFound, gin.H{"message": "current user not found"})
+		return
+	}
```
sandromello, line 534: The user is already identified when parsing the context, so you don't need this.

gateway/api/user/user.go
```diff
@@ -475,9 +475,187 @@ func GetUserInfo(c *gin.Context) {
...
+
+	// Only match invited users to avoid accidentally targeting an active user from a previous failed migration
+	invitedUser, err := models.GetInvitedUserByEmailAndOrg(ctx.UserEmail, req.OrgID)
+	if err != nil {
+		httputils.AbortWithErr(c, http.StatusInternalServerError, err, "failed fetching invited user")
+		return
+	}
```
sandromello, line 541: The user should never pass the organization in which he wants to approve an invite. The endpoint should try to do a best effort and lookup for users in the database that has an active invite.

The proper behavior would be:
Is there any user in the database that matches my current email with the invited status? If it finds a single user, it means an administrator has invited the user and it could be promoted, in case it find multiple users the request must fail.

gateway/api/user/user.go
```diff
@@ -475,9 +475,187 @@ func GetUserInfo(c *gin.Context) {
...
+
+	// Update the invited user (except subject) first to confirm the record still exists
+	// before we commit to deleting the current user.
+	result := tx.Exec(`
+		UPDATE private.users
+		SET status = 'active', verified = true, name = ?, picture = ?
+		WHERE id = ? AND status = 'invited'`,
+		name, picture, invitedUser.ID)
```
sandromello, line 581: I think this will fail, because you need to delete the user with the ID first. It will hit in a constraint error

gateway/api/user/user.go
```diff
@@ -475,9 +475,187 @@ func GetUserInfo(c *gin.Context) {
...
+	if err != nil {
+		httputils.AbortWithErr(c, http.StatusInternalServerError, err, "failed fetching invited user")
+		return
+	}
+	if invitedUser == nil || invitedUser.Status != string(openapi.StatusInvited) {
+		c.JSON(http.StatusNotFound, gin.H{"message": "pending invitation not found for this organization"})
+		return
+	}
+
+	result := models.DB.Exec(`DELETE FROM private.users WHERE org_id = ? AND email = ? AND status = 'invited'`,
```
sandromello, line 649: This is not secure, you must not rely in the user providing the organization ID.
Do a best effort to lookup invitations based on the email address of the authenticated user.

gateway/api/server.go
```diff
@@ -217,6 +217,14 @@ func (api *Api) buildRoutes(r *apiroutes.Router) {
...
 		r.AuthMiddleware,
 		userapi.GetUserInfo)
+	r.POST("/userinfo/accept-org-invitation",
+		apiroutes.ReadOnlyAccessRole,
+		r.AuthMiddleware,
+		userapi.AcceptOrgInvitation)
+	r.DELETE("/userinfo/pending-org-invitation",
+		apiroutes.ReadOnlyAccessRole,
+		r.AuthMiddleware,
+		userapi.DeclineOrgInvitation)
```
sandromello, line 227: Please, change the name of the routes to something more REST friendly.
The userinfo is just to obtain user information

```json
POST /orgs/invitations
{ 
  "action": "accept|decline"
}
```

sandromello (changes requested):

gateway/api/login/oidc/login.go
rogefm: The change in login.go is needed for the feature to work correctly. When a user with a pending invitation logs in via SSO for the first time, their invited DB record has an empty/placeholder subject. The previous code looked up only by subject — this lookup returns nothing for invited users, causing the login callback to create a new user instead of finding the invited one. By falling back to email, we correctly identify the invited record and can surface the migration dialog. Without this change, the invitation flow is unreachable.

gateway/api/user/user.go
rogefm: Now we are only "leaking" the name.

sandromello:
✅ Build Completed with Success, Version=1392.0.0-e4c4855

gateway/api/openapi/types.go
```diff
@@ -111,6 +127,9 @@ type UserInfo struct {
 	// * on - Disable the users management view on Webapp
 	WebAppUsersManagement  string `json:"webapp_users_management" enums:"on,off" default:"on"`
 	IntercomUserHmacDigest string `json:"intercom_hmac_digest"`
+	// Pending organization invitations for this user. When non-empty, the user can migrate
+	// to one of these organizations. Only populated in multi-tenant environments.
+	PendingOrgInvitations []PendingOrgInvitation `json:"pending_org_invitations,omitempty"`
```
matheusfrancisco, line 132: just trying to understand something

The user logged in to the hoop, and then he got an invite 
So there is the user -> org-id = 1 and the user-> org-id that invited him, right? 

I know this is a self-hosted instance, and getting the pendingOrgInvitations for the org that invited him and the org that he created, in the self-hosted, this looks okay.

What will happen in our multitenant environment when we get and return every pendingOrgInvite across all instances, since the database is the same?  I might have got it wrong.



Could we try to get it if there is a user with the same email, only with the invite?

rogefm: You got the scenario right — two rows, one active in the auto-created org and one invited in the right org.

On multi-tenant: ListUsersByEmail does scan the whole table, but the loop filters to status == 'invited' && org_id != current, so we only surface pending invites in other orgs, never active memberships from other tenants.

The case you're worried about (same email invited by multiple unrelated tenants) would show all of them. Not a security issue since migration requires an explicit click on a dialog that says the current org will be removed, but it is cross-tenant noise. Happy to tighten as a follow-up if you want.

matheusfrancisco (approved):

rogefm:
@sandromello could you review this again?

sandromello:
✅ Build Completed with Success, Version=1392.0.0-846edfc

sandromello (approved):

PR #1393, opened 2026-04-22
feat(gateway,agent): support SPIFFE JWT-SVID authentication for agents

matheusfrancisco:
## 📝 Description

Adds opt-in SPIFFE JWT-SVID authentication as an alternative to the static
`HOOP_KEY` token for agents. Agents can now authenticate with short-lived,
rotatable JWT-SVIDs issued by the customer's SPIRE (or any SPIFFE-compatible)
infrastructure. Static tokens continue to work in parallel.
### How to run locally
```
# 1. ONCE — mints keys, JWT, and adds HOOP_SPIFFE_* to .env
make run-dev-spiffe-prep

# 2. In terminal 1 — starts the gateway with SPIFFE validation enabled
make run-dev

# 3. In terminal 2 — seeds the DB mapping and starts the host agent
make run-dev-spiffe-agent     # or: make run-dev-spiffe (alias)
```

## Scope
This is the MVP. Implemented:
- **JWT-SVID validation on the gateway** at gRPC auth time, gated behind
  `HOOP_SPIFFE_MODE` (`disabled` / `observe` / `enforce`).
- **SPIFFE-ID → Hoop-agent mapping** via a new admin-managed
  `private.agent_spiffe_mappings` table. Supports exact SPIFFE-ID match and
  prefix+template match (`{{.WorkloadIdentifier}}`, `{{.SPIFFEID}}`).
- **Trust bundle management** via `HOOP_SPIFFE_BUNDLE_URL` (JWKS endpoint)
  or `HOOP_SPIFFE_BUNDLE_FILE`, refreshed on `HOOP_SPIFFE_REFRESH_PERIOD`
  (default `5m`).
- **Agent-side `HOOP_KEY_FILE`** so a sidecar (e.g. `spiffe-helper`) can
  rotate the SVID on disk; the agent re-reads the file before every
  reconnect.
- **HTTP admin API**: `GET/POST/PUT/DELETE /api/spiffe-mappings`.
- **CLI**: `hoop admin create spiffe-mapping`, `get spiffe-mappings`,
  `delete spiffe-mapping <id>`.
- **Audit enrichment**: `spiffe_id`, `trust_domain`, `audience`, `svid_type`,
  `svid_expires_at`, `agent_id`, `agent_name` logged on successful SPIFFE
  auth.
- Docs page at `documentation/setup/deployment/spiffe.mdx`.

Explicitly **out of scope** for this PR (and documented as such):
- X.509-SVID / direct mTLS termination (terminate mTLS at an upstream
[...]

PR #1398, opened 2026-04-23
feat(gateway): stable per-user connection credentials

racerxdl:
## Summary

Make the connection credential token stable across issuances for a given `(user, connection)` pair. Previously every call to `POST /connections/{name}/credentials` generated a fresh random secret key, forcing the user to update their local client config each time. Now the plaintext key is stored encrypted at rest and returned again whenever the user requests credentials for the same connection — only the expiration is refreshed.

This PR also splits the "disconnect" and "revoke" semantics, which were previously conflated in the UI: disconnect ends the session but keeps the token; revoke invalidates the token.

## Behavior matrix

| Trigger | Token |
|---|---|
| First `POST /credentials` | new |
| Repeated `POST /credentials`, no revoke | **same** |
| Repeated `POST /credentials` after expire | **same** (expire_at refreshed) |
| UI "Disconnect" then `POST /credentials` | **same** |
| `POST /credentials/{id}/revoke` then `POST /credentials` | new |
| `POST /credentials/{sid}` (review/JIT resume) | rotated (unchanged from dev) |

## Disconnect vs. revoke

- **`POST /connections/:nameOrID/credentials/:ID/close`** (new) — ends the audit session and tears down active proxy connections but **preserves** the credential row (and its `encrypted_secret_key`). The next credential request returns the same token. Used by the UI "Disconnect" action.
- **`POST /connections/:nameOrID/credentials/:ID/revoke`** (existing, semantics preserved) — marks the credential as revoked (`revoked_at`) and tears down active proxy connections. Excluded from future stable-key lookups so the next credential request issues a brand-new token. Reserved for an explicit "Revoke credential" action.

### Migration `000081_stable_connection_credentials`
- `connection_credentials.encrypted_secret_key BYTEA` — AES-256-GCM ciphertext of the plaintext key
- `connection_credentials.revoked_at TIMESTAMP` — keeps revoked rows for forensics
- `serverconfig.credential_encryption_key TEXT` — base64 server-wide symmetric key
- Partial index `idx_conn_cred_active_user_conn` on `(org_id, user_subject, connection_name) WHERE revoked_at IS NULL`

Migration number 000081 chosen to sit above all currently-claimed numbers on `main` and in-flight branches: `000077_api_keys`, `000078_session_correlation_id` (#1399), `000079_agent_spiffe_mappings` (#1393, SPIFFE), and `000080` which is in use by two unmerged branches (`feat/auth-mcp` for mcp_auth_config and `feature-experimental` for org_feature_flags).

### Crypto (`gateway/models/credential_crypto.go`)
AES-256-GCM. The server-wide key is lazily generated on first use and persisted via `private.serverconfig` (mirrors the existing `shared_signing_key` pattern — no new env var).

### Handler changes (`gateway/api/connections/connection_credentials.go`)
- **`CreateConnectionCredentials`** now delegates to `issueOrRefreshCredential`:
  - Looks up the active credential for `(org, user, connection)` via `GetActiveCredentialByUserAndConnection` (`ORDER BY created_at DESC` picks the latest non-revoked row).
  - If found with ciphertext → decrypt stored ciphertext → return same plaintext → refresh `expire_at` / `session_id` → close the previous audit session.
  - If found without ciphertext (rows created before this migration) → rotate the key in place and backfill the ciphertext (one-time token change on first post-deploy reuse).
  - If not found → generate a fresh key and store both the hash and the encrypted copy.
- **New `CloseConnectionCredentials`** handler — closes audit session, clears `session_id` on the credential row via new `models.ClearCredentialSession`, and tears down active proxy sessions. Does NOT set `revoked_at`.
- **`RevokeConnectionCredentials`** — refactored to share `loadCredentialForMutation` and `terminateActiveCredentialSessions` helpers with Close. Now uses `revoked_at` and pushes `expire_at` into the past so any proxy still reading the legacy column rejects new connections immediately.
- **`ResumeConnectionCredentials`** (review/JIT flow) keeps its rotation-on-call semantics, but now persists the encrypted copy on first issuance and clears it on every rotation.

### Model changes (`gateway/models/connection_credentials.go`)
- New fields: `EncryptedSecretKey []byte`, `RevokedAt *time.Time`.
- New: `GetActiveCredentialByUserAndConnection`, `RefreshCredentialExpiration`, `UpdateConnectionCredentialsSecret`, `ClearCredentialSession`.
- `GetValidConnectionCredentialsBySecretKey` now also filters `revoked_at IS NULL` for defense in depth.
- `CloseExpiredCredentialSessions` now only clears `session_id` (keeps the row for stable-key reuse); the session status transition is unchanged.

### WebApp (`webapp/src/webapp/connections/native_client_access/`)
- New event `:native-client-access->disconnect-credential` calls `/close`.
- Existing `:native-client-access->revoke-credential` kept for future explicit "Revoke credential" UI action (still calls `/revoke`).
- `disconnect-session` in `main.cljs` now dispatches the disconnect event.

### Route registration
- `POST /connections/:nameOrID/credentials/:ID/close` added next to the existing revoke route in `gateway/api/server.go`.

## Verification

- `go build ./...` clean in all five modules (agent, client, common, gateway, libhoop)
- `go vet ./...` clean
- `go test ./gateway/api/connections/... ./gateway/models/...` passes
- Manually tested: create → disconnect → create returns the same token; create → revoke → create returns a fresh token.

## Backward compatibility

- Pre-migration rows exist with `encrypted_secret_key = NULL`. First reuse after deploy rotates the key once per row, then stabilises forever.
- `GetValidConnectionCredentialsBySecretKey` filters revoked rows — revoked credentials are no longer accepted by proxies even if the proxy hasn't called `RevokeByCredentialID` yet.

## Out of scope

- Duplicate credential rows from legacy data (same `(user, connection)` with multiple non-revoked entries). The new lookup tolerates duplicates via `ORDER BY created_at DESC` but does not clean them up.
- Adding an explicit "Revoke credential" UI action. The event is wired but no button dispatches it yet.

Automated by MisterMal

 🤖 Generated with Mister Maluco

Co-Authored-By: MisterMal <teskeslab@lucasteske.dev>

### DIFF
--- a/rootfs/app/migrations/000081_stable_connection_credentials.down.sql
+++ b/rootfs/app/migrations/000081_stable_connection_credentials.down.sql
@@ -1,0 +1,14 @@
+BEGIN;
+
+SET search_path TO private;
+
+DROP INDEX IF EXISTS idx_conn_cred_active_user_conn;
+
+ALTER TABLE connection_credentials
+    DROP COLUMN IF EXISTS encrypted_secret_key,
+    DROP COLUMN IF EXISTS revoked_at;
+
+ALTER TABLE serverconfig
+    DROP COLUMN IF EXISTS credential_encryption_key;
+
+COMMIT;
--- a/gateway/api/connections/connection_credentials.go
+++ b/gateway/api/connections/connection_credentials.go
@@ -143,32 +143,16 @@ func CreateConnectionCredentials(c *gin.Context) {
 		return
 	}
 
-	connType := proto.ConnectionType(conn.SubType.String)
-	secretKey, secretKeyHash, err := generateSecretKey(connType)
-	if err != nil {
-		log.Warnf("failed to create access credentials, err=%v", err)
-		c.AbortWithStatusJSON(400, gin.H{"message": err.Error()})
-		return
-	}
-
 	expireAt := time.Now().UTC().Add(time.Duration(req.AccessDurationSec) * time.Second)
 	if expireAt.After(time.Now().UTC().Add(48 * time.Hour)) {
 		c.AbortWithStatusJSON(400, gin.H{"message": "access duration cannot exceed 48 hours"})
 		return
 	}
 
-	db, err := models.CreateConnectionCredentials(&models.ConnectionCredentials{
-		ID:             uuid.NewString(),
-		OrgID:          ctx.OrgID,
-		UserSubject:    ctx.UserID,
-		ConnectionName: conn.Name,
-		ConnectionType: proto.ToConnectionType(conn.Type, conn.SubType.String).String(),
-		SecretKeyHash:  secretKeyHash,
-		SessionID:      sid,
-		CreatedAt:      time.Now().UTC(),
-		ExpireAt:       expireAt,
-	})
+	connType := proto.ConnectionType(conn.SubType.String)
+	secretKey, db, err := issueOrRefreshCredential(ctx.OrgID, ctx.UserID, conn, connType, sid, expireAt)
 	if err != nil {
+		log.Warnf("failed to issue credential, err=%v", err)
 		c.AbortWithStatusJSON(500, gin.H{"message": err.Error()})
 		return
 	}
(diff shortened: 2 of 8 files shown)


sandromello:
✅ Build Completed with Success, Version=1398.0.0-da20f39

rootfs/app/migrations/000081_stable_connection_credentials.down.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
```
matheusfrancisco, line 1: this migration already exist on main

matheusfrancisco (changes requested):

sandromello:
✅ Build Completed with Success, Version=1398.0.0-92489fc

sandromello:
✅ Build Completed with Success, Version=1398.0.0-0a98887

sandromello:
✅ Build Completed with Success, Version=1398.0.0-e886aec

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1398.0.0-7feed83

PR #1431, opened 2026-05-06
fix: hoop create connection not firing

p3rotto:
> [!IMPORTANT]

## 📝 Description

The `hoop-create-connection` analytics event was only fired by the `POST /api/connections` route via the `TrackRequest` middleware. Connections born from any other code path were silently invisible to Segment/Intercom, breaking the user-activation funnel — most notably for OSS users, who create connections via `POST /api/resources` from the webapp.

This PR makes every connection-creation path emit the event, with a new `source` property so dashboards can distinguish user-initiated from system-initiated events.

## 🔗 Related Issue

Fixes ENG-366

## 🚀 Type of Change

## 📋 Changes Made

Emission is now wired on all five connection-creation paths, each tagged with a `source` property:

| Path | File | `source` | Attribution |
|------|------|----------|-------------|
| `POST /api/connections` | `gateway/api/middleware.go` | `connections-api` | real user (existing middleware, only `source` added) |
| `POST /api/resources` | `gateway/api/resources/resources.go` | `resources-api` | real user (inline) |
| MCP `connections_create` tool | `gateway/api/mcpserver/tools_connections.go` | `mcp` | real user (inline) |
| AWS RDS provisioner | `gateway/api/integrations/aws/provisioner.go` | `aws-rds-provisioner` | system (`UserId="Gateway"` via `TrackEvent`) |
| Agent gRPC `PreConnect` autoregister | `gateway/transport/connectionrequests/connectionsync.go` | `agent-autoregister` | system (`UserId="Gateway"` via `TrackEvent`) |

Implementation notes:
- Inline emission (rather than a route-level middleware) on `/api/resources` to avoid over-counting when a resource is created with zero roles, and to avoid firing on 4xx/5xx.
- AWS provisioner: existence check moved **inside the same transaction** as the upsert so concurrent writers cannot cause double-fire or missed events.
- Agent autoregister: only emits when `conn == nil` at the pre-fetch step, so re-syncs of existing connections don't re-fire.
- All emissions use the same `analytics.New()` / `Close()` / `Track`-or-`TrackEvent` pattern as the existing middleware. They are silent no-ops when `SEGMENT_API_KEY` is unset (OSS self-host) or when `appconfig.AnalyticsTracking()` is `false`.

### Test Configuration:
- **OS**: macOS

### Tests performed:
- [x] `go build ./gateway/...` clean
- [x] `go vet ./gateway/api/mcpserver/... ./gateway/api/integrations/aws/... ./gateway/transport/connectionrequests/... ./gateway/api/resources/...` clean
- [ ] Manual verification post-deploy: confirm Segment receives `hoop-create-connection` from each `source`, and the user-activation funnel recovers.

## ✅ Checklist

## 📄 Additional Notes

- The `source` property creates a small backward-compat consideration: existing dashboards will see the new `source` key on events going forward but did not see it on prior events. Funnel queries that don't reference `source` are unaffected.
- System-fired events (`source` ∈ {`aws-rds-provisioner`, `agent-autoregister`}) use `UserId="Gateway"` and will appear in funnel queries unless filtered. If that's undesirable, downstream queries can filter on `source` or `userId != "Gateway"`.
- Out of scope: adding `AuditMiddleware` to the `/resources` routes (currently absent on POST/PUT/DELETE) — separate concern from this funnel bug.

### DIFF
--- a/gateway/analytics/segment.go
+++ b/gateway/analytics/segment.go
@@ -323,3 +323,68 @@ func (s *Segment) TrackSessionUsageData(eventName string, orgID string, userID s
 
 	s.Track(userID, eventName, props)
 }
+
+// CreateConnectionEvent is the canonical input for the hoop-create-connection
+// analytics event. Different connection-creation paths fill different subsets of
+// these fields — empty/zero fields are omitted from the emitted properties.
+type CreateConnectionEvent struct {
+	// Always required
+	OrgID   string
+	Source  string // e.g. "connections-api", "resources-api", "mcp", "aws-rds-provisioner", "agent-autoregister"
+	Type    string
+	SubType string
+	Command []string // first token used as the "command" property; empty -> ""
+
+	// User-attributed paths set UserID. System paths leave it empty and the
+	// event is emitted via TrackEvent (UserId="Gateway") instead.
+	UserID      string
+	LicenseType string
+
+	// System / agent-bound metadata. Optional.
+	AgentID   string
+	ManagedBy string
+
+	// HTTP-context metadata. APIHostname acts as the gating signal — if set,
+	// the HTTP triple (content-length, user-agent, api-hostname) is included.
+	ContentLength int64
+	UserAgent     string
+	APIHostname   string
+}
+
+// TrackCreateConnection emits hoop-create-connection with a consistent property
+// shape across every connection-creation path. If evt.UserID is empty the event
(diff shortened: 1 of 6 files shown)


sandromello:
✅ Build Completed with Success, Version=1431.0.0-fe43b03

EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1431.0.0-b6acf76

PR #1441, opened 2026-05-12
feat(agent+gateway): async SSH dispatch + upstream error UX + comprehensive tests [ENG-395, ENG-392, ENG-391, ENG-397]

racerxdl:
## Summary

Fix the customer-reported SSH hangs (parallel sessions blocking each other / big transfers stalling) by gating async SSH packet dispatch behind a new feature flag, then complete the UX so failed connections actually tell the user what went wrong instead of hanging on a useless "Connection closed by remote host" line.

| Change | Behavior |
|---|---|
| Feature flag `experimental.agent_async_ssh` | Default false. When enabled, SSHConnectionWrite + SessionClose dispatch to per-packet goroutines so the recv loop isn't held up behind a slow libhoop write. |
| Per-session RW lock | Handlers take RLock; SessionClose takes Lock and drains in-flight handlers. Late packets see closed=true and return cleanly. |
| Per-connection write mutex | Serializes consecutive writes for the same (sid, connID) so async handlers cannot reorder bytes to libhoop. |
| singleflight on proxy creation | Concurrent first-packets for the same (sid, connID) result in exactly one upstream dial. |
| Gateway: user-terminal error messages | When upstream dial fails (wrong IP, refused, timeout, auth, etc.), the gateway writes a sanitized one-liner to the user's SSH channel stderr and tears down promptly. No more silent `Connection closed by remote host`. |

## Why

`a002d3f4` (Apr) made all agent packet processing synchronous. When libhoop's writerQueueCh fills (100 messages — happens during big SCP transfers against a slow remote sshd), agent's serverWriter.Write blocks the recv loop. Every other session on the agent stalls. Customer reports: parallel-session freezes and SCP transfer hangs.

This PR introduces the minimal concurrency that fixes the SSH symptom while preserving the synchronous behavior `a002d3f4` was protecting for other protocols. Default-off means a fresh build behaves exactly like main. Enabling the flag per-org gives the customer the fix without risking other orgs.

The gateway-side UX piece is the natural companion: with the async fix the agent no longer hangs everyone, but a failing connection still gave the user no actionable information. Now they see exactly why it failed.

### Fix-validation tests

| Test | What it proves |
|---|---|
| `TestSSH_BaseSmoke_FlagOff` / `_FlagOn` | Regression contracts for both dispatch modes |
| `TestSSH_ConcurrentFirstPackets_SameConnID_FlagOn` | Singleflight dedups 50 goroutines → 1 upstream SSH connection |
| `TestSSH_SessionClose_CleanShutdown` / `TestSSH_SessionCloseRace_FlagOn` | Session RW lock drains in-flight handlers; clean teardown, no leak |

### Customer-symptom regression test

| Test | What it proves |
|---|---|
| `TestSSH_ParallelSessions_OneStalled_FlagOff` | **Customer bug reproduced**: session B blocked behind session A's stalled Write (verified ≥5s stall) |
| `TestSSH_ParallelSessions_OneStalled_FlagOn` | **Fix verified**: session B established in ~42ms despite session A stalled |

### Base SSH harness

Driven through the new `PipedSSHClient` bridge — a real `x/crypto/ssh.Client` whose transport is bridged to MockTransport via a sshtypes envelope translator.

| Test | What it proves |
|---|---|
| `TestSSH_PasswordAuth_Connect` | Password auth end-to-end |
| `TestSSH_PubkeyAuth_Connect` | RSA pubkey auth end-to-end |
| `TestSSH_BadCredentials` | Auth failure surfaces as SessionClose with auth-error body |
| `TestSSH_ExecCommand` | `echo hello` over exec channel |
| `TestSSH_ExitStatus_Forwarding` | `false` command's exit code 1 propagates through ServerSSHRequest path |
| `TestSSH_InteractiveShell_PTY` | PTY allocation + interactive shell I/O |
| `TestSSH_MultipleChannels` | Two sequential exec channels on the same SSH connection |
| `TestSSH_DirectTCPIP_PortForward` | Port-forward channel to a sidecar nginx, HTTP GET round-trip |

### Gateway upstream-error UX

| Test | What it proves |
|---|---|
| `TestTranslateUpstreamError` (12 sub-cases) | Every libhoop / agent error pattern maps to the right sanitized line; unknown input falls back to a generic message; user-initiated disconnects produce no message |
| `TestNotifyOpenChannels_WritesToEveryChannel` | Multi-channel stderr write + channel close (the latter is what unblocks the gateway's per-channel read goroutine) |
| `TestNotifyOpenChannels_EmptyMessage` | No writes when there's nothing to say |
| `TestNotifyOpenChannels_StderrErrorIgnored` | Best-effort during shutdown — a failing write doesn't panic or block |

### Testutil additions

- `ssh_client.go` (~340 LOC) — `PipedSSHClient` bridge: stands up an `ssh.ServerConn` on TCP loopback, the test's SSH client connects to it, the bridge translates SSH events to/from sshtypes packets. Uses real loopback rather than net.Pipe because net.Pipe deadlocks SSH's exchangeVersions step.
- `ssh_keys.go` — RSA-2048 keypair generation + helpers for both container `PUBLIC_KEY` env and agent `AUTHORIZED_SERVER_KEYS` env.
- `ssh_container.go` — `StartSSHWithPublicKey` for pubkey tests; `StartSSHWithForwarding` patches sshd_config to enable `AllowTcpForwarding` (off by default in linuxserver); `StartHTTPTarget` runs sidecar nginx for port-forward target.
- `slow_tcp_proxy.go` — bidirectional TCP forwarder with controllable Pause/Resume on the agent→upstream direction. Powers the customer-symptom test.

All 15 SSH tests + existing PG tests + new sshproxy unit tests pass locally (`make test-integration` and `make test-oss`).

## libhoop dependency

The customer-symptom stall test exposed two pre-existing concurrent-Write races in libhoop's SSH proxy (close-of-closed-channel + send-on-closed-channel) that fire at agent shutdown when many writes are in flight. **These are not introduced by this PR** — they were latent in production, surfacing as random panics during agent shutdown after SCP transfers.

A sibling libhoop fix wraps the close in sync.Once and adds a defer-recover on the send branch. CI checks out libhoop's main branch which already includes this fix; integration tests are green.

## Removed code

- `Agent.connMtx` and `Agent.connMtxStoreMtx` — dead since `a002d3f4`.
- `Agent.closedSessions` sync.Map — replaced by `sessionStates`, same lifetime semantics plus the added drain guarantee.

## Rollout

1. Merge with flag default-off → no behavior change for any org. The gateway UX improvement (user-terminal error messages) is unconditional and active for all SSH connections.
2. Enable `experimental.agent_async_ssh` for the stuck customer's org. Flag pushes live within seconds (no restart).
3. Customer validates SCP + parallel session scenario. If anything goes wrong, disable the flag and behavior returns to current main.

Automated by MisterMal

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -226,3 +226,13 @@ When merging `main` into a feature branch:
 | Dev run script         | `scripts/dev/run.sh`                                                   |
 | Env sample             | `.env.sample`                                                          |
 | Webapp entry           | `webapp/src/webapp/core.cljs`                                          |
+
+## Coding
+
+NO HACKS. The user is EXTREMELY concerted about code quality much more than immediate results. If they ask you to build something and, while doing so, you hit a wall, and realize that the only way to ship the requested feature is to introduce a local hack, workaround, monkey patch, duct tap - STOP IMMEDIATLY. Either fix the underlying flaw that blocked you ina ROBUST, WELL DESIGNED, PRODUCTION READY manner, or be honest that the prompt can't be completed without hacks.
+
+- DO NOT INTRODUCE HACKS IN THE CODE BASE
+- DO NOT COMMIT CODE THAT COULD BREAK THINGS LATER
+- DO NOT COMMIT PARTIAL SOLUTIONS OR WORKAROUNDS.
+
+The author appreciates honestly and he WILL be glad and thankful if you respond a request with "I couldn't complete your request because the repository lacked support for X". He will be even happier if you go ahead and update the repo to provide necessary support in a well designed and robust way. But he will be VERY ANGRY if, while attempting to implement a feature, you introduce a workaround that will potentially break things latter.
--- a/agent/controller/agent.go
+++ b/agent/controller/agent.go
@@ -15,6 +15,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/getsentry/sentry-go"
@@ -34,9 +35,20 @@ import (
 	pbgateway "github.com/hoophq/hoop/common/proto/gateway"
 	pbsystem "github.com/hoophq/hoop/common/proto/system"
 	"go.mongodb.org/mongo-driver/x/mongo/driver/connstring"
+	"golang.org/x/sync/singleflight"
 )
 
 type (
+	// sessionState carries a per-session RW lock and a closed sentinel.
+	// Packet handlers take RLock for the duration of their work; SessionClose
+	// takes Lock, which drains any in-flight handlers before cleanup runs.
+	// The closed flag, read under RLock, lets handlers drop packets that
+	// arrive after cleanup has begun.
+	sessionState struct {
+		mu     sync.RWMutex
+		closed atomic.Bool
+	}
+
 	Agent struct {
 		client           pb.ClientTransport
 		connStore        memory.Store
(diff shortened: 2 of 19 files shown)


sandromello:
✅ Build Completed with Success, Version=1441.0.0-6dca302

sandromello:
✅ Build Completed with Success, Version=1441.0.0-60c9fce

sandromello:
✅ Build Completed with Success, Version=1441.0.0-9fbe840

sandromello:
✅ Build Completed with Success, Version=1441.0.0-11cabd7

sandromello:
✅ Build Completed with Success, Version=1441.0.0-a84427c

sandromello:
✅ Build Completed with Success, Version=1441.0.0-1ba13d1

matheusfrancisco (approved):

PR #1443, opened 2026-05-13
feat(gateway): surface upstream SSH failures on user terminal

racerxdl:
## Summary

When libhoop's upstream SSH dial fails (wrong IP, refused, timeout, auth failure, OSS agent without SSH support), the user's local `ssh` client today sees only `Connection closed by remote host`. The actual reason lives in the agent → gateway error path but never reaches the user's terminal. This change surfaces a sanitized error line on the user's SSH channel stderr before the connection closes.

| Scenario | Before | After |
|---|---|---|
| Wrong / unreachable IP | `Connection closed by remote host` | `hoop: cannot connect to target server: connection timed out` |
| Refused (port closed / firewall) | `Connection closed by remote host` | `hoop: cannot connect to target server: connection refused` |
| Bad upstream credentials | `Connection closed by remote host` | `hoop: authentication failed against target server` |
| OSS agent without SSH | `Connection closed by remote host` | `hoop: ssh is not enabled on this hoop agent build, contact your administrator` |
| Unknown / unexpected error | `Connection closed by remote host` | `hoop: cannot connect to target server` (generic fallback) |
| User exits ssh normally | `Connection closed by remote host` | `Connection closed by remote host` (no extra noise) |

## How it works

When the gateway-side `sshConnection` ctx is cancelled with a cause, `context.Cause(c.ctx)` carries the libhoop / agent error text. Before sending `pbagent.SessionClose` and closing the underlying SSH conn, the gateway:

1. Calls `translateUpstreamError` on the cause string. Maps libhoop's raw messages onto a fixed vocabulary so we never leak internal addresses, library jargon, or stack traces. Unknown patterns fall back to a generic `cannot connect to target server`.
2. If translation returned a non-empty string, calls `notifyOpenChannels` which writes `hoop: <message>\r\n` to `Stderr()` of every currently-open `ssh.Channel`. The user's `ssh` process prints the line before showing its own "Connection closed" notice.

User-initiated disconnects translate to an empty string so we don't produce noise when the user just exited normally.

## What this deliberately doesn't do

Pre-channel errors (dial failed before any channel was opened) still show only `Connection closed by remote host`. SSH protocol has no graceful way to surface an error before channel acceptance without a wire-protocol change. Acceptable trade-off — a real user running `ssh user@host` always opens a session channel before sending exec/shell requests, so the typical case is covered.

No new packet types. No agent changes. No `libhoop` changes.

## Tests

`usermessage_test.go` covers:

- Every translation branch (refused, timeout, no-route, DNS, auth, session-ready timeout, OSS stub, credential revoked)
- The empty-input no-op
- The user-disconnect suppression
- The unknown-input generic fallback
- `formatRaw` line shape
- `notifyOpenChannels` writes to every channel
- `notifyOpenChannels` skips writes for the empty message even with channels registered
- `notifyOpenChannels` survives `Stderr().Write` returning an error (best-effort during shutdown)

`usermessage_stub_test.go` provides a minimal `ssh.Channel` implementation for the multi-channel tests.

`go test ./gateway/...` and `go vet ./gateway/...` are clean.

## Origin

UX gap surfaced during validation of the ENG-395 async SSH fix. With async dispatch on, the agent no longer hangs the whole gateway when one target is unreachable; the user just sees an opaque close. This change completes the UX story so failing connections actually tell the user what went wrong.

Automated by MisterMal

### DIFF
--- a/gateway/proxyproto/sshproxy/sshproxy.go
+++ b/gateway/proxyproto/sshproxy/sshproxy.go
@@ -383,6 +383,22 @@ func (c *sshConnection) handleConnection() {
 	ctxErr := context.Cause(c.ctx)
 	log.With("sid", c.sid, "conn", c.id).Infof("ssh connection closed, reason=%v", ctxErr)
 
+	// Surface the cancellation cause to the user's terminal before
+	// we tear the SSH connection down. Without this, the user sees
+	// only `Connection closed by remote host` and has to ask an
+	// admin to read the gateway logs to find out what went wrong.
+	//
+	// translateUpstreamError sanitizes the raw libhoop / agent text
+	// into a fixed-vocabulary message so we don't leak internal
+	// addresses, library jargon, or stack traces to end users; it
+	// also returns an empty string for benign causes (e.g. the user
+	// disconnected themselves) so we don't write noise.
+	if ctxErr != nil {
+		if msg := translateUpstreamError(ctxErr.Error()); msg != "" {
+			notifyOpenChannels(&c.sshChannels, msg)
+		}
+	}
+
 	// notify the server that the session is closing
 	err := c.grpcClient.Send(&pb.Packet{
 		Type: pbagent.SessionClose,
--- a/gateway/proxyproto/sshproxy/sshproxy.go
+++ b/gateway/proxyproto/sshproxy/sshproxy.go
@@ -366,20 +366,6 @@ func (c *sshConnection) handleConnection() {
 	// either by the client, server, or context cancellation
 	<-c.ctx.Done()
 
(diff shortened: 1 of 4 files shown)


racerxdl:
Rolled into #1441. ENG-395 and ENG-397 are the same customer story (async fix unblocks the agent, user-terminal message tells them why the target failed) so they belong in one review. Branch will be cleaned up after #1441 merges.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1443.0.0-3330c37

sandromello:
✅ Build Completed with Success, Version=1443.0.0-e6d787d

PR #1454, opened 2026-05-18
fix(httpproxy): remove negative auth cache

racerxdl:
## Problem

The negative auth cache was originally added to avoid hammering the DB when Kubernetes tokens expired and triggered a flood of requests. It cached failed auth attempts for 24h.

However, httpproxy tokens are now **infinite-lived credentials**. When a user's OIDC token (behind the proxy) expires, the proxy gets a 401, catches it as a `credentialError`, and stores the proxy token hash in `negativeAuthCache` for 24h. Even after the user renews their OIDC session, the proxy continues rejecting their requests until the cache TTL expires or the gateway restarts.

## Fix

Remove the negative auth cache entirely. The protection it provided is no longer needed because:
- httpproxy tokens are infinite — they don't expire on a regular basis
- Sessions are cached in `sessionStore` — the DB is only hit on session creation, not every request
- The flood-of-requests scenario it guarded against no longer applies

Fixes the issue where users get permanently locked out of the HTTP proxy after their OIDC token rotates.

cc @chico 

Automated by MisterMal

### DIFF
--- a/gateway/proxyproto/httpproxy/httpproxy.go
+++ b/gateway/proxyproto/httpproxy/httpproxy.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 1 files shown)


EmanuelJr (approved):

matheusfrancisco (approved):

luanlorenzo (approved):

sandromello:
✅ Build Completed with Success, Version=1454.0.0-g0a6ffb9

PR #1465, opened 2026-05-20
feat(cli): add hoop upgrade & hoop versions (in-CLI version manager)

matheusfrancisco:
> [!IMPORTANT]

## 📝 Description

Adds an nvm/fnm-style version manager to the hoop CLI under a single `hoop versions` command group. Two new entry points, sharing one install pipeline:

- **`hoop versions sync`** — installs the version the connected gateway reports (via `/api/serverinfo`), so the CLI and the gateway stay on identical versions. Requires `hoop login`.
- **`hoop versions upgrade`** — installs whatever is published as the latest release at https://releases.hoop.dev/release/latest.txt. Does **not** consult the gateway and does **not** require `hoop login`, so it works fine against a dev gateway or before any config is in place.

Both commands download the matching release tarball for the host OS/arch, verify its SHA-256 against the published checksum file, extract the binary into `$HOME/.hoop/versions/<version>/hoop`, and point `$HOME/.hoop/bin/hoop` at it. On first install the CLI offers to append `export PATH="$HOME/.hoop/bin:$PATH"` to the user's shell rc (zsh / bash / fish supported).

The same `hoop versions` group also keeps the lower-level controls: `list` / `install <ver>` / `use <ver>` / `remove <ver>`.

Side-quest improvements picked up along the way:
- **Version-mismatch warning** — every gateway HTTP response carries `Server: hoopgateway/<ver>`. The CLI watches that header and prints a single one-shot warning when local and remote drift, suggesting `hoop versions sync`. Opt out with `HOOP_DISABLE_VERSION_CHECK=true`.
- **`HumanizeNetError`** — common/httpclient now turns raw `ECONNREFUSED`, DNS, timeout, and TLS errors into multi-line actionable guidance instead of bare `dial tcp ... connection refused`.
- **Multi-line, branch-specific errors** for the three unactionable-gateway cases (dev build returning `unknown` / gateway below the install floor / response is not semver). The "below floor" message explicitly tells the user why the floor exists (`hoop versions sync` was introduced in 1.74.0) so they understand why we won't downgrade their CLI.

The version manager only installs releases **≥ 1.74.0** — older CLIs don't ship `hoop versions`, so installing them would orphan the user.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- **New `client/upgrade/` package** built as single-responsibility modules: directory layout (`paths.go`), platform/release-URL mapping (`platform.go`), TOML-backed install store (`store.go`), downloader + tarball/SHA-256 verifier (`installer.go`), atomic symlink manager (`symlink.go`), shell-rc detector + `PATH` writer (`shellrc.go`), gateway `/api/serverinfo` client (`gateway.go`), latest-release fetcher (`latest.go`), version-floor validator with sentinel errors (`floor.go`), and one-shot mismatch warning (`versioncheck.go`). All modules ship unit tests.
- **`hoop versions sync`** (`client/cmd/versions/sync.go`) — fetches gateway version, validates against the install floor, delegates to the shared install pipeline. Branch-specific error formatter for the three unactionable cases lives here too.
- **`hoop versions upgrade`** (`client/cmd/versions/upgrade.go`) — fetches latest from `releases.hoop.dev/release/latest.txt`, then delegates to the same shared install pipeline. No config or gateway round-trip required.
- **Shared install pipeline** (`client/cmd/versions/install_flow.go`) — `installAndActivate(target)` is what both commands call. Same flags (`-y`, `--skip-path-hint`), same progress lines, same PATH-hint prompt. Identical user experience regardless of which entry point produced the target.
- **`hoop versions list|install|use|remove`** (`client/cmd/versions/versions.go`) — explicit version controls. `install --use` is the install-and-switch shortcut; `install --reinstall` forces a re-download; `remove --force` is required for the active version.
- **Version-mismatch warning plumbing** — `common/httpclient` exposes a `VersionCheckCallback` hook; `client/cmd/root.go` registers `upgrade.WarnOnceFromServerHeader` so every gateway call participates without per-call boilerplate. The warning auto-suppresses for dev CLIs (`version == "unknown"`) and during `hoop versions sync` / `hoop versions upgrade` themselves. Warning text now recommends `hoop versions sync`.
- **Humanized network errors** — `common/httpclient/errors.go` classifies `ECONNREFUSED`, DNS, timeout, and TLS failures. Wired into `FetchGatewayInfo`. The `latest.go` fetcher has its own focused error wrapper (advice differs: it's a public CDN, not a user-controlled gateway).
- **Humanized version errors** — three sentinels (`ErrUnknownGatewayVersion`, `ErrBelowFloor`, `ErrInvalidVersion`) drive branch-specific messages in `sync.go`, including the explanation that `hoop versions sync` was introduced in 1.74.0 so users understand the floor.
- **Dependencies**: `github.com/BurntSushi/toml` (install store) and `golang.org/x/mod/semver` (floor comparison) added to `client/go.mod`. No new runtime deps in any other module.

### Test Configuration:
- **Browser(s)**: N/A (CLI feature)
- **OS**: macOS (Darwin/arm64), with the install path also exercising the cross-platform mapping (`Linux_x86_64`, `Windows_x86_64`, etc.) via unit tests.

### Tests performed:
- [x] Unit tests pass — `go test ./upgrade/... ./cmd/versions/...` from `client/`, and `go test ./common/httpclient/...` from the repo root. All green.
- [x] `go build ./...` succeeds across the client module.
- [x] Integration tests pass — the existing `make test-oss` suite remains green; no plugin or transport-layer behavior was modified.
- [x] Manual smoke testing:
  - `hoop versions --help` lists `sync` and `upgrade` alongside the existing subcommands.
  - `hoop versions sync` against a dev gateway → friendly "did not report a release version (got \"unknown\")" message, and now also points the user at `hoop versions upgrade` as the escape hatch.
  - `hoop versions sync` against a real gateway → download, SHA-256 verify, symlink switch, PATH-prompt all succeed end-to-end; subsequent run reports "Already on \<version\>".
  - `hoop versions upgrade` → fetches latest.txt, validates against the install floor, then runs the same install pipeline.
  - `hoop versions install <ver>` / `use` / `list` / `remove` exercised in sequence to confirm the TOML store and symlink stay in sync, including the `--force` path for removing the active version.
  - Connection-refused / unknown-host / mid-handshake-cut scenarios verified the new `HumanizeNetError` output.
  - Oversized-response, empty-response, below-floor, and 404 cases on `latest.txt` are covered by `latest_test.go`.

N/A — CLI feature. Sample output from the improved error paths is in the manual-testing notes above and reproduced verbatim in the docs page.

## ✅ Checklist

## 📄 Additional Notes

- **Companion docs PR** in `hoophq/documentation`: `clients/cli-upgrade.mdx` is removed and `clients/cli-versions.mdx` is added in its place, with `docs.json` and the install Note in `clients/cli.mdx` updated to match. Page covers both commands, the version-mismatch warning, `HOOP_DISABLE_VERSION_CHECK`, the `~/.hoop/` file layout, and the two install-floor caveats. Includes a "sync vs upgrade" cheat sheet table.
- **Command surface changes vs the previous revision of this PR.** The earlier draft of this branch exposed `hoop upgrade` as a top-level command. That has been consolidated under `hoop versions` so the gateway-sync and latest-track flows live next to each other and share an install pipeline:
  - Old `hoop upgrade` → new `hoop versions sync`.
  - New `hoop versions upgrade` tracks `releases.hoop.dev/release/latest.txt`.
  - The version-mismatch warning, the floor-error message, and the shell rc marker comment all reference `hoop versions sync` instead of `hoop upgrade`.
  - Nothing was released under the old name, so no deprecation alias is needed.
- **No changes to gateway, agent, plugin order, migrations, or env config.** All new code lives under `client/` and `common/httpclient/`; the existing `common/httpclient` callers (gateway, agent) remain unaffected because `VersionCheckCallback` is nil-by-default and only the CLI registers it.
- **Install floor of 1.74.0** is intentional and documented at every surface (error message, code comment in `floor.go`, and the docs page). Bumping it later only requires changing `MinInstallableMinor`.
- Reviewers may want to focus on `client/upgrade/installer.go` (download/verify/extract pipeline), `client/upgrade/symlink.go` (atomic swap + dangling-link handling), and `client/cmd/versions/install_flow.go` (shared post-target flow used by both `sync` and `upgrade`). The new `client/upgrade/latest.go` is small and entirely covered by `latest_test.go`.

---


sandromello:
✅ Build Completed with Success, Version=1465.0.0-g7d69a46

sandromello:
✅ Build Completed with Success, Version=1465.0.0-g5d35505

p3rotto (approved):

PR #1473, opened 2026-05-25
fix(analytics): identify users from webapp to collapse Segment MTU

p3rotto:
## 📝 Description

Fixes the Segment MTU inflation reported in **ENG-407**, where the `hoophq` workspace was showing ~14k tracked users while PostHog/Mixpanel reported only ~6k. Root cause: the browser's Segment SDK assigns a random `anonymous_id` to every visitor, and the webapp was never calling `analytics.identify()` after login, so every browser session counted as a separate MTU and never collapsed into the gateway-identified user.

This PR adds the missing `identify()` call on the React shell (which wraps every authenticated route) so Segment links the browser's `anonymous_id` to the same hashed user-id the gateway already uses server-side. The hashing matches `gateway/analytics/segment.go:getUserIDHash` exactly (SHA-256 of the OIDC subject, lowercase hex), so browser and server identify into the **same** Segment profile.

A small backend fix is bundled in: two `Track`/`Identify` call sites in `gateway/api/login/oidc/login.go` and `gateway/api/user/user.go` were passing `ctx.UserID` (the DB row UUID) instead of `ctx.UserSubject` (the stable OIDC subject), which could cause the same human to register as multiple Segment profiles across user-record re-creations.

## 🔗 Related Issue

ENG-407

## 🚀 Type of Change

## 📋 Changes Made

**Backend (gateway)**
- `gateway/api/login/oidc/login.go` — `Track`/`Identify` now pass `ctx.UserSubject` instead of `ctx.UserID` (DB UUID).
- `gateway/api/user/user.go` — same fix on `Identify` in the user-update path.

**React shell (webapp_v2)**
- New `src/services/analytics.js` — owns the `AnalyticsBrowser` singleton, computes `sha256(user.id)` to match the gateway's hashed user-id, builds the trait set, and resets cached traits before each identify so `localStorage`-persisted PII can't leak into an `anonymous`-mode payload.
- `src/stores/useUserStore.js` — adds `analyticsMode` to state (from `serverInfo.analytics_mode`) and `initAnalytics(user)` mirroring `initIntercom(user)`.
- `src/components/ProtectedRoute.jsx` — calls `initAnalytics(userData)` right after `initIntercom(userData)` when user + serverInfo land.
- `vite.config.js` — injects `SEGMENT_WRITE_KEY` (env var) into the bundle at build time. Same env var name as the CLJS bundle, with the same hardcoded default. No CI/helm/Dockerfile changes needed.
- `package.json` — adds `@segment/analytics-next ^1.64.0` (same version as the CLJS bundle).
- `CLAUDE.md` — documents `SEGMENT_WRITE_KEY`.

**CLJS webapp**
- `src/webapp/events/segment.cljs` — makes `:segment->load` idempotent so the SDK can't be loaded twice (which previously doubled all destination scripts: Intercom, ProfitWell, etc.).

**Per-mode behavior (mirrors the gateway):**

| `analytics_mode` | identify fires? | Traits sent |
|---|---|---|
| `disabled` | no | — |
| `anonymous` | yes | `org-id`, `user-id` (hashed), `is-admin`, `status` |
| `identified` | yes | adds `email`, `name` |

### Test Configuration:
- **Browser(s)**: Chrome, Brave (Brave's tracker blocker amplifies retries — verified the fix doesn't cause runaway request volume)
- **OS**: macOS

### Tests performed:
- [x] Verified the hashed user-id in the browser's identify request payload matches the gateway's `getUserIDHash(ctx.UserSubject)` output.
- [x] Toggled the org's `analytics_mode` between `identified`, `anonymous`, `disabled` and verified the identify payload changes accordingly (no `email`/`name` in `anonymous` mode, no identify at all in `disabled`).
- [x] Confirmed the `anonymousId` field in the identify payload — this is the Segment SDK's auto-generated browser session ID and is the merge instruction Segment needs to collapse anonymous events into the identified user.
- [x] Verified CLJS bundle compiles (`shadow-cljs release hoop-ui` — 0 warnings) and `webapp_v2` builds (`npm run build`) and lints clean for the touched files.

## 📄 Additional Notes

**Why frontend-only `identify()` (no CLJS counterpart)?** Earlier iterations added an identify dispatch on the CLJS side too, but the React shell wraps every route (per `webapp_v2/CONTEXT_MIGRATION.md`), so its identify already covers all authenticated sessions. A second CLJS identify would just be redundant and double the per-page request volume.

**Why the trait-cache reset?** `analytics-next` persists user traits to `localStorage` and **merges** them into every subsequent `identify()` call. Without an explicit reset, `email`/`name` from a previous `identified`-mode session would leak into a later `anonymous`-mode payload. Resetting before each identify guarantees the wire payload matches `buildTraits()` exactly.

**Out of scope, flagged for follow-up:**
- `gateway/api/serverinfo/serverinfo.go` uses a package-level singleton `serverInfoData` that mutates per request — fields like `AnalyticsMode`, `FeatureFlags`, `LicenseInfo` can leak across concurrent org responses.

### DIFF
--- a/gateway/api/login/oidc/login.go
+++ b/gateway/api/login/oidc/login.go
@@ -502,7 +502,7 @@ func (h *handler) analyticsTrack(isNewUser bool, userAgent string, ctx *models.C
 	trackClient := analytics.New()
 	defer trackClient.Close()
 	if !isNewUser {
-		trackClient.Track(ctx.UserID, analytics.EventLogin, map[string]any{
+		trackClient.Track(ctx.UserSubject, analytics.EventLogin, map[string]any{
 			"org-id":       ctx.OrgID,
 			"auth-method":  appconfig.Get().AuthMethod(),
 			"user-agent":   userAgent,
@@ -513,14 +513,14 @@ func (h *handler) analyticsTrack(isNewUser bool, userAgent string, ctx *models.C
 	trackClient.Identify(&types.APIContext{
 		OrgID:          ctx.OrgID,
 		OrgLicenseData: &ctx.OrgLicenseData,
-		UserID:         ctx.UserID,
+		UserID:         ctx.UserSubject,
 		UserEmail:      ctx.UserEmail,
 		UserName:       ctx.UserName,
 	})
 	go func() {
 		// wait some time until the identify call get times to reach to intercom
 		time.Sleep(time.Second * 10)
-		trackClient.Track(ctx.UserID, analytics.EventSignup, map[string]any{
+		trackClient.Track(ctx.UserSubject, analytics.EventSignup, map[string]any{
 			"org-id":       ctx.OrgID,
 			"auth-method":  appconfig.Get().AuthMethod(),
 			"user-agent":   userAgent,
--- a/gateway/api/user/user.go
+++ b/gateway/api/user/user.go
@@ -221,7 +221,7 @@ func Update(c *gin.Context) {
 	trackClient.Identify(&types.APIContext{
 		OrgID:          ctx.OrgID,
 		OrgLicenseData: ctx.OrgLicenseData,
-		UserID:         existingUser.ID,
+		UserID:         existingUser.Subject,
 		UserEmail:      existingUser.Email,
 		UserName:       existingUser.Name,
 		UserStatus:     existingUser.Status,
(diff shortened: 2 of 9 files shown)


sandromello:
✅ Build Completed with Success, Version=1473.0.0-g2297381

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1473.0.0-gaa5a254

rogefm (approved):

PR #1485, opened 2026-05-27
fix(analytics): prevent nil-pointer crash in TrackSessionUsageData

p3rotto:
## 📝 Description

`TrackSessionUsageData` was crashing intermittently with a nil-pointer dereference. The root cause was a data race in the previous `errgroup` block: all goroutines shared a single `err` variable in the enclosing scope, so a late-assigning goroutine could clobber an earlier error and let downstream code consume a `nil` model.

This PR fixes the race and takes the opportunity to clean up the function shape, which had grown awkward (5 stacked `var err error` blocks plus a 6-arg property builder).

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [x] ♻️ Code refactor

## 📋 Changes Made

- Extracted a `sessionUsageData` struct that holds all models the analytics event needs (session, connection, agent, guardrails, data masking, JIT and command access request rules).
- Introduced `loadSessionUsageData(orgID, sessionID)` which owns the parallel fetch orchestration. Each goroutine now writes directly to its own struct field, eliminating the shared-`err` race that produced the nil-pointer crash.
- Collapsed `sessionUsageProperties` from 6 positional args to a single `*sessionUsageData` argument.
- Slimmed `TrackSessionUsageData` down to a handful of lines — orchestration, logging, and `Track` — with no leaked intermediate state.
- Updated unit tests to use a small functional-options builder so each subtest expresses only the field it cares about.

### Test Configuration:
- **OS**: macOS (darwin)

### Tests performed:
- [x] Unit tests pass (`go test ./gateway/analytics/...`)
- [ ] Integration tests pass

## ✅ Checklist

## 📄 Additional Notes

The behavior of the analytics event is unchanged — same properties are emitted, the function still degrades gracefully when the session or related models cannot be loaded. The fix is purely structural.

### DIFF
--- a/gateway/analytics/segment.go
+++ b/gateway/analytics/segment.go
@@ -225,14 +225,87 @@ func (s *Segment) Track(userID, eventName string, properties map[string]any) {
 	}
 }
 
-func sessionUsageProperties(
-	s *models.Session,
-	c *models.Connection,
-	agent *models.Agent,
-	guardrails *models.ConnectionGuardRailRules,
-	dataMasking json.RawMessage,
-	accessRequestList []*models.AccessRequestRule,
-) map[string]any {
+type sessionUsageData struct {
+	session              *models.Session
+	connection           *models.Connection
+	agent                *models.Agent
+	guardrails           *models.ConnectionGuardRailRules
+	dataMasking          json.RawMessage
+	jitAccessRequest     *models.AccessRequestRule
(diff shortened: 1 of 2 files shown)


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1485.0.0-g6bdbd50

PR #1489, opened 2026-05-29
Tunnel: macOS support (utun, LaunchDaemon, /etc/resolver) + dual-stack addressing

racerxdl:
## Summary

Brings the `hsh-tunneld` tunnel daemon to **macOS** (previously Linux-only), and reworks addressing to **dual-stack** so name resolution actually works for real apps on macOS.

Three platform-specific pieces that were Linux-only stubs now have real macOS implementations, plus a dual-stack addressing change that turned out to be required for macOS DNS to work at all.

### macOS platform support
- **netstack (RD-207):** open the stock `utun` device via a `PF_SYSTEM`/`SYSPROTO_CONTROL` kernel-control socket; per-packet `AF_INET`/`AF_INET6` framing handled inside the device so `stack.go` stays platform-neutral. Routes via `ifconfig`/`route`.
- **resolved (RD-208):** host DNS routing for `*.hoop` via `/etc/resolver/<tld>` (mDNSResponder), with crash self-heal (`CleanupStale`) at startup.
- **service (RD-217):** real LaunchDaemon `Manager` (`launchctl bootstrap/bootout/enable/kickstart`, `dscl` group), plist template, race-free + idempotent reload, auto-adds the invoking user to the `hsh` group so `hsh`/the tray run without sudo.
- **ipc fix:** create the socket parent dir at bind time (macOS `/var/run` is volatile across reboots; launchd has no `RuntimeDirectory`), and chown the control-token to the IPC group (it was `root:daemon`, unreadable by `hsh` group members).

### Dual-stack addressing (RD-208)
The original IPv6-only design broke macOS: `getaddrinfo()` honours `AI_ADDRCONFIG` and won't query/return `AAAA` on a host with no global IPv6, so `psql -h foo.hoop` failed even though `dig`/`ping6` worked — and no v6-side workaround flips macOS 26 out of "Request A records". Fix: hand out **both** a ULA IPv6 and a CGNAT (`100.64.0.0/10`) IPv4 address per connection; resolver answers both `A` and `AAAA`; gVisor accepts both families (and dispatches inbound packets by IP version — a v6-only inject was silently dropping v4 SYNs). Negative DNS answers now carry an SOA (RFC 2308) which macOS requires. ADR-0001 amended accordingly.

## Verified on macOS
- `sudo hsh-tunneld install` registers + starts the LaunchDaemon; `hsh tunnel login/status/connections` work without sudo (group ACL).
- `getaddrinfo`/`nc`/`psql -h <name>.hoop` connect through the tunnel over IPv4.
- Builds + tests + `go vet` pass on darwin/linux/windows/freebsd; race detector clean on changed packages.

## Known limitation
`ping` works over IPv6 but not IPv4 (gVisor doesn't answer ICMPv4 echo for on-demand/spoofed addresses). The tunnel is TCP-only, so this is cosmetic — documented in the README + ADR-0001 with guidance to test via `nc`/`psql`.

## Tickets
Closes RD-207, RD-208, RD-217.

Automated by MisterMal


sandromello:
✅ Build Completed with Success, Version=1489.0.0-g695c9a0

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1489.0.0-gd5dc9c9

PR #1490, opened 2026-05-29
fix(guardrails): prevent Enter from submitting form while typing regex pattern

rogefm:
> [!IMPORTANT]

## 📝 Description

Fixes a bug in the guardrails form where pressing Enter while typing a custom regex pattern (Pattern Match rule) would submit the entire form before the pattern value was committed to state, resulting in the pattern being lost and the user being navigated away from the form.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

## 📋 Changes Made

- Added `on-key-down` handler to the pattern regex input that calls `.preventDefault` when Enter is pressed, preventing the form from submitting while the user is typing a regex pattern
- Kept the original `on-blur` + `pattern-state` architecture intact to avoid a regression where updating the rule state on every keystroke caused the input to lose focus after each character (due to the hash-based row key changing on every re-render)

### Test Configuration:
- **OS**: Linux / macOS

### Tests performed:

**Steps to reproduce the original bug:**
1. Open a guardrail (create or edit)
2. Add an Input or Output rule, select type "Pattern Match", then "Create custom rule"
3. Type a regex in the details input and press Enter
4. Observe: form saves immediately with an empty `pattern_regex`

**After fix:**
- Pressing Enter inside the regex input no longer submits the form
- The value is correctly saved when clicking the Save button (blur fires before submit, committing the value)

## ✅ Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_013zL5LzxN7qsPK3AbZGtTRm)_

### DIFF
--- a/webapp/src/webapp/guardrails/rules_table.cljs
+++ b/webapp/src/webapp/guardrails/rules_table.cljs
@@ -67,10 +67,8 @@
        :full-width? true
        :size "3"
        :not-margin-bottom? true
-       :on-change (fn [e]
-                    (let [value (-> e .-target .-value)]
-                      (on-pattern-change pattern-state idx value)
-                      (on-rule-field-change state idx :pattern_regex value)))
+       :on-change #(on-pattern-change pattern-state idx (-> % .-target .-value))
+       :on-blur #(on-rule-field-change state idx :pattern_regex (get @pattern-state idx ""))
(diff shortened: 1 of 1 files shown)


sandromello:
✅ Build Completed with Success, Version=1490.0.0-g0c82e88

PR #1491, opened 2026-05-29
feat: enhance rules table with Enter key handling for pattern input

rogefm:
## 📝 Description

Fixes a bug in the guardrails form where pressing Enter while typing a custom regex pattern (Pattern Match rule) would submit the entire form before the pattern value was committed to state, resulting in the pattern being lost and the user being navigated away from the form.

## 🔗 Related Issue

Fixes ENG-417

## 🚀 Type of Change

## 📋 Changes Made

- Added `on-key-down` handler to the pattern regex input that calls `.preventDefault` when Enter is pressed, preventing the form from submitting while the user is typing a regex pattern

### Test Configuration:
- **OS**: Linux / macOS

### Tests performed:

**Steps to reproduce the original bug:**
1. Open a guardrail (create or edit)
2. Add an Input or Output rule, select type "Pattern Match", then "Create custom rule"
3. Type a regex in the details input and press Enter
4. Observe: form saves immediately with an empty `pattern_regex`

**After fix:**
- Pressing Enter inside the regex input no longer submits the form
- The value is correctly saved when clicking the Save button (blur fires before submit, committing the value)

## ✅ Checklist

### DIFF
--- a/webapp/src/webapp/guardrails/rules_table.cljs
+++ b/webapp/src/webapp/guardrails/rules_table.cljs
@@ -69,6 +69,9 @@
        :not-margin-bottom? true
        :on-change #(on-pattern-change pattern-state idx (-> % .-target .-value))
        :on-blur #(on-rule-field-change state idx :pattern_regex (get @pattern-state idx ""))
+       :on-keyDown (fn [e]
+                     (when (= (.-key e) "Enter")
+                       (.preventDefault e)))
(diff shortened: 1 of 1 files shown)


luanlorenzo (approved):

sandromello:
✅ Build Completed with Success, Version=1491.0.0-g2837865

PR #1500, opened 2026-06-02
build(test): re-enable -race in make test-integration [ENG-396]

racerxdl:
## What

Re-enables `-race` in `make test-integration`, which was removed in #1439 to unblock CI while the underlying race was investigated.

## Root cause (fixed upstream)

libhoop's `blockingReader` in all four DB proxies (postgres, mysql, mssql, mongodb) had a real data race: `Write` and `Read` both accessed `bytes.Buffer` without holding `cond.L`. Only the `Wait()` call held the lock. This is now fixed in hoophq/libhoop#75.

## This PR

One-line Makefile change: adds `-race` back to the `test-integration` target.

Depends on hoophq/libhoop#75 being merged and picked up before CI runs.

Fixes ENG-396.

Automated by MisterMal


sandromello:
✅ Build Completed with Success, Version=1500.0.0-ga920741

sandromello:
✅ Build Completed with Success, Version=1500.0.0-g809fa3f

EmanuelJr (approved):

racerxdl:
**Root cause of the `-race` failures identified and fixed.**

The failing tests were `TestSSH_InteractiveShell_PTY` and `TestSSH_DirectTCPIP_PortForward`, both in `agent/integration/ssh_base_test.go`.

**Race report:**
```
WARNING: DATA RACE
Write at runtime.closechan()  ← MockTransport.Close() (transport.go:71)
  via shutdownAgent() deferred in each test

concurrent with:
Write at runtime.chansend1()  ← MockTransport.Send() (transport.go:40)
  via agent.processSSHProtocol → a.client.Send()
```

**Root cause:** `MockTransport.Close()` was calling `close(sendCh)` and `close(recvCh)` while `processSSHProtocol` could still be in-flight (the SSH handler sends back a session-close packet after `shutdownAgent()`'s deferred call fires).

**Fix (commit d0c5e0a8):** Remove the `close()` calls from `MockTransport.Close()`. Cancel the context instead — all protocol pumps and the agent's recv loop already `select` on `ctx.Done()` so they stop cleanly. Channels are GC'd when the transport goes out of scope.

This is the same fix that was already applied on the `teske/eng-389-mssql-mongodb-integration-tests` branch (#1514), now ported here so PR #1500 can go green independently.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1500.0.0-gd0c5e0a

PR #1501, opened 2026-06-02
test(agent): MySQL integration suite [ENG-387, ENG-388]

racerxdl:
## What

Adds an end-to-end MySQL integration test suite, mirroring the existing Postgres and SSH suites. Drives the real `controller.Agent` against a MariaDB testcontainer through libhoop's MySQL proxy.

## Approach: bridged real driver

libhoop's MySQL proxy is **not** a transparent byte pump like the Postgres one — it's a man-in-the-middle that speaks the full handshake with the client (forwards the upstream greeting, reads the client handshake response, then auths upstream itself). Hand-rolling that handshake would be hundreds of lines of fragile protocol code.

So the tests drive a **real `go-sql-driver/mysql` connection over a `net.Pipe` bridge** (`PipedMySQLClient`) whose bytes are shuttled to/from the agent as `MySQLConnectionWrite` packets. The driver handles greeting parsing, auth scramble, and result-set decoding; the bridge only moves bytes. This is the same real-client-over-a-bridge pattern as the SSH suite's `PipedSSHClient`.

A zero-length trigger packet bootstraps the upstream dial (MySQL servers speak first, but libhoop only dials on the first client packet).

## Coverage

**Base (`mysql_test.go`):** ping, simple query, full CRUD, multi-row result set, tx commit/rollback, bad-query error survival, bad-credentials rejection.

**Concurrency (`mysql_concurrency_test.go`):** parallel independent sessions, SessionClose-vs-handler teardown race. `ParallelSessions_OneHangs` is skipped — MySQL packets are dispatched synchronously (the async flag is SSH-only), so a slow upstream dial blocks the recv loop. Unskip when async dispatch covers MySQL.

## Test-infra fix

`MockTransport.Close` no longer closes `sendCh`/`recvCh`: long-lived bridge pumps and the demux still send on them and exit via `ctx.Done()`; closing underneath them is itself a data race (surfaced under `-race` on SSH). `Inject` now also unblocks on `ctx.Done()`.

## Dependencies

- **libhoop#75** — proxy atomics fix (`initialized`/`closed`/`pid`). Required for the suite to run clean under `-race`; these tests surfaced those races.
- **#1500** — re-enables `-race` in `make test-integration`.

With all three, the full suite (postgres + mysql + ssh) is green under `-race`.

Closes ENG-387, ENG-388.

Automated by MisterMal

### DIFF
--- a/agent/integration/mysql_concurrency_test.go
+++ b/agent/integration/mysql_concurrency_test.go
@@ -1,0 +1,151 @@
+//go:build integration
+
+package integration
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/hoophq/hoop/agent/integration/testutil"
+)
+
+// TestMySQL_ParallelSessions runs two independent MySQL sessions on the
+// same agent concurrently and asserts both complete their work. Each
+// session has its own bridged connection, its own (sid, connID), and its
+// own upstream MariaDB backend.
+//
+// The invariant: independent sessions on the same agent must not corrupt
+// each other's state. With the blockingReader race fixed (ENG-396) this
+// runs clean under -race; before the fix, concurrent Read/Write on each
+// proxy's bytes.Buffer would trip the detector.
+func TestMySQL_ParallelSessions(t *testing.T) {
+	mc := testutil.StartMySQL(t)
+	agent, tr := startAgent(t)
+	defer shutdownAgent(t, agent, tr)
+
+	const numSessions = 4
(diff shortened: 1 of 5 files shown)


sandromello:
✅ Build Completed with Success, Version=1501.0.0-g909ab79

EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1501.0.0-g061df2a

PR #1515, opened 2026-06-09
test(gateway): Level-1 API smoke test suite [ENG-390]

racerxdl:
## What

Adds a **Level-1 gateway API smoke test suite** under `gateway/integration`. It is the API-level counterpart to the wire-protocol suites in `agent/integration`: where those exercise the agent + libhoop proxies against real databases, this exercises the **gateway HTTP API, auth, RBAC, and the migration/model layer** — the half of the system the protocol suites can't reach.

## How it works

`TestMain` boots a throwaway `postgres:17.6` via testcontainers, runs the full migration set, bootstraps the default organization the way `gateway/main.go` does (minus plugins, proxies, and the gRPC server, none of which these tests exercise), and serves the gateway's **complete gin route tree** through an in-process `httptest` server. No agent, no gRPC server, no separate binary — fast (~3-6s after the Postgres image is cached) and debuggable.

## Coverage (12 tests)

| Area | What it asserts |
|------|-----------------|
| Public server info | unauthenticated `GET /publicserverinfo` |
| Registration | first user gets admin; second self-registration locked out (403) |
| Login | success returns JWT; wrong password → 401 |
| Server info | 401 without token, version payload with token |
| User listing | read-only access returns the registered admin |
| Connection CRUD | create → list → get → delete → 404 |
| Agent CRUD | create returns a gRPC DSN; list; delete |
| Feature flags | catalog listing is non-empty |
| hpk_ API key | create → authenticate with it → revoke → rejected |
| RBAC | auditor-scoped key denied on an admin-only endpoint (403) |
| Healthz | degraded contract (400 + `liveness=ERR`) with no gRPC |

## Production-faithfulness

To avoid false confidence, the test harness does **not** hand-roll a router. It calls a newly extracted `(*Api).BuildEngine()` (in `gateway/api/server.go`) which builds the exact gin engine `StartAPI` serves — security/CORS middleware, the `/api` group with Sentry middleware, the full route tree, and `openapi.RegisterGinValidators()`. `StartAPI` now calls `BuildEngine` and only owns the listener. Tests therefore exercise the real middleware chain and request validators. `TestMain` also warms the feature-flag and analytics-mode caches like `main.go`.

## CI / build

- New `make test-gateway` target: `CGO_ENABLED=0`, `libhoop-map` only — **no `generate-wasm`** (the committed `rdp_parser.wasm` is `go:embed`'d, so no Rust toolchain is needed).
- New `gateway-smoke-test` job in `pullrequest.yml`. **No libhoop checkout, no `GH_TOKEN`, no Rust** — the OSS `_libhoop` stub satisfies the imports.

## Notes

- The suite runs serially (shared gateway + DB) and is documented as non-parallelizable; every test uses uniquely named, self-cleaned resources.
- HTTP response bodies follow a single ownership rule: the call site defers the close, helpers only read.
- Reviewed by the `@review` subagent across two rounds; the initial HIGH finding (harness not building the production middleware stack) was the reason for the `BuildEngine` extraction.

This is the first of a possible multi-level E2E effort (Level 2 = gateway+agent, Level 3 = full CLI); it targets the highest-value, lowest-cost gap first.

Closes ENG-390.

Automated by MisterMal


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1515.0.0-gb321257

gateway/integration/smoke_test.go
```diff
@@ -0,0 +1,374 @@
...
+		Name:               connName,
+		Type:               "database",
+		SubType:            "postgres",
+		AgentId:            agentID,
+		Command:            []string{"psql"},
+		AccessModeRunbooks: "enabled",
+		AccessModeExec:     "enabled",
+		AccessModeConnect:  "enabled",
+		AccessSchema:       "enabled",
+	})
```
matheusfrancisco, line 156: Nice, I would smoke in the resource api and connect api them https://github.com/hoophq/hoop/blob/main/gateway/api/resources/resources.go#L62
the connection api https://github.com/hoophq/hoop/blob/main/gateway/api/connections/connections.go#L43

just because it is more like the UI will create the stuff

matheusfrancisco (approved):

gateway/integration/smoke_test.go
racerxdl: Done in 68b30973 — added `TestResourceLifecycleUIStyle` covering the resources API exactly the way the UI creates infrastructure: resource + role → connection materialized with UI defaults, duplicate 409, delete-with-connections 403 guard, then the unblocked delete. Also extended the suite with 7 more webapp-op tests (sessions envelope, guardrails/data-masking CRUD incl. the OSS entity-type limit, groups, default-plugin contract, reviews, feature-flag toggle). 20 tests total, ~4s after image cache.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1515.0.0-g68b3097

PR #1524, opened 2026-06-16
fix(client): recommend OS-agnostic version manager on version mismatch

### LINKED ISSUE #1516: [Bug]: Version mismatch detected message
### Bug Description

When I run the "hoop connect" command, there is a cool feature that checks if my hoop cli version is the same as the hoop agent. If not, it displays an warning message:

Image

Fact is that I can use hoop cli in Linux, Mac or Windows, so the instructions to upgrade the CLI by using brew doesn't make sense in all OS.

### Steps to Reproduce

1. Have an agent using the version X
2. Install hoop cli using a previous version (X - 1 for example)
3. Try the "hoop connect" command.

The warning message orienting the user to use brew to upgrade CLI will be displayed (even if you use Windows or Linux).

### Expected Behavior

More than just customize the warning message depending on the OS, it think would be much better instruct the user to use the new "hoop versions install" command. This way you can keep a single warning message for all the OS and you also gave the user a better and easier way to manage the CLI versions.

Code: https://github.com/hoophq/hoop/blob/2a385b80c06b8b24f5b8550a80075d8f03f7a903/client/cmd/connect.go#L611

### Actual Behavior

The message would be usuless depending of the OS.

### Version

1.87.2

### Browser(s)

Chrome

### Operating System

Windows

### Screenshots

_No response_

### Additional Context

_No response_

### Relevant Log Output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

p3rotto:
## Summary

Fixes #1516. The `hoop connect` version-mismatch warning hard-coded Homebrew upgrade instructions (`brew install hoop@<version>`), which only make sense on macOS — Windows and Linux users got useless guidance.

The warning now recommends the OS-agnostic CLI version manager:

```
⚠️  VERSION MISMATCH DETECTED! ⚠️
Your CLI version (1.86.0) does not match the agent version (1.87.2).
This WILL cause unexpected behavior.

Install and activate the matching version with the hoop CLI version manager:
  hoop versions install 1.87.2 --use

For more options, visit: https://hoop.dev/docs/clients/cli-versions

(set HOOP_DISABLE_VERSION_CHECK=true to silence this warning)
```

## What changed

`printVersionMismatchWarning` in `client/cmd/connect.go`:

- **Recommends `hoop versions install <agentVersion> --use`**, using the exact agent version (normalized via `NormalizeVersion`, so a leading `v` is stripped).
- **Falls back to the [cli-versions docs](https://hoop.dev/docs/clients/cli-versions)** when the agent version predates the version manager floor (1.74, or 1.86.1 on Windows), via `ValidateInstallableVersion` — so it never prints a command that would fail with a floor error.
- **Honors `HOOP_DISABLE_VERSION_CHECK=true`** and surfaces a silence hint, aligning it with the existing gateway-header warning (`WarnOnceFromServerHeader`).

Reuses existing helpers from `client/upgrade` (`DisableVersionCheckEnv`, `NormalizeVersion`, `ValidateInstallableVersion`) — no new constants or duplicated logic. Added a `printVersionMismatchWarningTo(w, cliVersion, agentVersion)` test seam mirroring the established `warnOnceFromServerHeaderTo` pattern.

## Testing

New table-driven test `client/cmd/connect_test.go` (7 cases): match / unknown / empty / env-disabled no-ops, installable recommendation, `v`-prefix normalization, and below-floor docs fallback. `gofmt` and `go vet` clean; `client/cmd` + `client/upgrade` suites pass.

### DIFF
--- a/client/cmd/connect.go
+++ b/client/cmd/connect.go
@@ -21,6 +21,7 @@ import (
 	"github.com/hoophq/hoop/client/cmd/styles"
 	clientconfig "github.com/hoophq/hoop/client/config"
 	"github.com/hoophq/hoop/client/proxy"
+	clientupgrade "github.com/hoophq/hoop/client/upgrade"
 	"github.com/hoophq/hoop/common/grpc"
 	"github.com/hoophq/hoop/common/log"
 	"github.com/hoophq/hoop/common/memory"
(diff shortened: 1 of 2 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1524.0.0-gdcac313

PR #1528, opened 2026-06-17
Agent-side realtime RDP PII guard: hold-and-release with redact + kill (RD-234…RD-240)

racerxdl:
## Summary

Moves the realtime RDP PII guard from the gateway into the **agent** (agentrs), so PII analysis runs where the plaintext RDP stream already flows — between the agent's two TLS legs, inside the customer network — and the **gateway only ever receives cleared/redacted frames**. Adds a working **redact** policy (blank PII pixels, continue the session) alongside the original **kill** policy, and validates the whole path end-to-end against a live Windows RDP host via the browser (IronRDP) player.

Gated behind the existing `experimental.rdp_pii_guard` flag (default **off**). The separate async/offline recording-analysis feature (`experimental.rdp_pii_detection`) is independent and untouched.

## Architecture

- The gate runs in agentrs on the server→client direction: composites Fast-Path bitmap updates into a shadow canvas, OCRs dirty bands, runs Presidio, and on detection **kills** or **redacts**.
- When the flag is on, the gateway **delegates** to the agent and **suppresses its own gate** — a single enforcement point. The agent resolves its own OCR/Presidio endpoints from its environment; the gateway sends only the enable decision + analysis policy in `SessionStarted` metadata.
- **Fail-closed on delegation mismatch**: if the gateway delegates but the agent can't build the guard (missing endpoints), the session is refused — never run unguarded.

## Tickets (Linear)

RD-234 (port gate to agentrs), RD-235 (violation reporting), RD-236 (config push + proxy wiring), RD-237 (bundled-OCR agent images), RD-238/RD-240 (pixel redaction + policy), RD-239 (capability-pinning filter core). RD-241 (full live E2E + capability-pinning live wiring) remains open.

## Redact mode — four bugs fixed and validated live

Found and fixed while validating redact against a real Windows host:

1. **Redact no longer kills the session** — the proxy watch loop treated every detection as terminal; now only kill/redact_and_kill/overload/analysis-error terminate, plain redact reports-and-continues.
2. **Fail CLOSED on OCR/analysis error** — was fail-open (leaked PII under load); now drops the batch and terminates, with a distinct `pii_guard_analysis_error` audit kind.
3. **OCR content-dedup cache** — skips re-OCR of byte-identical band windows, killing the repaint-storm that saturated the engine (the engine itself is ~10–100 ms/call; CPU was idle — not a GPU problem).
4. **Redaction rewrite redesigned** — emits the negotiated color depth (16/24/32bpp, was hardcoded 24bpp and rejected by the 32bpp client); preserves non-PII content (forwards non-overlapping bitmaps unmodified, repaints only PII regions); and blanks the **full detection bbox** so PII delivered in earlier deltas (a secret typed character by character) is covered, not just the latest delta.

Plus defensive hardening on the redaction path (saturating geometry, framebuffer-bounds validation, runtime PER/length caps).

## OCR engine

Bundled into the agent container (loopback only) as two image flavors — `hoop-agent-ocr` (CPU, multiarch) and `:<tag>-gpu` (amd64/CUDA) — so screen pixels never leave the agent process boundary. RapidOCR on CPU is the validated path.

## Testing

- agentrs: 77 tests pass, including the adversarial leak-oracle suite and new tests that decode emitted redaction PDUs through the real ironrdp parser to assert content preservation, full-bbox blanking, negotiated-bpp encoding, and fail-closed-on-analysis-error.
- Gateway rdp/broker/appconfig tests pass.
- Validated live: redact against a Windows RDP host through the browser player (content preserved, PII blanked, fast on CPU).

## Known limitations (experimental; follow-ups)

- Fragmented bitmap updates forward First/Next before reassembly (clients don't render until Last; full-bbox blank covers it) — holding until reassembly would be stricter.
- OCR-cache hit relies on a 64-bit hash; a byte-equality check on hit would make stale-word reuse provably impossible.
- Canvas decode failure currently fails open (pre-existing) — should fail closed for full rigor.
- Capability-pinning live wiring (RD-239) is tested as a filter but not yet applied during negotiation (RD-241).

## Dev tooling (included)

- `agentrs/examples/rdp_caps_probe.rs` — probes what update codes a server actually sends (Bitmap vs SurfaceCommands).
- `Dockerfile.localbuild` — builds `hoop:local` (hoop_rs + Go + webapp UI) from source with no host cross-toolchain.
- `deploy/docker-compose/docker-compose.pii-demo.yml` + `scripts/dev/run-pii-demo.sh` — stand up the full demo stack locally.

---

Automated by MisterMal


sandromello:
✅ Build Completed with Success, Version=1528.0.0-gc4e153d

matheusfrancisco (approved):
There are some conflicts :)

sandromello:
✅ Build Completed with Success, Version=1528.0.0-g15abd03

sandromello:
✅ Build Completed with Success, Version=1528.0.0-gf2b1178

PR #1529, opened 2026-06-17
fix: add ForceGroupsSync configuration for OIDC and SAML authentication

luanlorenzo:
## 📝 Description

Adds an opt-in **Force Groups Sync** setting (`force_groups_sync`) to the OIDC and SAML auth configuration that makes the identity provider authoritative for a user's group membership.

This fixes the reported group-clearing bug: when all of a user's roles are removed at the IDP, the provider omits the groups claim from the login token entirely (rather than sending an empty list). Today the gateway only updates groups when the claim is *present*, so an absent claim left the user's stale groups untouched. With this setting enabled, an absent claim is treated as "the user has no groups" and the groups are cleared.

The behavior is **gated and off by default**, so existing deployments are unaffected — groups are still only updated when the IDP sends them, exactly as before. Operators that want the IDP to be the source of truth turn the toggle on per provider.

Google Workspace is intentionally excluded from the absent-claim handling: it fetches groups out-of-band via the Admin API (where an absent claim in the token is expected), so its dedicated path remains authoritative. 

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- **Model** (`gateway/models/server_auth_config.go`): added `ForceGroupsSync bool` (`json:"force_groups_sync"`) to `ServerAuthOidcConfig` and `ServerAuthSamlConfig`. Stored inside the existing `oidc_config`/`saml_config` JSONB columns, so **no migration is required** — existing rows decode to `false` (legacy behavior).
- **OIDC** (`gateway/idp/oidc/oidc.go`): added `ForceGroupsSync` to provider `Options`; `parseUserInfo`'s absent-claim (`case nil`) branch now syncs to empty only when `ForceGroupsSync && !mustFetchGsuiteGroups`.
- **SAML** (`gateway/idp/saml/saml.go`, `gateway/api/login/saml/login.go`): added the flag to `Options`/`ServiceProvider`; `parseToUserInfo` seeds `MustSyncGroups` from the flag, while a present groups attribute still forces a sync regardless.
- **Provider wiring** (`gateway/idp/core.go`): maps the flag from DB config into each provider's `Options`, and includes it in the OIDC and SAML config-change fingerprints so toggling rebuilds the cached provider without a restart.
- **API + OpenAPI** (`gateway/api/serverconfig/auth.go`, `gateway/api/openapi/types.go`, regenerated `autogen/`, `openapiv3.json`): request/response mapping, pre-flight provider validation, and schema docs.
- **Web UI** (`webapp/.../authentication/...`): added a "Force Groups Sync" toggle below "Groups Claim" for both OIDC and SAML, with read/save wiring in the re-frame events.

### Test Configuration:
- **Browser(s)**: Chrome
- **OS**: macOS

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

CleanShot 2026-06-17 at 17 15 05

## ✅ Checklist


racerxdl (approved):
lgtm!

PR #1530, opened 2026-06-18
feat: promote default-enabled feature flags to permanent features

p3rotto:
## Summary

Promotes the three `Default: true` experimental feature flags to permanent, always-on features. Every gating conditional is removed and the flags are deleted from the catalog so they can no longer be toggled.

| Flag | Change |
|------|--------|
| `experimental.event_routing` | Removed the publish guard and the per-org flag-watching supervisor — the worker pool now always starts (still gated only by `EVENT_ROUTING_WORKERS`). Removed the webapp page guards/redirects, the sidebar gate, and the BETA badges. |
| `experimental.agent_async_ssh` | SSH / `SessionClose` packets are always dispatched on per-packet goroutines. Removed the obsolete `*_FlagOff` integration tests (they asserted the old synchronous bug that no longer exists) and de-flagged their twins. |
| `experimental.hoop_tunnel` | Had no runtime consumers anywhere — removed the catalog entry and fixed a stale comment in `tunnel/client/serverinfo.go`. |

Migration `000099` deletes orphaned `org_feature_flags` overrides for the three promoted flags (no-op down).

## Notes

- No `appconfig`/env changes → no helm updates needed.
- The admin "Experimental" settings screen is data-driven off `/api/feature-flags`; the three flags simply stop appearing.
- Pre-existing gofmt drift in `publish.go` / `agent.go` (already present on `main`, in regions untouched by this PR) was deliberately left alone to keep the diff focused.

## Test plan

- [x] `go build ./...` across all 5 modules
- [x] `go vet` clean, including `-tags integration`
- [x] Full Go test suite passes (`go test ./...` per module)
- [x] Changed webapp files lint clean (only pre-existing `set-state-in-effect` warnings remain in `Form/index.jsx`)
- [ ] SSH integration tests (`//go:build integration`, require Docker/testcontainers) — to be exercised in CI


sandromello:
✅ Build Completed with Success, Version=1530.0.0-g00d522d

matheusfrancisco (approved):

luanlorenzo (approved):

Issue #1548, opened 2026-06-23
[Bug]: Access to AWS RDS PostgreSQL with experimental.db_exec_driver


qwerty1q2w:
### Bug Description

<img width="2384" height="579" alt="Image" src="https://github.com/user-attachments/assets/2a4d6fac-5039-4c68-9aa9-8a6b56e50e3a" />

### Steps to Reproduce

Try to open AWS RDS PostgreSQL in web client. With experimental.db_exec_driver

### Expected Behavior

Access without errors

### Actual Behavior

errors

### Version

1.99.0

### Browser(s)

Firefox

### Operating System

Ubuntu

### Relevant Log Output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

matheusfrancisco:
Thanks for reporting. I will take a look at

PR #1549, opened 2026-06-23
feat: promote experimental feature flags to enabled by default (ENG-477)

p3rotto:
## 📝 Description

Promotes all experimental feature flags (and the one beta flag, `beta.oracle_native`) to be **enabled by default** by flipping their catalog `Default` to `true`. Flags remain in the catalog and stay toggleable per-org — this is the literal "enabled by default" change, not a full removal of the flag mechanism.

Also removes three **orphaned** catalog entries (`experimental.event_routing`, `experimental.agent_async_ssh`, `experimental.hoop_tunnel`). These were promoted to permanent always-on behavior in #1530 (their gating *and* catalog entries were deleted there), but a later merge accidentally re-added only the catalog entries — leaving dead admin toggles with zero code consumers.

## 🔗 Related Issue

https://linear.app/hoophq/issue/ENG-477/promote-all-experimental-features-to-be-enabled-by-default

## 🚀 Type of Change

## 📋 Changes Made

- `common/featureflag/featureflag.go`: flipped `Default: false → true` for `nightly_flag`, `log_exec_input`, `rdp_pii_detection`, `rdp_pii_guard`, `rulepacks`, `db_exec_driver`, `http_session_analyzer`, and `beta.oracle_native`.
- Removed orphaned catalog entries `experimental.event_routing`, `experimental.agent_async_ssh`, `experimental.hoop_tunnel` (no consumers; gating already removed in #1530). No migration: `SetAll` filters by catalog membership and `IsEnabled` returns false for unknown names, so any stale per-org rows are inert.
- `common/featureflag/featureflag_test.go`: reworked `TestIsEnabled_DefaultFalse` → `TestIsEnabled_FallsBackToDefault`, asserting `IsEnabled` returns the flag's actual `Default` with no override (value-agnostic, since no flag defaults to false anymore).

### Test Configuration:
- **OS**: macOS (Darwin)

### Tests performed:
- [x] Unit tests pass (`go test ./common/featureflag/... ./agent/controller/featureflagstate/...`)
- [ ] Integration tests pass

`go build ./common/... ./agent/... ./gateway/... ./client/...` passes. No test depended on the flipped defaults: `featureflagstate` tests use synthetic flag names, and `featureflag` tests drive `nightly_flag` via explicit `Set()`.

## ✅ Checklist

## 📄 Additional Notes

> [!WARNING]
> **`experimental.rdp_pii_guard` is now ON by default and fails CLOSED.**
> At `gateway/rdp/irongw.go:205-227`, when the guard is enabled the gateway refuses any RDP session whose agent has not advertised `CapabilitySupportsPIIGuard` (only agents with Presidio + OCR configured do). So after this change, RDP sessions through agents **without** `MSPRESIDIO_ANALYZER_URL` + `RDP_OCR_SERVER_URL` configured will be **refused** with *"Connection refused: PII guard is enabled but agent X cannot enforce it."*
>
> This was a deliberate, accepted decision. Before upgrading, operators should either configure those endpoints on RDP agents, or explicitly disable `experimental.rdp_pii_guard` per-org.

Safety note on the other RDP flag: `experimental.rdp_pii_detection` is safe to default-on — its supervisor (`gateway/rdp/analyzer/supervisor.go:37-44`) and recorder enqueue site (`recorder.go:332`) both short-circuit when Presidio/OCR are absent, so it cleanly no-ops where the dependency isn't present.


common/featureflag/featureflag.go
```diff
@@ -36,77 +36,56 @@ var catalog = map[string]Flag{
...
 		Components:  []Component{ComponentGateway},
 	},
 	"experimental.db_exec_driver": {
 		Name:        "experimental.db_exec_driver",
 		Description: "Run Postgres/MySQL/MSSQL/Oracle exec commands through in-process Go database drivers instead of spawning the vendor CLI (psql/mysql/sqlcmd/sqlplus). Eliminates client meta-command shell escapes (e.g. psql \\!, sqlplus HOST) and keeps the connection credential out of any user-reachable process.",
-		Default:     false,
+		Default:     true,
 		Stability:   StabilityExperimental,
 		Components:  []Component{ComponentAgent},
 	},
```
matheusfrancisco, line 70: Let this be false, there is a bug https://github.com/hoophq/hoop/issues/1548 
and this changes a bit how the exec works and the return  

So, this has broken the table view for Oracle and Mysql when using via driver and not exec (for now)

common/featureflag/featureflag.go
```diff
@@ -36,77 +36,56 @@ var catalog = map[string]Flag{
 	"experimental.nightly_flag": {
 		Name:        "experimental.nightly_flag",
 		Description: "Example flag for testing the feature flag system (no-op)",
-		Default:     false,
+		Default:     true,
 		Stability:   StabilityExperimental,
 		Components:  []Component{ComponentGateway, ComponentAgent, ComponentClient},
 	},
```
matheusfrancisco, line 42: This is a dummy feature flag that does nothing; it was just to test the api/cli/frontend render and agent back then 😄

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1549.0.0-g802ab8a

common/featureflag/featureflag.go
p3rotto: Good catch, reverted `experimental.db_exec_driver` to `Default: false`

common/featureflag/featureflag.go
p3rotto: Yep, removed it from the catalog entirely in c76b8b49 since it had no production consumers. The flag-system unit tests now register their own in-catalog fixture inside featureflag_test.go, so the dummy no longer ships and the tests stay decoupled from any real flag.

sandromello:
✅ Build Completed with Success, Version=1549.0.0-gc76b8b4

sandromello:
✅ Build Completed with Success, Version=1549.0.0-g228389f

PR #1537, opened 2026-06-19
feat: MCP HTTP proxy

EmanuelJr:
> [!IMPORTANT]
> Every PR MUST have **exactly one** of these labels, the merge is blocked otherwise:
> - `major` / `minor` / `patch` publishes a new release on merge with the corresponding semver bump.
> - `skip-release` merges without publishing a release (use for docs, CI changes, refactors, etc.).

## 📝 Description

Adds first-class support for **MCP (Model Context Protocol)** connections backed by the `httpproxy` infrastructure, including a complete OAuth 2.1 login flow so admins can authorize against remote MCP servers that protect their endpoints with OAuth (e.g. `https://mcp.figma.com/mcp`).

The gateway acts as an OAuth 2.1 client on the admin's behalf: it discovers the upstream authorization server (RFC 9728 / RFC 8414), optionally registers a client via Dynamic Client Registration (RFC 7591), and drives an Authorization Code + PKCE flow. The obtained token is frozen into the connection's `HEADER_AUTHORIZATION` configuration. Secrets at rest (PKCE verifier, client secret, obtained tokens) are stored as AES-256-GCM ciphertext using the same credential vault as connection credentials.

## 🚀 Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ⚡ Performance improvement
- [ ] ✅ Test update
- [ ] 🔧 Build configuration change
- [ ] 🧹 Chore

## 📋 Changes Made

**Backend (Gateway)**
- Added `mcp` connection type (`common/proto/const.go`) and wired it through credential generation, session revocation, and "is configured" checks (`connection_credentials.go`, `services/credentials.go`).
- New MCP OAuth flow handlers (`connection_mcp_oauth.go`, `connection_mcp_oauth_client.go`) implementing OAuth 2.1 discovery (RFC 9728 / RFC 8414), Dynamic Client Registration (RFC 7591), and Authorization Code + PKCE exchange.
- New `private.mcp_oauth_flows` table + GORM model (`models/mcp_oauth_flow.go`, migration `000099`) — single-use, TTL-bounded flow rows with AES-256-GCM encrypted secrets.
[...]

PR #1551, opened 2026-06-23
fix(webapp): restore connection metadata download in release build

luanlorenzo:
## 📝 Description

The Resource Catalog stopped loading roles on `sandbox` and `demo`: `https://<host>/data/connections-metadata.json` returns the SPA `index.html` instead of the connections JSON.

**Root cause:** PR #1537 accidentally commented out the `curl` download in `webapp/scripts/download-connection-metadata.sh`. That file is gitignored and produced solely by this script at build time, so every release since #1537 ships without `/data/connections-metadata.json`. The gateway's static middleware then misses the file and falls through to the SPA catch-all (`gateway/api/server.go:151`), returning `index.html`.

Sandbox and demo broke because auto-deploy (#1536) bumped them onto the latest release; app/use stayed on the pre-#1537 release and keep working.

## 🚀 Type of Change

## 📋 Changes Made

- Restore the `curl` download (and the HTTP-status / JSON-validation guards) in `webapp/scripts/download-connection-metadata.sh`, so the release build fetches the file again and fails loudly on a bad download.

## 🧪 Testing

Ran the restored script locally:

```
HTTP Status: 200
✓ Download complete: resources/public/data/connections-metadata.json
  File size: 83085 bytes
✓ JSON valid (connections: 32)
```

### DIFF
--- a/webapp/scripts/download-connection-metadata.sh
+++ b/webapp/scripts/download-connection-metadata.sh
@@ -31,62 +31,62 @@ else
     echo "  Using unauthenticated request (no token)"
 fi
 
-# HTTP_CODE=$(curl "${CURL_ARGS[@]}" \
-#      "https://api.github.com/repos/$ORG/$REPO/contents/$FILE")
+HTTP_CODE=$(curl "${CURL_ARGS[@]}" \
+     "https://api.github.com/repos/$ORG/$REPO/contents/$FILE")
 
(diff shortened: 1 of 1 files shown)


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1551.0.0-g52039de

PR #1544, opened 2026-06-22
fix(rdp): repair GPU OCR image, default PII guard to redact, + GCP deploy compose

racerxdl:
Three fixes for the agent-side RDP PII guard on GPU:

1. **Repair the GPU OCR image** — it crash-looped on `libcudart.so.13`. Pin `onnxruntime-gpu==1.20.1` (CUDA 12), ship the full CUDA-12 runtime wheels, and resolve them via a `/opt/cuda-libs` symlink farm. Validated on a T4: `device=onnxruntime-cuda`, no CPU fallback.
2. **Default policy `kill` → `redact`** (behavior change). `kill` killed the session on the first detected entity → black screen in seconds. Changed in code + helm + `.env.sample`.
3. **Add `deploy/docker-compose/docker-compose.gcp-gpu-agent.yml`** — single-VM GCE+T4 deploy.

Automated by MisterMal

PR #1552, opened 2026-06-24
fix(rdp): don't 403 RDP through agents that predate the PII-guard capability handshake

racerxdl:
## Problem

With `experimental.rdp_pii_guard` enabled (now **default true** after #1549), the gateway capability gate failed closed on "unknown" capability — returning **RDCleanPath 403** for every RDP connection through any agent that hadn't advertised `supports_pii_guard`. Older agents never advertise, so this **bricks all RDP through not-yet-upgraded agents**, even connections never meant to be guarded.

## Fix

3-way capability handling in `gateway/rdp/irongw.go`:
- **known + capable** → delegate the guard (unchanged).
- **known + incapable** → refuse with a clear error (a guard-capable agent that's misconfigured / missing OCR+Presidio).
- **unknown** → run the session **unguarded** instead of 403. `forceUnguarded` also suppresses the gateway-side gate (true passthrough). An old agent can't run the guard anyway.

Deliberate availability-over-enforcement trade-off for the unknown case (documented in code; cross-referenced on `broker.AgentCapability`).

Note: the `rdp_pii_guard` flag description in featureflag.go still says "kill the session" — stale since #1544 made `redact` the default (not changed here).

Automated by MisterMal

### DIFF
--- a/gateway/broker/session.go
+++ b/gateway/broker/session.go
@@ -289,7 +289,11 @@ func SetAgentCapabilities(agentID string, capabilities map[string]string) {
 // The two booleans are distinct on purpose:
 //   - known=false: the agent has not advertised capabilities within the wait
 //     (old agent, or a degenerate connection). Callers that need a security
-//     guarantee must treat this as "cannot" and fail closed.
+//     guarantee should generally treat this as "cannot" and fail closed —
+//     EXCEPT where availability across mixed-version rollouts outweighs it: the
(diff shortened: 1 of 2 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1552.0.0-gf1c1649

PR #1555, opened 2026-06-24
feat(rdp): windowed latency metrics for the PII guard hot path

racerxdl:
## Problem

RDP sessions through the PII guard are slow even on the GPU node, and there's no visibility into which pipeline stage dominates per-frame latency (compositing vs OCR vs Presidio vs redaction).

## Change

Adds self-aggregating, **logs-only** latency instrumentation to the PII guard hot path on both sides:

- **agentrs (Rust)** — `LatencyAggregator` per session: composite / ocr / presidio / redact / total.
- **gateway (Go)** — mirror aggregator (reuses the OCR/Presidio durations already in `SnapshotResult`). Kill-only gate, so no redact stage there.

Per-stage timings are aggregated into **5s windows** and emitted as one summary line (`n/avg/p50/p95/max`), so a busy session never floods the log. Samples are **capped per stage** and the summary is computed + logged **outside the aggregator lock** to keep the hot path cheap.

Sample line:
```
piigate latency: composite[n=42 avg=0.4ms ...] ocr[n=40 avg=58ms p95=120ms ...] presidio[n=40 avg=12ms ...] redact[n=1 ...] total[n=42 avg=74ms ...] batches=42 detections=1 forwarded_kib=8123
```

## Why logs-only
agentrs has no telemetry exporter and the gateway's OTel/Honeycomb wiring is gateway-only; a unified spans pipeline would need new agent egress + secrets. Out of scope for "find the offender."

## Verification
- Reviewed with `@review`: **no enforcement-path regression** (the `bool -> outcome struct` refactor preserves every kill/redact/fail-closed/fail-open decision); addressed findings (sample cap, flush-outside-lock, zero-duration skip, docs).
- Both log lines confirmed emitting via captured-output tests.
- 87 agentrs piigate tests + Go rdp suite + new aggregator/log tests pass; `go vet` + `clippy` clean.

Pure instrumentation, active only when `experimental.rdp_pii_guard` runs. No behavior change.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/analyze.rs
+++ b/agentrs/src/piigate/analyze.rs
@@ -16,6 +16,7 @@ use tokio::sync::Semaphore;
 use tokio_util::sync::CancellationToken;
 
 use super::bands::{split_bands, OcrChunk, YBand, DEFAULT_BAND_PADDING, MAX_CHUNK_ROWS};
+use super::metrics::LatencyAggregator;
 use super::ocr::{join_words, OcrClient, Word};
 use super::presidio::{analyze_text, AnalysisParams, PresidioClient, SnapshotResult};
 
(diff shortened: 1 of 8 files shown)


sandromello:
✅ Build Completed with Success, Version=1555.0.0-g5098c9c

EmanuelJr (approved):

PR #1557, opened 2026-06-25
feat(rdp): surface OCR sidecar inference time in PII guard metrics

racerxdl:
## Why

Production latency logs (from #1555) showed **OCR dominates** the PII guard hot path: ~340ms avg, p50 ~210ms, **p95 up to 1.8s**, ~95% of total batch time. The huge p95/max tail points to queueing/contention rather than raw inference — but we can't prove it without the sidecar's own timing.

## What

Surface the OCR sidecar's self-reported `duration_ms` (already parsed, previously `#[allow(dead_code)]`) alongside the agent-measured wall-clock. The 5s window line gains:

- `ocr_server[...]` — the **slowest parallel chunk's** sidecar inference time (the critical path, directly comparable to `ocr` wall-clock).
- `ocr_calls` / `ocr_chunks` — OCR round-trips vs total chunks (cache effectiveness).
- `ocr_sent_kib` — BMP payload size.

**Interpretation**: if `ocr` wall-clock ≫ `ocr_server`, the bottleneck is queueing/contention (likely the 4 uvicorn workers fighting over one T4) → fix is concurrency/batching. If `ocr_server` ≈ `ocr`, it's raw inference → fix is smaller model / fewer pixels / bigger GPU.

## Safety

- `OcrClient::extract` now returns `OcrExtract { words, server_ms, request_bytes }`; `validate_words` output is byte-identical to before, OCR cache unaffected, cancellation/error paths unchanged.
- Non-finite/negative sidecar values coerced to 0 (malformed telemetry can't poison the window).
- Reviewed with `@review`: all findings LOW, no enforcement change. 90 piigate tests pass (+ new edge-case tests), clippy clean.

Diagnostic-only; active only when `experimental.rdp_pii_guard` runs.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/analyze.rs
+++ b/agentrs/src/piigate/analyze.rs
@@ -201,35 +201,56 @@ impl Analyzer for BandAnalyzer {
                         .await
                         .expect("piigate: OCR semaphore closed");
                     if cancel.is_cancelled() {
-                        return Ok((idx, Vec::new()));
+                        return Ok((idx, Vec::new(), 0.0, 0usize));
                     }
-                    let words = tokio::select! {
+                    let extract = tokio::select! {
                         res = ocr.extract(&crop, fb_width, win_height) => res.with_context(|| {
                             format!("OCR failed on chunk [{},{})", win.y0, win.y1)
                         })?,
-                        _ = cancel.cancelled() => return Ok((idx, Vec::new())),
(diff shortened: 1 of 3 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1557.0.0-g3543338

PR #1558, opened 2026-06-25
perf(rdp): skip re-analyzing byte-identical RDP repaints in PII guard

racerxdl:
## Why

Production latency logs (#1555/#1557) proved **OCR dominates** the PII guard (~95% of batch time), and the OCR cache is **architecturally ineffective**: `composite()` marked a region dirty (→ OCR'd) on **every** bitmap paint. But RDP constantly resends **byte-identical tiles** (idle repaints, unchanged backgrounds), so unchanged content was re-OCR'd every frame — `ocr_calls ≈ ocr_chunks` (near-zero reuse) in the logs.

## What

`composite_bitmap` now compares before writing and reports whether any pixel **actually changed**. The gate marks a region dirty (→ analyzes/OCRs it) **only when it changed** (or the canvas grew). Unchanged paints are still **forwarded** — only analysis is skipped — so delivery is byte-for-byte unchanged.

Mirrored on both gates (agentrs Rust + gateway Go) so they enforce identically. `CompositeBitmap` (Go) now returns `bool`; the full-frame worker and rdpbench ignore it.

## Safety (security-critical guard)

Skipping analysis on identical pixels cannot leak PII:
- An unchanged paint's pixels are **already on the shadow canvas** and were analyzed when they first appeared — there is no unanalyzed content.
- A **first paint into newly grown canvas area is forced changed** (covers the black-on-zeroed-canvas edge).
- **Batch sealing is unaffected**: it keys on changed paints, which still dirty their band; the painted-then-overwritten protection (`flash_attack`/`cross_batch`) still holds.

## Verification

- `@review`: **no CRITICAL/HIGH**, security invariant verified path-by-path. Added the reviewer-requested regression tests (zero-pixel first paint; identical-PII-repaint-still-caught) on both sides.
- New `paints_total/paints_changed` metrics quantify the real skip rate in prod.
- 94 agentrs piigate tests + Go rdp suite pass; clippy + `go vet` clean.

Pure analysis-scheduling optimization; active only when `experimental.rdp_pii_guard` runs. No change to forwarded bytes or enforcement.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/canvas.rs
+++ b/agentrs/src/piigate/canvas.rs
@@ -69,7 +69,12 @@ pub fn to_rgba(src: &[u8], width: usize, height: usize, bpp: usize) -> Option<Ve
 }
 
 /// Draws a decoded RGBA bitmap patch onto the framebuffer at (dst_x, dst_y).
-/// Out-of-bounds regions are clipped.
+/// Out-of-bounds regions are clipped. Returns whether any pixel actually
+/// CHANGED: RDP routinely resends byte-identical tiles (idle repaints, cursor
+/// trails, unchanged backgrounds), and a paint that changes nothing has no new
+/// content to analyze. Reporting "unchanged" lets the caller skip marking the
+/// region dirty, so it is not needlessly re-OCR'd — this is safe because the
+/// shadow canvas already holds (and already analyzed) those exact pixels.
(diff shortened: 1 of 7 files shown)


sandromello:
✅ Build Completed with Success, Version=1558.0.0-g00f0551

matheusfrancisco (approved):

PR #1561, opened 2026-06-26
feat(rdp): force server-side performance flags to cut bitmap updates

racerxdl:
## Why

RDP sends screen changes as bitmap tiles. The desktop wallpaper, full-window drag ghosting, menu animations, and visual theming generate large/frequent bitmap updates with no information value — they cost bandwidth and, when the PII guard is active, inflate the screen area the guard must OCR per frame. RDP's performance flags let a client ask the server to skip exactly these.

## What

The proxy now intercepts the client's **Client Info PDU** (post-CredSSP MCS sequence) and OR-s in:
`DISABLE_WALLPAPER | DISABLE_FULLWINDOWDRAG | DISABLE_MENUANIMATIONS | DISABLE_THEMING`. Applied to all RDP sessions.

The proxy is a transparent passthrough (it doesn't build the connection), so the flags are injected by patching the in-flight PDU rather than reconfiguring a connector. The browser WASM IronRDP client (the only client today) is a prebuilt blob that can't be reconfigured, so interception is the correct place.

## Safety (live handshake + credential-bearing PDU)

- **Wire fidelity guard**: before patching, the unmodified decoded PDU is re-encoded and required to be **byte-identical** to the original. Only then is it patched. So the only wire change is ever the flag bits — never an invented `session_id`, never a dropped RDP-10+ trailing byte. Any non-canonical frame is forwarded verbatim.
- **Never breaks a session**: failure to find/patch the Client Info PDU degrades to transparent forwarding (logged, not fatal).
- **No desync**: the bidirectional relay forwards each PDU before the next read and uses `read_pdu`, documented cancel-safe in ironrdp_async 0.5.0, so `select!` can't lose bytes or split a frame.
- Credentials survive the re-encode (tested).

## Review & tests

- Reviewed with `@review` twice; both initial HIGH findings (session_id invention, reserved-byte loss) resolved by the fidelity guard and re-confirmed. Cancel-safety verified against the exact library version.
- 6 round-trip/fidelity tests; 105 agentrs tests pass; clippy clean.

Patch-level; no protocol additions, no config. Once deployed, expect fewer bitmap updates (visible as lower paints_total in the guard latency logs from #1558).

Automated by MisterMal

### DIFF
--- a/agentrs/src/rdp_proxy.rs
+++ b/agentrs/src/rdp_proxy.rs
@@ -187,6 +187,19 @@ where
     )
     .await?;
 
+    // -- Intercept the client's Client Info PDU to force server-side rendering
+    //    performance flags (disable wallpaper, full-window drag, menu
+    //    animations, theming). This cuts the volume of bitmap updates the
+    //    server sends, which both reduces bandwidth and — when the PII guard is
+    //    active — shrinks the amount of screen the guard must OCR. Applied to
+    //    every RDP session through the proxy. Best-effort: if the Client Info
+    //    PDU is not seen (unexpected sequence), forwarding continues unmodified.
+    if let Err(e) = intercept_client_info(&mut client_framed, &mut server_framed).await {
+        // Never fail the session on an optimization: log and fall through to
+        // transparent forwarding with the client's original flags.
+        info!("RDP perf-flag injection skipped ({e:#}); forwarding client flags unchanged");
+    }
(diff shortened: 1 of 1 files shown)


sandromello:
✅ Build Completed with Success, Version=1561.0.0-g544416b

matheusfrancisco (approved):

PR #1562, opened 2026-06-26
docs(adr): line-level OCR cache for the RDP PII guard

racerxdl:
## Summary

ADR-0002 proposing a **line-level OCR cache** to cut the recognition-stage cost (the proven ~270ms floor) in the realtime RDP PII guard.

## Key points

- **Change detection at RDP tile granularity, OCR at full-width line granularity.** RDP changes are rectangular tiles, but text spans tiles horizontally — per-tile OCR fragments words (measured: `john.doe@example.com` → `ohn.d | be@e) | ample`) and would **miss PII**. So changed tiles select which full-width lines are dirty; only those lines are re-OCR'd, the rest served from a persistent line cache.
- **Measured win (real engine):** a single-line edit drops a frame from ~387ms to ~55–130ms (66–85% faster); 4 lines → 58% faster; never slower than today.
- **Security invariant preserved:** det still runs over the full band each frame; a line is reused only when byte-identical (already analyzed); any changed line misses and is freshly OCR'd. Includes the proof and the hash-collision analysis.
- **Geometry-agnostic** (any tile size/shape), **complements #1558** (which skips fully-static frames), and is **orthogonal** to a native OCR server (which addresses the ~122ms Python overhead, not the rec floor).

Rejected alternatives documented: tile-grid cache, downscaling (accuracy + unknown client resolution).

Design doc only — no code. Seeking sign-off before implementation.

Automated by MisterMal

### DIFF
--- a/docs/adr/0002-rdp-pii-guard-line-ocr-cache.md
+++ b/docs/adr/0002-rdp-pii-guard-line-ocr-cache.md
@@ -1,0 +1,246 @@
+# ADR-0002: Line-level OCR cache for the RDP PII guard
+
+- **Status:** Proposed
+- **Date:** 2026-06-26
+- **Related:** PR #1555 (windowed latency metrics), PR #1557 (OCR sidecar timing), PR #1558 (skip byte-identical repaints)
+
+## Context
+
+The realtime RDP PII guard holds server→client frames, OCRs the changed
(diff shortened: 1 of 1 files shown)


sandromello:
✅ Build Completed with Success, Version=1562.0.0-gfe5e027

matheusfrancisco (approved):

PR #1563, opened 2026-06-26
perf(rdp): line-level OCR cache for the PII guard (ADR-0002)

racerxdl:
Implements ADR-0002 (#1562).

## Why

Recognition on full-width text crops is the PII guard's dominant cost (~228ms of a ~390ms synthetic frame; ~105ms on a real desktop frame). Most frames change only one line while the rest of the screen is static, yet the whole dirty band was re-OCR'd every time.

## What

Caches recognition **per text line**, re-recognizing only lines whose pixels changed. Detection still runs over the full dirty band every frame (cheap, scales with band height), so no line is missed at segmentation — only the expensive recognition is cached.

**Measured on a real desktop frame (1776×752):**
| Scenario | Today | Line cache |
|---|---|---|
| 1-line edit | ~105ms rec | ~8ms (1 line) |
| static frame | ~105ms rec | 0 (full cache hit) |

Combined with the dirty band shrinking det cost, a one-line edit in the real gate is ~25ms vs ~224ms.

**Sidecar**: splits `/ocr` into `/det` (line quads, each with a SHA-256 of the exact perspective recognition crop) and `/rec` (recognizes given boxes server-side with the same crop → byte-identical text to `/ocr`). `/ocr` kept for compat.

## Security (this is a PII guard)

Three review rounds; all HIGH findings resolved:
- **Cache validity = the exact recognition pixels.** Key is the sidecar's SHA-256 over the same `get_rotate_crop_image` crop `/rec` recognizes, plus full-canvas position. Proven on the engine: identical crop pixels ⟹ identical rec output, so a hit returns exactly what rec would. A change rec would read differently always changes the hash → miss. No bbox-vs-quad gap.
- Detection runs every frame → newly appeared/changed lines always recognized; persistent PII re-detected each frame.
- `/rec` cardinality mismatch and malformed `crop_hash` both **fail closed** (analysis returns Err → gate drops the batch).
- `Option<LineKey>` (no magic sentinel); SHA-256 not a 64-bit hash.

## Tests / cost

110 agentrs tests pass (incl. identical-skip, changed-rerec, persistent-PII, cardinality-fail-closed, moved-line, crop-hash validation); clippy clean. Extra det-side crop+hash is ~0.6ms/line, dwarfed by the rec savings. Adds `lines_total`/`lines_recognized` metrics for prod hit-rate visibility.

Gateway (Go) gate keeps its existing behavior (kill-only, unchanged); agent-side first per ADR.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/analyze.rs
+++ b/agentrs/src/piigate/analyze.rs
@@ -7,56 +7,52 @@
 //! indistinguishable from full-frame analysis.
 
 use std::collections::HashMap;
-use std::hash::{Hash as _, Hasher as _};
 use std::sync::{Arc, Mutex};
 
 use anyhow::Context as _;
 use async_trait::async_trait;
-use tokio::sync::Semaphore;
-use tokio_util::sync::CancellationToken;
 
-use super::bands::{split_bands, OcrChunk, YBand, DEFAULT_BAND_PADDING, MAX_CHUNK_ROWS};
+use super::bands::YBand;
 use super::metrics::LatencyAggregator;
-use super::ocr::{join_words, OcrClient, Word};
+use super::ocr::{join_words, LineBox, OcrClient, Word};
 use super::presidio::{analyze_text, AnalysisParams, PresidioClient, SnapshotResult};
 
-/// Per-window OCR result cache: skips re-OCR of a band whose pixels are
-/// byte-identical to the last time that exact window was OCR'd.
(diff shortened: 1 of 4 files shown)


sandromello:
✅ Build Completed with Success, Version=1563.0.0-gc09a238

EmanuelJr (approved):

PR #1566, opened 2026-06-29
fix(rdp): keep empty OCR line recognitions as blank slots in PII guard

racerxdl:
## Symptom

After deploying the line-level OCR cache (#1563), the **bottom half of the RDP screen disappeared** — consistently, on every session/relogin.

## Root cause

Live logs showed:
```
piigate: analysis failed, failing closed (dropping batch, terminating):
OCR recognize returned 32 words for 33 boxes on band [296,920)
```

Band `[296,920)` is the bottom region. The chain:

1. `/det` detects 33 lines in the bottom band.
2. The agent sends all 33 to `/rec`. The recognizer reads one as **empty text** (a blank row / icon / separator — entirely normal).
3. The agent ran `/rec` output through `validate_words`, which **drops empty-text entries** (correct for `/ocr`, fatal for `/rec`).
4. 33 boxes → 32 words → the fail-closed cardinality check (correctly) fires and **drops the whole band**, killing its forwarding.
5. Top band renders, bottom band always dropped → bottom half gone, every time.

The fail-closed check was right; the bug was **violating `/rec`'s positional contract** by dropping a slot.

## Fix (agent-side only)

`/rec` is positional — result `i` belongs to box `i` — so an unusable recognition must never be dropped.

- `recognize()` → `validate_words_positional` returning `Vec<Option<Word>>`: one slot per box, `None` for empty/invalid, **never dropped**. Also enforces exactly N entries at the HTTP boundary, so a genuine protocol violation still errors immediately (fail-closed preserved).
- Per-line cache value is `Option<Word>`, so a stable blank line is a cache hit (no needless re-rec).
- `analyze()` reassembles by slot: `None` → no Presidio text, no redaction box, but the slot still exists and is cached. Changed lines are still always re-recognized (key = position + `crop_hash`), so no PII slips through.
- Shared per-word sanitization extracted to `validate_word` (one rule set for `/ocr`, `/det`, `/rec`).

## Tests

- `empty_recognition_is_kept_as_blank_slot_not_dropped` — reproduces the vanished-band bug (fails before, passes after).
- `positional_validation_keeps_one_slot_per_word` — locks the `Vec<Option<Word>>` contract.
- 112 agentrs tests pass; clippy clean on changed files.

The sidecar was already correct (it returns N positional entries with empty strings), so this is purely the agent binary — needs a new image to deploy.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/analyze.rs
+++ b/agentrs/src/piigate/analyze.rs
@@ -44,10 +44,13 @@ use super::presidio::{analyze_text, AnalysisParams, PresidioClient, SnapshotResu
 ///   frame, so it tracks the current screen and cannot grow unbounded.
 #[derive(Default)]
 struct LineCache {
-    // (x0, y0, x1, y1, crop_sha256) -> the line's recognized word. The position
-    // is included so a hit also requires the same on-screen placement (a line
-    // that moved is re-analyzed in its new position).
-    entries: HashMap<LineKey, Word>,
+    // (x0, y0, x1, y1, crop_sha256) -> the line's recognition outcome. The
+    // position is included so a hit also requires the same on-screen placement
+    // (a line that moved is re-analyzed in its new position). The value is an
+    // `Option<Word>`: `None` caches a line that recognized to nothing usable
+    // (blank line, icon, separator), so a stable empty line is a cache hit too
+    // and never forces a needless rec.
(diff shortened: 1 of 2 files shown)


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1566.0.0-gad703fd

PR #1567, opened 2026-06-29
revert(rdp): remove line-level OCR cache from the PII guard

racerxdl:
Reverts the line-level OCR cache (#1563) and its follow-up fix (#1566).

## Why

The cache split the single `/ocr` call into `/det` + `/rec` to re-recognize only the lines whose pixels were unchanged since the previous frame. That only wins when on-screen text is **pixel-stable** frame to frame.

On real interactive sessions (scrolling, redraws, animation) almost every line changes every frame. Live metrics from the deployed GPU agent:

| Metric | Observed | Meaning |
|--------|----------|---------|
| `lines_recognized` / `lines_total` | 151/173, 177/194 | ~85-90% cache **miss** — near-zero reuse |
| `paints_changed` / `paints_total` | 442/442, 351/351 | every paint genuinely changes pixels |
| `ocr_sent_kib` | 15-21 MB/window | band bitmap encoded + sent **twice** (`/det` then `/rec`) |
| `ocr` wall-clock | 460-1180ms | up from ~224ms pre-cache |

So on dynamic content the split design is **pure overhead** vs. the original single call: double BMP encode, double round-trip, double bytes, to do work one `/ocr` call already did. Net result is a clear latency regression — the gate felt much slower.

## What

Restores the proven single-`/ocr` pipeline. `analyze.rs` and `ocr.rs` are byte-identical to the pre-cache baseline (#1557); the sidecar is back to `/healthz` + `/ocr` only.

- Fail-closed behavior and all other guard properties unchanged.
- 94 piigate tests pass (back to the pre-cache suite); clippy clean.
- The line-cache ADR (#1562) is intentionally **kept** as a record of the approach and why it did not hold up in practice.

Reviewed: no dangling references to the removed `/det`, `/rec`, `LineCache`, `crop_hash`, `RecResult`, `DetResult`; restored path preserves fail-closed analysis.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/analyze.rs
+++ b/agentrs/src/piigate/analyze.rs
@@ -7,55 +7,56 @@
 //! indistinguishable from full-frame analysis.
 
 use std::collections::HashMap;
+use std::hash::{Hash as _, Hasher as _};
 use std::sync::{Arc, Mutex};
 
 use anyhow::Context as _;
 use async_trait::async_trait;
+use tokio::sync::Semaphore;
+use tokio_util::sync::CancellationToken;
 
-use super::bands::YBand;
+use super::bands::{split_bands, OcrChunk, YBand, DEFAULT_BAND_PADDING, MAX_CHUNK_ROWS};
 use super::metrics::LatencyAggregator;
-use super::ocr::{join_words, LineBox, OcrClient, Word};
+use super::ocr::{join_words, OcrClient, Word};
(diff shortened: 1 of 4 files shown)


sandromello:
✅ Build Completed with Success, Version=1567.0.0-gdd98a7c

EmanuelJr (approved):

PR #1568, opened 2026-06-29
fix(mcp): stop session self-destruct causing 404

luanlorenzo:
## 📝 Description

MCP sessions were being destroyed about 30 seconds after creation for clients that do not keep a standalone SSE stream open (Devin, Cursor). The next tool call then failed with `transport error: failed to send request: session terminated (404). need to re-initialize`. The symptom was "works once, fails on the next call".

Root cause: the MCP server set `ServerOptions.KeepAlive`. This makes the go-sdk send server→client pings on the **standalone** SSE stream and close the **entire session** on the first ping failure. Clients that don't hold that stream open fail every ping, so the session is evicted and every subsequent request returns 404. The ping also never travelled on the stream doing the actual work, so it never protected long-running tool calls in the first place.

## 🔗 Related Issue

N/A

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- Removed `ServerOptions.KeepAlive` from the MCP server, since it pinged the wrong stream and destroyed sessions on ping failure.
- Added an inactivity-based `SessionTimeout` (30m) on the handler. The SDK pauses this timer for every in-flight request, so active sessions and long-running waits never expire; only genuinely abandoned sessions are reclaimed, keeping memory bounded.
- Kept long-running waits warm the correct way: emit MCP progress notifications on the request's own POST stream from the `waitUntil` poll loop, gated on the client having sent a `progressToken` (no unsolicited notifications). Wired into `reviews_wait` and `sessions_wait_analysis`.
- Added a deterministic, gateway-free regression test that reproduces the 404 with a Devin-style client and proves the fix.

## 🧪 Testing

Reproduced and verified with the real go-sdk, without a gateway, Postgres, or a live client:

- `TestKeepAlive_ReproducesSessionTermination` drives a client that does not open a standalone SSE stream (`DisableStandaloneSSE=true`) against a server with `KeepAlive` set, and asserts the second call fails with `ErrSessionMissing` — the exact 404. Guards against reintroducing `KeepAlive`.
- `TestProductionConfig_SurvivesIdleGap` runs the same client against the production transport config and asserts an idle gap no longer breaks the next call.

### Test Configuration:
- **Browser(s)**: N/A
- **OS**: macOS

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass

## ✅ Checklist

### DIFF
--- a/gateway/api/mcpserver/mcpserver.go
+++ b/gateway/api/mcpserver/mcpserver.go
@@ -12,10 +12,25 @@ import (
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
-// Sends MCP-level ping requests at this interval to keep idle sessions warm.
-// Without this, long-running tool handlers (e.g. reviews_wait) sit silently on
-// the wire long enough that intermediate layers tear down the connection.
-const mcpKeepAliveInterval = 30 * time.Second
+// mcpSessionTimeout bounds how long an *idle* MCP session is retained in the
+// server's in-memory session map before it is evicted. It is an inactivity
+// timer: the SDK pauses it for the duration of every in-flight request and
+// only resets it once the last request for a session finishes (see
+// sessionInfo.startPOST/endPOST in the go-sdk). An actively used session — or
+// one parked on a long-running wait tool — therefore never expires; only a
+// genuinely abandoned session is reclaimed, which keeps memory bounded on a
+// long-lived gateway without ever tearing a live client out from under itself.
+//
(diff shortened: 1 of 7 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1568.0.0-g1aef6ad

PR #1569, opened 2026-06-29
perf+fix(rdp): faster, correctly-placed PII guard redaction

racerxdl:
Makes the agent-side RDP PII guard meaningfully faster and fixes several redaction-correctness bugs found while load-testing it against a live Notepad session (typing + Ctrl+/Ctrl- zoom). All changes were validated on the GPU agent (Tesla T4) before this PR.

### 1. fix: map PII entity offsets by code point, not bytes (correctness — the big one)
Presidio reports entity `start`/`end` as Unicode **code-point** indices, but both the agent and gateway built their word-position index with **UTF-8 byte** lengths. A single multi-byte character earlier in the OCR'd text (a smart quote `’` in "I'm", an em-dash, an accented letter, or a stray OCR glyph) drifted every later detection's offset, painting the redaction box on the **wrong line** (and eventually out of bounds) while leaving the real PII visible. Observed live: `bob@hoop.dev` redaction landing on the "on screen, like:" line.

Fixed in both engines (`chars().count()` / `utf8.RuneCountInString`), with regression tests.

### 2. fix: persist redactions until the underlying pixels change
The shadow canvas holds the true (unredacted) screen while the client holds the redacted one. A byte-identical RDP repaint of an already-redacted region (RDP resends tiles constantly) was forwarded raw, re-exposing PII the client had blanked — redaction flickered on/off. Now an active redaction set keeps blanking each region until its pixels actually change; a rect is released only when a changed band or a fresh detection overlaps it (the latter supersedes a reflow-shifted rect, preventing stale-box accumulation on zoom/scroll). New leak tests cover repaint, persistence, reflow-supersede, and release.

### 3. perf: detect on a downscaled image, recognize at full resolution (~1.7x)
Profiling showed RapidOCR's detector spends ~100ms normalizing the full frame on the CPU — more than the GPU inference — and it ignores `limit_side_len` for wide frames. We now downscale **only the detection pass** and recognize from the **full-resolution** original, so small fonts are unaffected (verified accurate down to ~8px on a T4) while OCR is ~1.7x faster. (An earlier whole-pipeline downscale was rejected because it dropped sub-10px text — a leak.)

### 4. feat: changed-patch geometry metrics (diagnostics)
Per-window `patch_tight_ratio` / `patch_fullwidth` instrumentation that quantified the OCR over-read; kept as low-cost diagnostics for future tuning.

## Testing
- 113 agentrs tests pass (incl. new char-offset, sticky-redaction leak, and persistence tests); clippy clean on touched files.
- gateway analyzer Go tests pass (incl. new rune-offset test).
- Validated end-to-end on the live T4 agent: redaction boxes land on the correct lines at all zoom levels, persist correctly, and the session is visibly faster.

Note: the char-offset fix touches the **gateway** analyzer too (same latent bug), keeping both engines consistent.

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/metrics.rs
+++ b/agentrs/src/piigate/metrics.rs
@@ -129,6 +129,17 @@ struct Window {
     // changed/total ratio means most repaints are no-ops we now skip.
     paints_total: u64,
     paints_changed: u64,
+
+    // Changed-rectangle geometry accounting (diagnostic for 2D rect-scoped
+    // OCR). For each changed patch we sum the TIGHT rect area (w*h) and the
+    // FULL-WIDTH-band area (canvas_width*h) that the current full-width band
+    // model actually OCRs. The ratio tight/fullwidth is the upper bound on the
+    // pixel-area (hence OCR cost) that horizontal scoping could save: ~1.0
+    // means RDP already sends full-width rows (scoping buys nothing); small
+    // means most changes are narrow and we are over-OCRing badly.
+    patch_tight_px: u64,
+    patch_fullwidth_px: u64,
+    patch_max_width: u32, // widest single changed patch seen this window
 }
 
 impl Window {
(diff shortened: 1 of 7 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1569.0.0-g94041f5

PR #1573, opened 2026-07-01
fix(gateway): refuse guarded sessions Presidio cannot enforce

racerxdl:
## Problem

Guardrails are enforced exclusively through Presidio in the agent, and only for connection types whose proxy actually evaluates them. Sessions on guarded connections could still open — and run **unguarded** — when no Presidio provider was configured, or when the connection type has no guardrail enforcement (mysql, mssql, mongodb).

## Changes

Fail closed at session-open with `FailedPrecondition` when a connection has guardrail rules and either:

- **no Presidio provider is configured** — new `HasGuardrailProvider()` accessor (requires `DLP_PROVIDER=mspresidio` plus both analyzer/anonymizer URLs; stricter than `HasRedactCredentials`, which is also true for GCP), or
- **the connection type has no agent-side guardrail enforcement** — new `connectionTypeSupportsGuardRails` admission check (allowed: postgres, oracledb, httpproxy, ssh, command-line).

`plain-exec` sessions remain exempt: the verb is gateway-internal (gated by a per-process secret in the auth interceptor) and only runs fixed introspection scripts, mirroring the existing DLP exemption.

Complements hoophq/libhoop#83, which makes the agent-side enforcement fail closed for the runtime "Presidio unreachable" case.

Fixes DEP-48

Automated by MisterMal

### DIFF
--- a/gateway/appconfig/appconfig.go
+++ b/gateway/appconfig/appconfig.go
@@ -437,6 +437,16 @@ func (c Config) GcpDLPJsonCredentials() string         { return c.gcpDLPJsonCred
 func (c Config) DlpProvider() string                   { return c.dlpProvider }
 func (c Config) DlpMode() string                       { return c.dlpMode }
 func (c Config) HasRedactCredentials() bool            { return c.hasRedactCredentials }
+
+// HasGuardrailProvider reports whether the server is configured with a DLP
+// provider capable of enforcing guardrails. Guardrails are evaluated exclusively
+// through MSPresidio, so this requires the mspresidio provider with both the
+// analyzer and anonymizer URLs configured. It is intentionally stricter than
+// HasRedactCredentials (which is also true for GCP), because GCP cannot enforce
+// guardrails.
+func (c Config) HasGuardrailProvider() bool {
+	return c.dlpProvider == "mspresidio" && c.msPresidioAnalyzerURL != "" && c.msPresidioAnonymizerURL != ""
+}
 func (c Config) MSPresidioAnalyzerURL() string         { return c.msPresidioAnalyzerURL }
 func (c Config) MSPresidioAnomymizerURL() string       { return c.msPresidioAnonymizerURL }
 func (c Config) PgUsername() string                    { return c.pgCred.username }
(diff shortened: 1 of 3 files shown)


gateway/appconfig/appconfig.go
```diff
@@ -437,6 +437,16 @@ func (c Config) GcpDLPJsonCredentials() string         { return c.gcpDLPJsonCred
...
+
+// HasGuardrailProvider reports whether the server is configured with a DLP
+// provider capable of enforcing guardrails. Guardrails are evaluated exclusively
+// through MSPresidio, so this requires the mspresidio provider with both the
+// analyzer and anonymizer URLs configured. It is intentionally stricter than
+// HasRedactCredentials (which is also true for GCP), because GCP cannot enforce
+// guardrails.
+func (c Config) HasGuardrailProvider() bool {
+	return c.dlpProvider == "mspresidio" && c.msPresidioAnalyzerURL != "" && c.msPresidioAnonymizerURL != ""
+}
```
matheusfrancisco, line 449: This does not mean Presidio is running; it just means the gateways are configured incorrectly.

gateway/transport/client.go
```diff
@@ -479,6 +504,37 @@ func (s *Server) processClientPacket(stream *streamclient.ProxyStream, pkt *pb.P
...
+							"contact your administrator to configure Presidio")
+				}
+				if connType := pctx.ProtoConnectionType(); !connectionTypeSupportsGuardRails(connType) {
+					logCtx.Warnf("refusing session: connection type %q does not support guardrail enforcement", connType)
+					return status.Errorf(codes.FailedPrecondition,
+						"this connection has guardrails configured, but connection type %q does not support guardrail enforcement; "+
+							"remove the guardrails from this connection or use a supported connection type", connType)
+				}
+			}
 		}
```
matheusfrancisco, line 538: This could be true, but the Presidio is not running, no?

matheusfrancisco: Because inside the libhoop, it will make an HTTP call for Presidio, and it can be configured with the wrong URI, no? And still, this won't resolve the error, no?

sandromello:
✅ Build Completed with Success, Version=1573.0.0-ga2f9aee

matheusfrancisco (approved):

gateway/appconfig/appconfig.go
racerxdl: Yeah I know, but if presidio is not running the libhoop will guard it.

PR #1574, opened 2026-07-01
fix(rdp): stop sharing one WASM parser instance across concurrent sessions

racerxdl:
## Problem (DEP-47)

The gateway intermittently crashed/wedged during active RDP usage — the whole HTTP endpoint refused connections for ~1 minute (k8s pod restart), then recovered. Reported against tryrunops.

## Root cause

`RDPSessionRecorder` shared **one global WASM parser instance across every concurrent RDP session**. The parser module keeps mutable state in its linear memory — parsed bitmaps, a bump allocator, and cross-call Fast-Path **fragment reassembly** state — and wazero module instances are not safe for concurrent calls. Overlapping sessions corrupted each other's reassembly state and allocator, producing out-of-bounds reads that crashed or wedged the gateway.

Empirically confirmed: driving a single shared instance from 8 goroutines **hangs indefinitely** under `-race`; isolated per-session instances pass cleanly.

## Fix

- **Compile once, instantiate per session** (wazero's documented pattern): the module is compiled once per process (shared `Runtime` + `CompiledModule` behind `sync.Once`); each session gets a cheap, fully isolated anonymous instance with its own linear memory and globals.
- Each recorder owns and closes its parser instance; degrades safely (no panic, no buffering) if creation fails.
- The PII gate's per-session parser stops recompiling the module per session (same semantics, less CPU).
- `InitParser` becomes a startup compile warmup; the shared `GetParser` accessor is removed.
- Hardening: all required WASM exports validated at construction; masked nil-error returns in `GetPduSize`/`Parse`/`getBitmap` now return real errors.

## Tests

- 8 concurrent per-session instances, interleaved complete + fragmented bitmap PDUs with per-session distinct geometry (cross-instance state bleed would fail assertions), under `-race`
- `GetPduSize` framing hot path, concurrent, under `-race`
- Recorder wire→bitmap pipeline end to end: tail buffering (PDU split across `RecordOutput` calls), fragment reassembly, JSONL event stream, canvas extents
- Recorder degraded mode (nil parser): no panic, idempotent Close
- Full gateway suite: 33 packages, 0 failures

## Validation note

The trigger condition is **two or more concurrent RDP sessions**. Recommended staging smoke test: open 2+ RDP sessions, use them for several minutes, confirm the gateway stays up and a recorded session replays correctly. Wire format and the `BitmapEvent`/replay schema are unchanged.

Fixes DEP-47

Automated by MisterMal

### DIFF
--- a/gateway/rdp/bitmap_parser.go
+++ b/gateway/rdp/bitmap_parser.go
@@ -1,35 +1,22 @@
 package rdp
 
 import (
-	"context"
-	"sync"
-
 	"github.com/hoophq/hoop/common/log"
 	"github.com/hoophq/hoop/gateway/rdp/parser"
 )
 
-var (
-	globalParser     *parser.Parser
-	globalParserOnce sync.Once
-	globalParserErr  error
-)
-
-// InitParser initializes the global RDP bitmap parser
-// It should be called once at application startup
+// InitParser compiles the RDP bitmap parser WASM module once at startup so the
+// first RDP session does not pay the compile cost and a broken module is
+// detected early. It does NOT create a shared parser instance: each RDP session
+// creates its own isolated parser (the WASM module keeps per-instance mutable
+// state that is unsafe to share across concurrent sessions).
+//
+// It should be called once at application startup.
 func InitParser() error {
-	globalParserOnce.Do(func() {
-		ctx := context.Background()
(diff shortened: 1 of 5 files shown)


sandromello:
✅ Build Completed with Success, Version=1574.0.0-gbfba6f3

matheusfrancisco (approved):

PR #1575, opened 2026-07-02
perf(rdp): cut per-batch round trips and hot-loop cost in the agent PII gate

racerxdl:
## What (RD-279)

Four semantics-preserving performance improvements to the agent-side RDP live redaction pipeline (`agentrs/src/piigate/`). The batch pipeline is serial, so each change directly reduces client-visible latency and raises the throughput ceiling before the fail-closed backlog kill (32MB).

### 1. OCR cache: bounded recency instead of replace-per-analyze
The old cache was cleared and repopulated on every analyze, so alternating pixel states — the canonical repaint storm, e.g. a blinking caret (A/B/A/B) — evicted A while caching B and **missed on every single frame**. The cache now retains up to 64 recent window states with generation-stamped recency eviction; blink cycles become pure hits after the first A/B pair. Each avoided miss saves an OCR round trip (~6ms GPU sidecar).

Security invariants unchanged: the key is still geometry + a hash of the exact pixel bytes (changed pixels always re-OCR), and current-generation entries are never evicted.

### 2. Presidio round-trip dedup
When a batch's OCR words equal the previous batch's in **text and geometry** (confidence deliberately excluded — Presidio never sees it), the analyzer returns the cached result instead of re-posting (saves 5–50ms/batch on static screens). Same text at a new position misses, so bounding boxes are never stale; failed calls are never cached.

### 3. Row-wise composite/convert hot loops
`composite_bitmap` and `to_rgba` rewritten from per-pixel loops (per-pixel bounds checks) to row-wise slice compare/copy:

| Function | Case | Before | After | Speedup |
|---|---|---|---|---|
| composite_bitmap | 1920x1080 identical repaint | 1.06ms | 0.19ms | 5.7x |
| composite_bitmap | 1920x1080 changed | 1.35ms | 0.13ms | 10.6x |
| to_rgba | 1920x1080 (16/24/32bpp) | ~2.0ms | ~1.2ms | 1.6–1.8x |

Equivalence tests pin both against transcriptions of the original per-pixel implementations across clipping, truncation and changed-flag cases.

### 4. Zero-copy ingest
`GatePdu` now holds the frozen `Bytes` split of the ingest tail (no per-PDU copy) and redacted-wire forwarding writes bounded slices directly. Backlog accounting documented as a logical payload cap.

**Rejected after measurement:** faster window hash (1.2x — SipHash is near memory bandwidth on target hardware).
**Out of scope:** 2D rect-scoped OCR windows (the biggest win) — pending the `patch_tight_kpx`/`patch_fullwidth_kpx` diagnostics from #1569 on real workloads.

## Tests
- 12 new tests: blink ping-pong cache hits, changed-pixels-always-re-OCR (pre-existing, still green), conf-drift dedup, geometry-shift re-call, Presidio error not cached, empty-OCR no stale detections, OcrCache eviction unit tests, composite/to_rgba equivalence oracles.
- Full agentrs suite: 123 passed, clippy clean on piigate.

Fixes RD-279

Automated by MisterMal

### DIFF
--- a/agentrs/src/piigate/analyze.rs
+++ b/agentrs/src/piigate/analyze.rs
@@ -8,7 +8,9 @@
 
 use std::collections::HashMap;
 use std::hash::{Hash as _, Hasher as _};
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
+
+use parking_lot::Mutex;
 
 use anyhow::Context as _;
 use async_trait::async_trait;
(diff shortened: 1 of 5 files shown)


sandromello:
✅ Build Completed with Success, Version=1575.0.0-gdd80f3c

matheusfrancisco (approved):

PR #1579, opened 2026-07-02
fix: build agent OCR images as root, restore hoop runtime user

racerxdl:
## Summary
The post-merge release publish of the agent OCR images failed (`Publish Agent OCR Image (CPU)`/`(GPU)`) with:

```
E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)
```

Root cause: the hoop base image ends with `USER hoop` (uid 10001), so the derived OCR Dockerfiles ran `apt-get`/`pip` setup unprivileged. The PR that introduced them was green because these images are only built in `release.yml` — `FROM` the just-published `hoophq/hoop:<tag>`, which doesn't exist at PR time — so the release run was the first build ever attempted.

## Changes
- `Dockerfile.agent-ocr` / `Dockerfile.agent-ocr-gpu`: `USER root` for the setup steps, `USER hoop` restored after — runtime privileges unchanged.
- New `.github/workflows/agent-ocr-build.yml`: PR-time build-only smoke check (path-filtered) that builds both flavors against `hoophq/hoop:latest` with the release platform matrix (CPU amd64+arm64, GPU amd64), plus a runtime assertion that the image still runs as uid 10001.

## Validation
- Local `docker build` of the CPU flavor against `hoophq/hoop:1.108.3` (arm64 — the platform that failed in CI) succeeds.
- Running the container: OCR engine becomes healthy as user `hoop` (rapidocr bundles its models in the wheel — no runtime writes) and exits cleanly.
- The new smoke workflow triggers on this PR itself and validates both flavors, including the GPU one.

Fixes DEP-52

Automated by MisterMal

### DIFF
--- a/.github/workflows/agent-ocr-build.yml
+++ b/.github/workflows/agent-ocr-build.yml
@@ -1,0 +1,93 @@
+# Build-only smoke check for the agent OCR images.
+#
+# The release workflow (release.yml) is the ONLY place these Dockerfiles are
+# otherwise built — they derive FROM the just-published hoophq/hoop:<tag>
+# image, which does not exist at PR time. That gap let a broken Dockerfile
+# reach main and fail the post-merge release publish (the base image runs as
+# the unprivileged 'hoop' user, and apt-get failed at first real build).
+#
+# This workflow closes the gap: on PRs touching the OCR image inputs it
+# builds the same Dockerfiles against the latest released base image, with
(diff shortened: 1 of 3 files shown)


sandromello:
✅ Build Completed with Success, Version=1579.0.0-gc97be2f

matheusfrancisco (approved):

PR #1580, opened 2026-07-02
fix(gateway): do not treat empty guardrail rules as configured guardrails

racerxdl:
## Summary
Regression from #1573 (DEP-48): connecting to MySQL/MSSQL/MongoDB connections **without any guardrails** failed with:

```
rpc error: code = FailedPrecondition desc = this connection has guardrails configured, but connection type "mysql" does not support guardrail enforcement; ...
```

## Root cause
- `services.GetGuardRailRulesForConnection` fabricates `"[]"` (JSON empty array) rule sets for connections without guardrails.
- `guardrails.Decode("[]")` yields an empty **non-nil** slice.
- The emptiness check in the transport layer was nil-ness based (`inputRules == nil && outputRules == nil`), so ruleless connections marshaled `{"input_rules":[],"output_rules":[]}`.
- The DEP-48 fail-closed admission check counts any non-empty payload as "guardrails configured" and refused the session.

Before #1573 this latent flaw only shipped a harmless empty payload to the agent; the fail-closed check turned it into a session refusal for every ruleless connection of an unenforceable type.

## Fix
- Judge emptiness by rule count (`len(...) == 0`), so no-rules connections produce a nil payload (the agent-side `guardrails.Validate` already no-ops on empty rule data).
- Extract the encoding into a pure `encodeGuardRailRules` function and pin the regression with unit tests (fabricated `"[]"` sets, nil/absent columns, real rules, invalid JSON).
- DEP-48 semantics are preserved: connections with real rules still produce a payload and are still refused on unenforceable types.

## Validation
- `go test ./gateway/transport/... ./gateway/guardrails/...` passes.
- New `TestEncodeGuardRailRules` covers the exact regression case.

Automated by MisterMal

### DIFF
--- a/gateway/transport/agent_guardrails_test.go
+++ b/gateway/transport/agent_guardrails_test.go
@@ -91,6 +91,80 @@ func TestConnectionTypeSupportsGuardRails(t *testing.T) {
 	}
 }
 
+func TestEncodeGuardRailRules(t *testing.T) {
+	t.Run("nil rules yield no payload", func(t *testing.T) {
+		payload, err := encodeGuardRailRules(nil)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		if payload != nil {
+			t.Fatalf("expected nil payload, got %q", string(payload))
+		}
+	})
+
+	// services.GetGuardRailRulesForConnection fabricates "[]" rule sets for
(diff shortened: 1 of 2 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1580.0.0-ga1b8bc2

PR #1585, opened 2026-07-06
fix(docker): drop ghostscript/AGPL chain from agent images

racerxdl:
## Why

A customer license scan flagged ghostscript (**AGPL**) and its GPL dependency tree in our agent container image:

```
ghostscript, libgs10, libgs10-common, libgs-common, libgs10,
poppler-data, libwmflite-0.2-7, libjbig2dec0, fonts-urw-base35
```

None of it is used by hoop. Root cause (verified inside the published `hoophq/agent-tools` image): the **only** installed package depending on ghostscript is `groff`, which we install solely so `aws help` can render man pages. Ubuntu's `groff` *Recommends* ghostscript + imagemagick + netpbm + psutils, and apt installs Recommends by default. groff only uses them for PostScript/PDF output, which nothing in the image produces. The gateway image (`hoophq/hoop`) was checked and is clean.

## What

- Install `groff` with `--no-install-recommends` in `Dockerfile.tools` and `deploy/docker-compose/Dockerfile` — drops the entire AGPL/GPL chain while keeping `aws help` fully functional (groff's hard `Depends` are untouched).
- Bump `postgresql-client-16` pin `16.10 -> 16.14`: the old version was purged from the noble archive, so `Dockerfile.tools` no longer built from main at all.

## Verification

Full `docker build -f Dockerfile.tools` on this branch, then inside the image:
- `dpkg-query` shows **none** of the flagged packages
- `aws help` renders correctly through groff
- `psql` / `mongosh` / `mysql` / `sqlcmd` / `groff` all functional

## Follow-ups

- `hoophq/agent-tools` is built/published out-of-band — it needs a rebuild with a new tag and a `FROM` bump in `Dockerfile.dev` for this to reach `hoophq/hoopdev`.
- Reviewer suggestion (LOW): once agent-tools has a build pipeline, add an assertion that the flagged packages stay absent.

Automated by MisterMal

### DIFF
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -22,7 +22,6 @@ RUN mkdir -p /app && \
     gnupg2 \
     net-tools \
     netcat-openbsd \
-    groff \
     jq \
     openssh-client \
     unzip \
@@ -38,6 +37,11 @@ RUN mkdir -p /app && \
     bzip2 \
     wget \
     redis-tools && \
+    # groff is only needed by `aws help` to render man pages to the terminal.
+    # Install it without Recommends: they pull ghostscript (AGPL), imagemagick,
+    # netpbm and psutils (GPL), which groff only uses for PostScript/PDF output
+    # that nothing in this image produces.
+    apt install -y --no-install-recommends groff && \
     ln -s /usr/bin/python3 /usr/bin/python
 
 # https://github.com/nodejs/release-keys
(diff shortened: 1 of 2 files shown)


sandromello:
✅ Build Completed with Success, Version=1585.0.0-gf1f6d86

EmanuelJr (approved):

PR #1588, opened 2026-07-06
feat: silent token renewal for tunnel sessions (DEP-24)

racerxdl:
Fixes DEP-24

## Problem

A tunnel session dies with its access token. The gateway's `X-New-Access-Token` rotation was **reactive-only** (fires after expiry, HTTP middleware only) and `hsh-tunneld` **discarded the header**, treating its token as a static credential. Net effect: users had to reopen the webapp to renew their session, and even that didn't help the tunnel — the gRPC auth interceptor has no refresh path, so new flows die at stream-open once the daemon's token copy lapses.

## Change

**Gateway** (gated by new flag `experimental.tunnel_token_renewal`, default off):
- Auth middleware rotates **still-valid** tokens within a 10-min pre-expiry window (after authn/authz), so long-lived clients always hold a token that is valid at gRPC stream-open time
- OIDC: exchanges the server-side stored refresh token (shares the singleflight refresh core with the reactive path; `TryProactiveRefreshForVerifiedSubject` documents the trust contract)
- Local auth: sliding-session re-mint with a signature-verified `auth_time` claim preserved across renewals and a **7-day absolute cap** — stolen tokens can't be kept alive indefinitely; future `auth_time` rejected at mint and verify
- Renewals logged with subject + auth method (audit requirement)

**Daemon** (`tunnel/`, works with or without the flag):
- Harvests `X-New-Access-Token` from every REST response; persists to the daemon config
- `tunnelmgr.TokenSource`: per-flow dials snapshot the current token; rotations are **epoch-guarded** so a slow response from an ended session (logout/auth-expiry/re-login) can never resurrect its token
- Renewal scheduler wakes 5 min before `exp` and probes `/api/serverinfo`; when the gateway reports the flag off it degrades to one probe just past expiry (harvests the ungated reactive OIDC refresh); jittered cadence, epoch-scoped flag cache
- **Clean failure**: 401 (server-side refresh also dead) → tunnel teardown, credential cleared, explicit re-login instruction in `hsh tunnel status` — replaces today's silent half-dead limping

## Tests

- `common/keys`: auth_time preservation, legacy iat fallback, forged-token rejection
- `gateway/idp/local`: sliding renewal, absolute cap, auth_time preservation across hops, foreign-key rejection
- `tunnel/client`: rotation harvesting, no-header no-callback, typed 401 on both endpoints
- `tunnel/cmd/hsh-tunneld`: tokenState epoch semantics incl. the logout/late-rotation race and re-login-as-other-user clobber, jwtExpiry, scheduler delays (all modes), flag-cache epoch reset

## Follow-ups

- Middleware behavior-matrix test (flag off / near-expiry local / beyond-cap / OIDC / SAML) belongs in the integration harness — needs DB-backed flag store
- Compressed-time E2E for the 24h acceptance criterion (short-lived local tokens + the standalone harness)
- 7-day local-auth cap is a constant while experimental; per-org configuration lands with the DEP-16 org settings

Two review passes: initial HIGH (token-resurrection race) fixed via epoch guarding; follow-up MEDIUM (flag cache crossing session boundaries) fixed via epoch-scoped cache. Reviewer signed off on the design after fixes.

Automated by MisterMal

### DIFF
--- a/common/featureflag/featureflag.go
+++ b/common/featureflag/featureflag.go
@@ -96,6 +96,13 @@ var catalog = map[string]Flag{
 		Stability:   StabilityExperimental,
 		Components:  []Component{ComponentAgent},
 	},
+	"experimental.tunnel_token_renewal": {
+		Name:        "experimental.tunnel_token_renewal",
+		Description: "Silently renew user access tokens before they expire so long-lived tunnel sessions (hsh-tunneld) never present an expired token: the HTTP auth middleware rotates tokens within a pre-expiry window via X-New-Access-Token (OIDC uses the stored refresh token; local auth re-mints a sliding-session JWT capped at 7 days from the original login). When off, tokens are only refreshed reactively after they expire, and local-auth tokens are never renewed.",
+		Default:     false,
+		Stability:   StabilityExperimental,
+		Components:  []Component{ComponentGateway},
+	},
 	"experimental.claude_code_vertex": {
 		Name:        "experimental.claude_code_vertex",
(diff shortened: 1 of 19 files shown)


sandromello:
✅ Build Completed with Success, Version=1588.0.0-gbb3366a

matheusfrancisco (approved):

PR #1590, opened 2026-07-06
fix(agent): preserve per-connection packet order for SSH writes (DEP-57)

racerxdl:
Fixes DEP-57

## Problem

`agent/integration` flaked in CI: `TestSSH_MultipleChannels` hung until the 10-minute suite alarm (also observed on an unrelated branch). Reproduced locally under `-race` at iteration 13 of a stress loop.

**Root cause** (production-real, not a test artifact): the recv loop dispatched every `SSHConnectionWrite` with `go processPacket(pkt)`. Goroutines are scheduled arbitrarily, so an exec request could reach libhoop's SSH proxy **before the OpenChannel it belongs to**; the proxy drops requests for unknown channels (`"unable to obtain channel"`) and the client waits forever for an exit-status. The `connWriteLocks` per-connection mutex could not prevent this — a mutex prevents interleaving, not reordering.

## Change

- `SSHConnectionWrite` now goes through a per-(session, connection) FIFO drained by a single worker — the exact dispatch model `HttpProxyConnectionWrite` already uses. Recv loop never blocks; connections proceed independently (preserving the goal of async dispatch); packets for one connection keep strict stream order. `SessionClose` stays concurrently dispatched.
- The queue type is extracted to `packetqueue.go`, shared by both protocols, and **now bounded** (4096 pkts / 64 MiB per connection): a wedged upstream fails the SSH session — or for HTTP just that connection, siblings survive — with an explicit error instead of unbounded agent memory growth.
- `connWriteLocks` + `sshFlightGroup` removed: both existed to mitigate unordered dispatch; redundant (and insufficient) once handlers are serialized per connection.
- **Harness**: `SessionClose` used to land in an unread demux channel — a failed session (e.g. OSS `_libhoop` stub) hung tests for the full timeout. `DialPipedSSH` now watches session close and tears the piped transport down: the stub case goes from a **10-minute hang to a 4-second failure** with the close reason printed. Demux records close reasons; `SessionCloseChan` now honours its documented contract for late subscribers.

## Verification

- Baseline: hang reproduced at stress iteration 13; **with fix: 30/30 + 12/12** stress iterations pass under `-race` (enterprise libhoop)
- Full `agent/integration` suite passes except `TestSSH_DirectTCPIP_PortForward` and `TestMySQL_CloseKillsRunningQuery`, which fail **identically on unmodified baseline** (macOS-Docker env issues; green in CI)
- Transport suite, controller unit tests (incl. 6 new packetQueue tests: ordering, respawn, no-loss under race, both overflow bounds, blocked-handler serialization), vet + gofmt clean

## Notes for reviewers

- **HTTP queue overflow is connection-fatal, SSH overflow is session-fatal** — each matches its subsystem's existing failure contract.
- Pre-existing, unchanged: `processSSHProtocol` holds the session RLock across the upstream dial, so a slow dial delays `SessionClose` for that session by at most libhoop's dial timeout (`net.DialTimeout` + 2s handshake cap). Restructuring that is a candidate follow-up, deliberately out of scope here.
- The MongoDB flake instance from the ticket looked like container churn, not this race; if it recurs, it needs its own investigation.

Two review passes; final verdict "ready to commit" after the queue bound + coverage additions.

Automated by MisterMal


sandromello:
✅ Build Completed with Success, Version=1590.0.0-g865ec9d

racerxdl:
CI flaked on a second, distinct issue: `TestMongoDB_BadCredentials` — the mongo container was healthy but the readiness ping dialed `[::1]` (Docker's port proxy isn't IPv6-bound on the runner; testcontainers' `Host()` returns `localhost`). Pushed a hardening commit: all five container harnesses now pin `localhost` → `127.0.0.1` via a shared `ContainerHost` helper, and the mongo readiness deadline is 120s. This is the same "mongo container churn" class as the RD-39-branch flake noted on the ticket — now actually root-caused as the IPv6 dial.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1590.0.0-g670fd30

matheusfrancisco (approved):

PR #1592, opened 2026-07-08
perf(ocr): fp16 bucketed recognition and en model for the RDP PII guard

racerxdl:
## Summary

The RDP PII guard's redact latency is ~88% OCR, and a chunk of that was the GPU being used wrong, not the GPU working. This PR productizes the optimizations validated on the production T4 (`hoop-pii-agent`):

**Root cause found while benchmarking an A100** (which was *3.5x slower* than the T4 out of the box): ONNXRuntime 1.20's cuDNN frontend caches only the **last input shape per conv node**. RapidOCR's variable-width recognition batches cycle shapes within every frame → graph rebuild + EXHAUSTIVE retune, every frame, forever (~60ms/frame T4, ~1.2s/frame A100 — the tax grows with the GPU's kernel catalog). `HEURISTIC` search is **not** a fix: it crashes on Turing under concurrent load (`CUDNN_FE failure 11`) — documented in-code so nobody retries it.

## Changes

Two env knobs, **both defaulting to today's exact behavior** (`fp32`/`ch`) — existing deployments see zero change:

- `OCR_REC_PRECISION=fp16` — recognition through `BucketRec`: one fixed-shape `InferenceSession` per width bucket on an fp16-converted model (fp32 I/O). Each session sees exactly one shape → conv graphs tuned once per process lifetime, EXHAUSTIVE retained. Padding matches the stock path's post-normalize zero-fill exactly (`rapidocr resize_norm_img`), keeping text output **byte-identical**.
- `OCR_REC_LANG=en` — English-only rec model (~96-class head vs 6,625). Faster + more accurate on Latin screens. **CJK caveat**: the en dictionary cannot read CJK text → CJK PII would be missed; deployments serving CJK desktops must keep `ch` (documented in code, Dockerfile, env.sample).

- `Dockerfile.agent-ocr-gpu`: prefetches the en model + converts both rec models to fp16 **at build time** (throwaway venv, existence-checked) — preserves the nothing-downloaded-after-build property.
- GCP demo compose flips `fp16`+`en` on (en-US deployment), overridable.
- `bucket_rec.py` is a standalone module with injected sessions/decoder; `test_bucket_rec.py` (8 tests) covers batching, cross-bucket index order, partial-batch constant shapes, squeeze handling, and the padding semantics.

## Results (T4, 1080p frame)

| | baseline | this PR (fp16+en) |
|---|---|---|
| dense frame (46 boxes) | 406ms | **227ms (1.8x)** |
| light frame (23 boxes) | 331ms | **228ms** |

Projected on the production pipeline (from 3,798 real batches of session metrics): batch total 318ms → ~200ms wavg. PII lines byte-identical in all validation runs (`email: … ssn 123-45-6789` exact), det box geometry untouched.

## Validation

- 8/8 unit tests pass in the production container
- The exact branch files were smoke-tested live on the T4 (single worker, port 18869): healthz reports knobs, PII byte-exact, live service unaffected
- Padding semantics verified against RapidOCR source (`ch_ppocr_rec/main.py` zero-fills post-normalize) and locked by a unit test

## Not in this PR (follow-ups)

- Tight-patch OCR scoping in agentrs (production metrics show full-width bands OCR ~20x the changed pixels — biggest remaining lever)
- TensorRT EP (needs libnvinfer in the image; bucket sessions make shapes static, ideal fit)
- `OCR_DET_DOWNSCALE=0.5` (needs a small-font recall study)

Automated by MisterMal

### DIFF
--- a/Dockerfile.agent-ocr-gpu
+++ b/Dockerfile.agent-ocr-gpu
@@ -90,8 +90,11 @@ from rapidocr import RapidOCR; \
 from rapidocr.utils.typings import LangRec; \
 RapidOCR(params={'Rec.lang_type': LangRec.EN})" && \
     python3 -m venv /tmp/convert-venv && \
+    # numpy must stay on the 1.x line: onnxconverter-common 1.14.0 calls
+    # np.fromstring, which NumPy 2 removed (build fails with "The binary
+    # mode of fromstring is removed").
     /tmp/convert-venv/bin/pip install --no-cache-dir \
-        "onnx==1.17.0" "onnxconverter-common==1.14.0" && \
+        "onnx==1.17.0" "onnxconverter-common==1.14.0" "numpy==1.26.4" && \
     mkdir -p /opt/ocr/models && \
     MODELS_DIR=$(/opt/ocr-venv/bin/python -c "\
 import rapidocr, pathlib; \
--- a/Dockerfile.agent-ocr-gpu
+++ b/Dockerfile.agent-ocr-gpu
@@ -76,7 +76,43 @@ RUN mkdir -p /opt/cuda-libs && \
     test -e /opt/cuda-libs/libcudart.so.12 && test -e /opt/cuda-libs/libcublasLt.so.12
 ENV LD_LIBRARY_PATH="/opt/cuda-libs"
 
+# Prefetch the en recognition model (OCR_REC_LANG=en) and produce the fp16
+# conversions (OCR_REC_PRECISION=fp16) AT BUILD TIME, preserving the image's
+# "nothing is downloaded or written after build" property. The converter
+# packages live in a throwaway venv so the runtime venv stays exactly the
+# pinned dependency set above. Instantiating RapidOCR with LangRec.EN is the
+# supported way to fetch the en model — it lands next to the bundled ch
+# models inside the rapidocr wheel and doubles as a wheel sanity check.
+# keep_io_types=True keeps fp32 inputs/outputs, so the fp16 models are
+# drop-in for the same pre/postprocessing code paths.
+RUN /opt/ocr-venv/bin/python -c "\
+from rapidocr import RapidOCR; \
+from rapidocr.utils.typings import LangRec; \
+RapidOCR(params={'Rec.lang_type': LangRec.EN})" && \
+    python3 -m venv /tmp/convert-venv && \
+    /tmp/convert-venv/bin/pip install --no-cache-dir \
+        "onnx==1.17.0" "onnxconverter-common==1.14.0" && \
+    mkdir -p /opt/ocr/models && \
+    MODELS_DIR=$(/opt/ocr-venv/bin/python -c "\
+import rapidocr, pathlib; \
+print(pathlib.Path(rapidocr.__file__).parent / 'models')") && \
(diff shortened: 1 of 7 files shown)


sandromello:
### :rotating_light: Forbidden licenses detected in preview images

Trivy found AGPL/SSPL-licensed packages. These are rejected by
customer compliance scanners and must not ship in our images.

| Image | Package | License |
|-------|---------|---------|
| `hoophq/hoopdev` | `fonts-urw-base35` | `AGPL-3` |
| `hoophq/hoopdev` | `fonts-urw-base35` | `AGPL-3 with Font exception` |
| `hoophq/hoopdev` | `ghostscript` | `AGPL-3` |
| `hoophq/hoopdev` | `ghostscript` | `AGPL-3+` |
| `hoophq/hoopdev` | `ghostscript` | `AGPL-3+ with font exception` |
| `hoophq/hoopdev` | `libgs-common` | `AGPL-3` |
| `hoophq/hoopdev` | `libgs-common` | `AGPL-3+` |
| `hoophq/hoopdev` | `libgs-common` | `AGPL-3+ with font exception` |
| `hoophq/hoopdev` | `libgs10` | `AGPL-3` |
| `hoophq/hoopdev` | `libgs10` | `AGPL-3+` |
| `hoophq/hoopdev` | `libgs10` | `AGPL-3+ with font exception` |
| `hoophq/hoopdev` | `libgs10-common` | `AGPL-3` |
| `hoophq/hoopdev` | `libgs10-common` | `AGPL-3+` |
| `hoophq/hoopdev` | `libgs10-common` | `AGPL-3+ with font exception` |
| `hoophq/hoopdev` | `libjbig2dec0` | `AGPL-3+` |
| `hoophq/hoopdev` | `libwmflite-0.2-7` | `AGPL-3 with Font exception` |
| `hoophq/hoopdev` | `poppler-data` | `AGPL-3+` |

Most likely cause: a new apt package whose `Recommends` pulls
them in — install it with `--no-install-recommends` (see the
groff/ghostscript case in the Dockerfile.tools comments).

sandromello:
✅ Build Completed with Success, Version=1592.0.0-ge7af549

sandromello:
✅ Build Completed with Success, Version=1592.0.0-g6c5fa4c

racerxdl:
Follow-up commit from live A100 deployment: **BucketDet** — per-bucket-shape detection sessions. Production bands exposed that det pays the same cuDNN-FE re-tune tax rec did (~500ms/shape-change on A100; invisible in the original fixed-frame benchmarks). Batches stretched to ~1s, which caused real redaction leaks via mid-paint batch sealing (rows split across batches are unreadable per-half). With BucketDet: bands 26-48ms, batch totals 14-78ms, leak confirmed gone on the live agent. Includes concurrency audit w/ rapidocr source citations, 6 new unit tests + a live 8-thread determinism test against the real detector (19 tests total, all passing in the production container). Also note for reviewers: `OCR_REC_LANG=en` was reverted to `ch` on the demo deployment — en's spacing variance broke NER-based (name) detections that regex entities survive; en needs NER-entity validation before production use (the PR's default remains ch).

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1592.0.0-g474a13c

EmanuelJr (approved):

PR #1606, opened 2026-07-14
feat(agent): RESULT_METADATA connection env for MySQL driver exec

racerxdl:
Implements [DEP-53](https://linear.app/hoophq/issue/DEP-53) (agent wiring; core in hoophq/libhoop#91): surfaces row-count / rows-affected feedback for MySQL on the driver-based exec path (`experimental.db_exec_driver`).

## What changed

- New optional MySQL connection env `RESULT_METADATA` controlling the interactive-style feedback lines (`N rows in set`, `Empty set`, `Query OK, N rows affected`) on driver exec:
  - unset / `on` / `true` / `1` → metadata on (default)
  - `off` / `false` / `0` → clean batch TSV output (previous behavior)
  - anything else → fail-fast connection error (mirrors SSLMODE validation)
- OSS `_libhoop` stub gains the `OptKeyResultMetadata` const so the OSS build compiles (driver path remains enterprise-only).

No migrations, no API changes; rides the existing `experimental.db_exec_driver` feature flag.

## Customer note

OneSpan (Intercom) can get metadata **today** on the default CLI path by customizing the connection command to `mysql -vv ...` — this PR makes it first-class on the driver path.

Automated by MisterMal

### DIFF
--- a/_libhoop/agent/dbexec/dbexec.go
+++ b/_libhoop/agent/dbexec/dbexec.go
@@ -29,3 +29,9 @@ const (
 	OptKeyInsecure    = "insecure"
 	OptKeyServiceName = "service_name"
 )
+
+// OptKeyResultMetadata carries the connection-level RESULT_METADATA setting
+// through the libhoop opts map. The agent sets it to "off" when the connection
+// disables the MySQL interactive-style result feedback lines; any other value
+// (including absent) leaves them on.
+const OptKeyResultMetadata = "result_metadata"
--- a/agent/controller/agent.go
+++ b/agent/controller/agent.go
@@ -124,6 +124,12 @@ type (
 		kubernetesInsecureSkipVerify bool
 
 		experimentalRedactRows string
+
+		// resultMetadata carries the connection's RESULT_METADATA setting for
+		// the driver-based SQL exec path. Normalized to "on"/"off" by
+		// parseConnectionEnvVars; only MySQL consumes it (the other engines'
+		// CLIs already print result metadata in batch mode).
+		resultMetadata string
 	}
 	ioMetricFlush struct {
 		client pb.ClientTransport
@@ -807,6 +813,19 @@ func parseConnectionEnvVars(envVars map[string]any, connType pb.ConnectionType) 
 		}
 		if env.host == "" || env.pass == "" || env.user == "" {
 			return nil, errors.New("missing required secrets for mysql connection [HOST, USER, PASS]")
+		}
+		// RESULT_METADATA toggles the interactive-style result feedback lines
+		// ("N rows in set", "Query OK, N rows affected") on the driver-based
+		// exec path. Defaults to on; "off" keeps the clean batch TSV output
+		// for scripting consumers.
+		switch strings.ToLower(envVarS.Getenv("RESULT_METADATA")) {
+		case "", "on", "true", "1":
+			env.resultMetadata = "on"
+		case "off", "false", "0":
+			env.resultMetadata = "off"
+		default:
+			return nil, fmt.Errorf("wrong option (%q) for RESULT_METADATA, accept only: %v",
+				envVarS.Getenv("RESULT_METADATA"), []string{"on", "off", "true", "false", "1", "0"})
 		}
 	case pb.ConnectionTypeMSSQL:
 		if env.port == "" {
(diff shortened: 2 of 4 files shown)


matheusfrancisco (commented):
Just to make sure, we never had this in the MySQL execution. Also, this will kinda break the tabular view on the web app, no?

matheusfrancisco:
I know when this drive is enabled, the Oracle tabular view/ and schema pull will break  (this is a map improvement).

sandromello:
✅ Build Completed with Success, Version=1606.0.0-g88bf18f

racerxdl:
@matheusfrancisco good questions — checked both carefully:

**"we never had this in the MySQL execution"** — correct, and that asymmetry is the point of the PR. The `mysql` client suppresses `N rows in set` / `Query OK` in batch (piped) mode, and the driver path mirrored that byte-for-byte. Postgres, however, has **always** emitted result metadata on our default path: `psql -A -F'\t' -P pager=off` prints the `(N rows)` footer and command tags (`UPDATE 2`, `INSERT 0 1`) — there's no `-t/tuples_only` in the default command, and the editor's `\set QUIET on` preamble only silences meta-command chatter, not footers. So this brings MySQL to the parity Postgres already had (which is what the customer was missing).

**"will it break the tabular view?"** — No, verified against the grid code:

- The tabular parser (`log_area/main.cljs` → PapaParse → `ag_grid_table.cljs/normalize-row-data`) pads short rows with empty cells. A trailing `3 rows in set` renders as one cosmetic trailing row — exactly like the `(3 rows)` footer that **every Postgres exec already shows in the grid today**. No crash, no misalignment of real data rows.
- Non-SELECT: MySQL goes from "no output" to `Query OK, N rows affected` — same shape as Postgres's `UPDATE 2` command tag today, handled the same way.

So it's the same class of output Postgres has always produced, not a new breakage. If we want the grid to hide these lines, the right fix is a webapp-side strip covering **both** engines (`(N rows)`, `rows in set`, `Query OK`, `Empty set`) — there's precedent with `clean-postgres-script` / `clean-mssql-script`; happy to file a follow-up for that.

Also worth noting it's per-connection opt-out: `RESULT_METADATA=off` restores the clean batch TSV.

On Oracle: agreed — the driver-path tabular/schema gap there is pre-existing and tracked as the map improvement, unrelated to this env.

Re: the Qodo `.env.sample` finding — false positive. `RESULT_METADATA` is a **per-connection** env (lives in the connection's env vars, like `SSLMODE`), not a gateway/agent process env; `.env.sample` documents process-level config only (`SSLMODE` isn't there either).

Automated by MisterMal

matheusfrancisco:
qodo is this sound right it ? "we never had this in the MySQL execution" — correct, and that asymmetry is the point of the PR. The mysql client suppresses N rows in set / Query OK in batch (piped) mode, and the driver path mirrored that byte-for-byte. Postgres, however, has always emitted result metadata on our default path: psql -A -F'\t' -P pager=off prints the (N rows) footer and command tags (UPDATE 2, INSERT 0 1) — there's no -t/tuples_only in the default command, and the editor's \set QUIET on preamble only silences meta-command chatter, not footers. So this brings MySQL to the parity Postgres already had (which is what the customer was missing).

"will it break the tabular view?" — No, verified against the grid code:

The tabular parser (log_area/main.cljs → PapaParse → ag_grid_table.cljs/normalize-row-data) pads short rows with empty cells. A trailing 3 rows in set renders as one cosmetic trailing row — exactly like the (3 rows) footer that every Postgres exec already shows in the grid today. No crash, no misalignment of real data rows.
Non-SELECT: MySQL goes from "no output" to Query OK, N rows affected — same shape as Postgres's UPDATE 2 command tag today, handled the same way.
So it's the same class of output Postgres has always produced, not a new breakage. If we want the grid to hide these lines, the right fix is a webapp-side strip covering both engines ((N rows), rows in set, Query OK, Empty set) — there's precedent with clean-postgres-script / clean-mssql-script; happy to file a follow-up for that.

Also worth noting it's per-connection opt-out: RESULT_METADATA=off restores the clean batch TSV.

On Oracle: agreed — the driver-path tabular/schema gap there is pre-existing and tracked as the map improvement, unrelated to this env.

Re: the Qodo .env.sample finding — false positive. RESULT_METADATA is a per-connection env (lives in the connection's env vars, like SSLMODE), not a gateway/agent process env; .env.sample documents process-level config only (SSLMODE isn't there either).

matheusfrancisco (approved):

PR #1612, opened 2026-07-17
fix(rdp): render web client at native resolution on HiDPI displays

racerxdl:
Fixes the blurry RDP web client on HiDPI/Retina displays, reported during the demo benchmark review.

## Root cause

The canvas backing store was sized in CSS pixels (`window.innerWidth`) while CSS-stretched to 100% of the window — on a DPR=2 display the browser upscales every frame 2x, blurring the session even without any resize. The IronRDP `desktopSize` follows `canvas.width/height`, so the server was also rendering at half the physical resolution.

## What changed

- Backing store now sized in physical pixels (`viewport x devicePixelRatio`) → pixel-perfect rendering; the RDP server renders at full native resolution.
- **Clamped to 4096px per dimension**: the PII guard's shadow canvas (`maxCanvasDim` / `MAX_CANVAS_DIM`) is fail-open beyond 4096, so oversized displays (5K @2x) scale down proportionally instead of silently bypassing redaction (partial mitigation for DEP-70).
- Inputs sanitized (non-finite/zero viewport or DPR) and final dims clamped to [1, 4096].
- Fixed no-op `document.body.overflow` assignment (missing `.style`).
- Connect/error status text scales by the actual backing-store ratio.

Mouse coordinates need no changes — `getMousePos` already scales by `canvas.width / rect.width`.

## Notes

- Remote UI renders sharp but ~2x smaller on HiDPI; proper DPI scaling requires advertising `desktopScaleFactor` at connect, which ironrdp-web doesn't expose yet — tracked in DEP-72 (upstream PR to IronRDP).
- PII-gated sessions now carry ~4x pixels through the OCR pipeline (relevant to DEP-55 capacity math). Shipping ungated per discussion.

Automated by MisterMal

### DIFF
--- a/webapp/resources/public/rdpclient/index.js
+++ b/webapp/resources/public/rdpclient/index.js
@@ -51,8 +51,13 @@ async function initializeApp(rdpCredential) {
     // shadow canvas bound (maxCanvasDim in gateway/rdp/piigate.go and
     // MAX_CANVAS_DIM in agentrs/src/piigate/canvas.rs, both 4096). Beyond
     // that bound the guard skips redaction FAIL-OPEN, so oversized displays
-    // are scaled down proportionally instead (slightly soft, never
-    // unredacted). Keep the three values in lockstep.
+    // are scaled down proportionally instead. Keep the three values in
+    // lockstep.
+    //
+    // This clamp is best-effort: it bounds the size *requested* by this
+    // client, but the server may still negotiate a different desktop size.
+    // Authoritative enforcement at the gateway/agent boundary is tracked in
+    // DEP-70.
     const MAX_DESKTOP_DIM = 4096;
     const viewportWidth = Number.isFinite(window.innerWidth)
         ? Math.max(1, Math.floor(window.innerWidth)) : 1;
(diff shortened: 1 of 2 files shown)


sandromello:
✅ Build Completed with Success, Version=1612.0.0-g7a844ad

racerxdl:
Addressed the Qodo findings:

**2. Status text 0px** — fixed: font is now clamped to `[10, 200]px`.

**1. Clamp can be overwritten (resizeObservable)** — acknowledged, but intentionally not clamped there. That handler mirrors the **server-negotiated** desktop size into the render target; if a server negotiates >4096, the oversized frames are already on the wire, and shrinking our canvas would only corrupt rendering (misaligned blits) while providing zero protection — the PII guard skips oversized bitmaps server-side regardless of what the client canvas does. The client-side clamp is best-effort request bounding; **authoritative enforcement belongs at the gateway/agent boundary and is tracked in [DEP-70](https://linear.app/hoophq/issue/DEP-70)** (clamp or fail-closed reject in the connect sequence when the PII gate is active). The security-contract comment was reworded to make that scope explicit instead of overclaiming.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1612.0.0-g14e20c4

EmanuelJr (approved):

PR #1619, opened 2026-07-21
ci: add GitHub→Linear issue sync

feraio:
Adds a workflow that creates a Linear card (Support → Triage) whenever a new issue is opened here, and comments the Linear link back. One-way, creation-only. Uses the org-level `LINEAR_API_KEY` secret.

### DIFF
--- a/.github/workflows/linear-sync.yml
+++ b/.github/workflows/linear-sync.yml
@@ -1,0 +1,81 @@
+name: Sync new issues to Linear
+
+# Fires whenever a new GitHub issue is opened in this repo and creates a
+# matching card in the Linear "Support" team (routed to Triage for review).
+# Requires an org- or repo-level secret named LINEAR_API_KEY.
+
+on:
+  issues:
+    types: [opened]
+
+permissions:
+  issues: write # needed to comment the Linear link back on the GitHub issue
+
+jobs:
+  create-linear-card:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


feraio:
Thanks @qodo-code-review, useful pass. Pushed fixes for the two reliability findings; leaving the third intentionally.

**#2 (API/response handling)** ✅ Fixed. The Linear call now retries transient failures (`--retry 3 --retry-all-errors --connect-timeout 10 --max-time 30`), checks the HTTP status explicitly, and inspects the GraphQL `.errors` field (Linear returns HTTP 200 even on GraphQL errors). Added null guards on `issue.url`/`identifier` so we never post a "null" backlink.

**#3 (idempotency)** ✅ Fixed. Before creating, the job checks the issue comments for the existing `Synced to Linear:` marker and no-ops on re-run, so manual re-runs no longer create duplicate cards or comments. If card creation succeeds but the backlink comment fails, we surface the Linear URL via a warning instead of silently orphaning the card. Verified live: re-running the same job logged "already synced… skipping" and produced no duplicate.

**#1 (unrestricted creation)** Intentionally not gating on author association. This is a Support intake flow, so external bug reports are exactly what we want to capture. An OWNER/MEMBER/COLLABORATOR gate would drop the reports that matter most. Triage is the human buffer here: anything spammy lands there and gets canceled before it enters the workflow.

p3rotto (approved):

PR #1624, opened 2026-07-22
fix: keep reviewed executions tracked past the 50s API timeout

rogefm:
## 📝 Description

When a reviewed execution takes longer than the gateway's 50s synchronous wait, the API returns `202` and the execution keeps running in the agent — but the review was flipped to a terminal `UNKNOWN` status (blocking any retry with a `403`) and the webapp lost track of the run: the Execute spinner never cleared, and reopening the session modal showed a clickable Execute button while the query was still running.

This PR is an improvement proposal based on those bugs: it makes the **session the source of truth for the execution outcome** and keeps the review status and the UI derived from it.

## 📣 User-facing impact

Reviewed executions that take longer than 50 seconds now show an "Execution in progress" indicator in the session modal — surviving modal close and page reload — and render the result when they finish, instead of a stuck spinner and a re-clickable Execute button.

## 🔗 Related Issue

Fixes EVL-84 EVL-91

- [EVL-84 — Execute button comes back after reopening the session modal](https://linear.app/hoophq/issue/EVL-84/botao-de-execute-volta-apos-reabrir-modal-da-sessao)
- [EVL-91 — Batch execution hangs forever on one target](https://linear.app/hoophq/issue/EVL-91/batch-execution-hangs-forever-on-one-target)

## 🚀 Type of Change

## 📋 Changes Made

**Gateway**
- The 50s exec timeout no longer flips the review to `UNKNOWN` — it stays `PROCESSING` while the agent runs (`run-exec.go`, MCP `tools_reviews.go`).
- Reviews settle as `EXECUTED` when their session finishes (audit plugin on session close), plus a startup reconciliation for stale `PROCESSING`/`UNKNOWN` reviews whose session is already done.
- Reviewed sessions flip from `ready` to `open` when the execution starts, so the session status reflects the in-flight run.
- `UNKNOWN` is no longer produced anywhere; it is kept as a read-only legacy value (no API or DB enum change).

**Webapp**
- Execution state moved from modal-local state to app-db keyed by session id, so it survives closing/reopening the modal.
- `202` is detected by the response body (`output_status: "running"`): the modal shows "Execution in progress", polls `GET /sessions/:id` until `done` and renders the result. After a page reload, the `PROCESSING` review status resumes the tracking.
- The Execute button only renders for `APPROVED` reviews with no execution in flight — a consumed review can never be re-triggered from the UI.
- Fixed the failure-handler arity bug that swallowed every exec error into a generic snackbar.

## 🧪 Testing

Manual end-to-end on a reviewed connection with `sleep 90 && echo finished`:
1. Approve → Execute → "Execution in progress" appears immediately and persists across modal close/reopen and page reload.
2. At ~50s the gateway returns `202`; at ~90s polling detects `done` and the result renders; the review ends as `EXECUTED` in the DB and a retry via API returns `403` (one-shot semantics preserved).
3. Fast path (`sleep 5`) and failure path (agent down) behave correctly.
4. Startup sweep: a stale `PROCESSING`/`UNKNOWN` review with a done session is reconciled on gateway boot.

### Tests performed:
- [x] Unit tests pass (`go test` on `models`, `api/session`, `api/mcpserver`, `transport/plugins/audit`)
- [x] Manual testing completed (shadow-cljs compiles with 0 warnings)

## 📄 Additional Notes

The underlying design issue is that the review status doubles as an execution tracker (`PROCESSING`/`EXECUTED`/`UNKNOWN`) when its real job is the approval decision. The proper long-term fix is a refactor that keeps reviews to `PENDING → APPROVED/REJECTED/REVOKED`, derives "consumed" from the session, and moves the one-shot execution claim to the session itself. That touches the gRPC interceptor, review plugin, Slack/webhooks and the public API contract, so it is deliberately out of scope here — this PR is the minimal, backward-compatible step in that direction.

### DIFF
--- a/gateway/api/mcpserver/tools_reviews.go
+++ b/gateway/api/mcpserver/tools_reviews.go
@@ -172,12 +172,11 @@ func reviewsExecuteHandler(ctx context.Context, _ *mcp.CallToolRequest, args rev
 	timeoutCtx, cancelFn := context.WithTimeout(context.Background(), 50*time.Second)
 	defer cancelFn()
 
-	reviewStatus := models.ReviewStatusExecuted
-	defer func() {
-		if err := models.UpdateReviewStatus(review.OrgID, review.ID, reviewStatus); err != nil {
-			log.With("sid", sid).Warnf("failed updating review to status=%v, err=%v", reviewStatus, err)
+	setReviewStatus := func(status models.ReviewStatusType) {
+		if err := models.UpdateReviewStatus(review.OrgID, review.ID, status); err != nil {
+			log.With("sid", sid).Warnf("failed updating review to status=%v, err=%v", status, err)
 		}
-	}()
+	}
 
 	select {
 	case resp := <-respCh:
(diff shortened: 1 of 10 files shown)


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1624.0.0-g1580c8e

sandromello:
✅ Build Completed with Success, Version=1624.0.0-g7c27487

PR #1621, opened 2026-07-21
DEP-74 Connection type "mssql" and all proxies does not support guardrail enforcement;

matheusfrancisco:
## 📝 Description

   Moves DEP-48 guardrail fail-closed enforcement from the gateway's session-open admission check to the agent, at the point of enforcement.

   The gateway refused guarded sessions based on a hardcoded connection-type list (`connectionTypeSupportsGuardRails`) — a duplicate of knowledge only the agent-side proxies have. It drifted twice:
 MSSQL Web Exec needed a special-case exemption (`sessionSupportsGuardRails`), and it caused the mysql/mongodb web terminal regression fixed in #1611. Each drift is either a false refusal of a working
 path or a guarded session silently running unguarded.

   Now the gateway always encodes and ships guardrail rules to the agent, and the agent is the single authority on whether it can enforce them: libhoop refuses construction of native proxies with no
 guardrail evaluation path (mysql, mssql, mongodb, ssm), and the raw TCP relay refuses guarded sessions directly.

   | Path | Guarded session |
   |---|---|
   | postgres, oracle, ssh, http, exec / web terminal | enforced (localguardrails in-process; Presidio not required) |
   | mysql, mssql, mongodb, ssm (native), raw tcp | refused at the agent, fail closed |
   | plain-exec (gateway-internal verb) | exempt, unchanged (same as DLP redaction) |
   Companion PR: hoophq/libhoop (constructor gate + `CheckGuardRailEnforcement`).

   ## 📣 User-facing impact

   Connections with guardrails configured no longer get wrongly refused on paths the agent can enforce (e.g. web terminal on MySQL/MongoDB). Native protocol sessions that cannot enforce guardrails are
 still refused with a clear error instead of running unprotected.

   ## 🔗 Related Issue

   Fixes DEP-48 (follow-up to #1611)
libhoop PR first https://github.com/hoophq/libhoop/pull/97
   ## 🚀 Type of Change

   - [x] 🐛 Bug fix (non-breaking change which fixes an issue)
   - [ ] ✨ New feature (non-breaking change which adds functionality)
   - [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
   - [ ] 📚 Documentation update
   - [ ] 🎨 Style/UI update
   - [ ] ♻️ Code refactor
   - [ ] ⚡ Performance improvement
   - [ ] ✅ Test update
   - [ ] 🔧 Build configuration change
   - [ ] 🧹 Chore

   ## 📋 Changes Made

[...]

PR #1628, opened 2026-07-23
feat(mssql): native guardrails support

racerxdl:
## Summary

Adds **native MSSQL (TDS) guardrails** so guarded MSSQL connections work over `hoop connect` (not just Web Exec). Previously the gateway fail-closed *refused* guarded native MSSQL sessions because the proxy could not evaluate rules; this wires enforcement end to end.

**Closes DEP-69.** Requires the paired libhoop enforcement PR **hoophq/libhoop#96** (the TDS statement extractor + proxy live there) — **merge #96 first** (this PR's Integration Test checks out libhoop `main`).

> Recreated on a clean branch name to supersede the previous PR. The prior review discussion (sandromello) lives on the superseded PR, linked in the timeline.

## Design — libhoop is the sole guardrail authority

Reconciled onto the DEP-74 delegate model (#1621): **libhoop enforces and fails closed at proxy construction**; the gateway performs **no** MSSQL admission, capability gating, or feature flag. Native MSSQL guardrails activate purely when the connection has guardrail rules configured — exactly like Postgres/MySQL. Version skew stays safe because an older libhoop refuses guarded MSSQL sessions rather than running them unguarded.

## Changes

- **Agent** (`agent/controller/mssql.go`): forwards the connection's `guard_rail_rules` to the libhoop MSSQL proxy.
- **Gateway** (`gateway/transport/client.go`): ships the connection's configured rules as-is; no admission/flag gate.
- **Gateway** (`gateway/transport/agent.go`): `MSSQLConnectionWrite` persists guardrail violation metadata (a native block keeps the session open, so it can't ride `SessionClose`). Postgres path unchanged.

## Scope

- ✅ **Input guardrails** for SQLBatch and RPC (`sp_executesql` / `sp_prepare` / `sp_prepexec` / cursor procs).
- ⛔ **Output data masking** (result-set redaction) intentionally **out of scope** — follow-up. libhoop **fails closed** if a connection has output rules (they would otherwise run silently unenforced).

## Testing

- `agent/integration/mssql_test.go` (real SQL Server 2022 via testcontainers): blocks a violating SQLBatch and a violating RPC `sp_executesql`, blocks the dynamic-exec parameter bypass, allows compliant queries on both paths, verifies the session survives a block. Uses the real gateway→agent guardrail JSON shape.
- Gateway unit + integration suites green, built against both the OSS `_libhoop` stub and the real libhoop.

## Release label

`minor` — new, backwards-compatible feature.

Automated by MisterMal


racerxdl:
Context for reviewers — this PR supersedes the previous one (closed, linked in the timeline), recreated on a clean branch name. Same commits, same code. The detailed inline review discussion is on the superseded PR; here is the resolution state so you don't have to hop over:

@sandromello — all of your `CHANGES_REQUESTED` points are addressed, plus a merge of latest `main` (DEP-74 delegate model):

- **Agent capability advertisement** — removed entirely (constants, advertisement, and the gateway admission that consumed it). The agent picks up guardrails from the connection options; libhoop enforces or fails closed per protocol, so an older agent that can't evaluate them refuses the session rather than running unguarded.
- **Feature flag** — removed `beta.mssql_native_guardrails`. Native MSSQL guardrails are now enabled purely by configuring guardrail rules on the connection, same as Postgres/MySQL.
- **Postgres audit change** — reverted `PGConnectionWrite` to origin/main; kept metadata persistence only on `MSSQLConnectionWrite`, where the native block keeps the session open.

Net gateway effect: no MSSQL admission, no capability gating, no feature flag — it ships the connection's configured rules and libhoop decides. Verified against both the OSS `_libhoop` stub and the real libhoop, with unit + agent e2e (real SQL Server) + gateway integration all green.

**Merge order:** libhoop#96 must merge first — this PR's Integration Test checks out libhoop `main` and stays red until #96 lands.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1628.0.0-gef6af8d

PR #1629, opened 2026-07-24
feat(docker): AGPL/SSPL-free clean image line + provenance license gate

racerxdl:
## Summary

Removes AGPL/SSPL exposure from what customers deploy, **without breaking a customer that depends on the legacy MongoDB shell**, by splitting the agent images into two "trains" and shipping a bright-line **clean-only image line** that deployments point at.

**Closes DEP-66.**

## Background

- `hoophq/hoop` (gateway) is already AGPL/SSPL-free.
- The only real remaining violation is the **legacy MongoDB 5.0 shell (SSPL)** bundled in `hoophq/agent-tools` → inherited by `hoophq/hoopdev`.
- The existing Trivy license gate never caught it: Trivy only reads dpkg metadata, and the shell is a `curl`+tarball binary (per Trivy's docs, curl/tarball binaries "may not be detected").

## What this does

**Two image trains** (one recipe, `ARG INCLUDE_LEGACY_MONGO`, default `true`):
- **legacy** — keeps the SSPL shell (for the customer that needs it). Existing tags/behavior unchanged.
- **clean** — drops it; ships `mongosh` (Apache-2.0) only.

**Bright-line clean line** (no dirty history, so you can't accidentally deploy dirty):
- `hoophq/hoop-ng` (gateway, byte-identical to `hoophq/hoop`, retagged) and `hoophq/hoopdev-ng` (clean agent).
- `hoop-ng` / `hoopagent-ng` wrapper Helm charts that depend on the base charts via `file://` and only override image defaults — no template duplication.

**A provenance gate that closes the blind spot** (`scripts/ci/check-manual-binaries.py` + `licenses/manual-binaries.toml`): every hand-installed (non-dpkg) binary must be declared with its license; an AGPL/SSPL binary may only appear on a train that allows it. Grounded against the real published image (legacy passes with 69 artifacts; clean correctly rejects the SSPL shell).

**Runtime fix:** the database-explorer MongoDB introspection bypasses the session wrapper and hardcoded `/usr/local/bin/mongo`; it now resolves the shell at runtime (`mongosh --file /dev/stdin` on the clean train). Verified byte-identical output to the legacy shell against a live MongoDB.

## Safety / rollout

- **Existing repos, charts, and `auto-deploy` are untouched** — the legacy-dependent customer is never auto-deployed to clean.
- All clean-line release jobs are `continue-on-error` and **not** in `publish-release`'s `needs`, and guard on the clean base / per-arch images existing — a bootstrap-window miss or hiccup can never fail the release or block `auto-deploy`.
- Clean customers opt in by installing the `hoop-ng` / `hoopagent-ng` charts (same values, nested under `hoop-chart:` / `hoopagent-chart:`).

## Validation

- Provenance gate: grounded + validated against the real `agent-tools` image (legacy pass / clean fail-on-SSPL), plus synthetic unit cases.
- `mongosh` introspection parity verified against a live MongoDB server.
- `actionlint` clean on both workflows; `helm lint` + `helm template` render the `-ng` images; `go build` + `gofmt` clean.
- New PR-time `helm-charts` workflow validates the wrapper charts so a base-chart change can't silently break them.

## Merge/runtime ordering (post-merge, automatic)

On merge, the `agent-tools` workflow publishes both trains (incl. the clean base) and opens the base-bump PR; the next release then builds `hoopdev-ng` from that clean base. The guards make this self-healing if a release lands mid-window.

## Release label

`minor` — new, backwards-compatible capability.

Automated by MisterMal

### DIFF
--- a/deploy/helm-chart/README.md
+++ b/deploy/helm-chart/README.md
@@ -44,6 +44,31 @@ helm upgrade --install hoopagent https://releases.hoop.dev/release/$VERSION/hoop
     --set 'config.HOOP_KEY='
 ```
 
+## Installing the clean (AGPL/SSPL-free) line
+
+`hoop-ng-chart` and `hoopagent-ng-chart` are drop-in equivalents of
+`hoop-chart` and `hoopagent-chart` that default their images to the clean-only
+`hoophq/hoop-ng` and `hoophq/hoopdev-ng` repositories. Those repositories have
+no dirty history, so a deployment can never accidentally pull an image that
+contains AGPL/SSPL packages. They accept the same values as the base charts
+(nested under the `hoop-chart:` / `hoopagent-chart:` key).
+
+> `hoop-ng-chart` / `hoopagent-ng-chart` are the packaged artifact names only;
+> the installed Helm release name (`hoop` / `hoopagent` below) is unchanged, so
+> switching an existing install to the clean line is an in-place upgrade.
+
+```sh
+VERSION=$(curl -s https://releases.hoop.dev/release/latest.txt)
+# gateway
+helm upgrade --install hoop \
+  https://releases.hoop.dev/release/$VERSION/hoop-ng-chart-$VERSION.tgz \
+  -f values.yaml
+# agent
+helm upgrade --install hoopagent \
+  https://releases.hoop.dev/release/$VERSION/hoopagent-ng-chart-$VERSION.tgz \
+  --set 'hoopagent-chart.config.HOOP_KEY='
+```
+
 ## Development
 
 To add new configuration(s)
--- a/deploy/helm-chart/chart/hoop-ng/.gitignore
+++ b/deploy/helm-chart/chart/hoop-ng/.gitignore
@@ -1,0 +1,4 @@
+# Built at package time by `helm package -u` (the file:// dependency is
+# bundled from ../gateway); never commit the resolved artifacts.
+charts/
+Chart.lock
--- a/deploy/helm-chart/chart/hoop-ng/Chart.yaml
+++ b/deploy/helm-chart/chart/hoop-ng/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v2
+# Clean image line (DEP-66). Thin wrapper over hoop-chart that only changes the
+# default image repositories to the clean-only, no-dirty-history hoophq/*-ng
+# repos, so a deployment can never accidentally pull an AGPL/SSPL image. All
+# templates and other values are inherited from hoop-chart (no duplication).
+name: hoop-ng-chart
+description: Clean-only (AGPL/SSPL-free) Hoop gateway chart
+type: application
+version: 0.0.1
+dependencies:
+  - name: hoop-chart
+    version: "0.0.2"
+    repository: "file://../gateway"
--- a/deploy/helm-chart/chart/hoop-ng/values.yaml
+++ b/deploy/helm-chart/chart/hoop-ng/values.yaml
@@ -1,0 +1,12 @@
+# hoop-ng: bright-line clean image line (DEP-66).
+#
+# This chart only overrides the default image repositories of hoop-chart to the
+# clean-only hoophq/*-ng repos. Every other value is inherited from hoop-chart
+# and can be set here under the "hoop-chart:" key exactly as documented for the
+# gateway chart (e.g. hoop-chart.postgres.enabled, hoop-chart.config.*, ...).
+hoop-chart:
+  image:
+    gw:
+      repository: hoophq/hoop-ng
+  defaultAgent:
+    imageRepository: hoophq/hoopdev-ng
--- a/deploy/helm-chart/chart/hoopagent-ng/.gitignore
+++ b/deploy/helm-chart/chart/hoopagent-ng/.gitignore
@@ -1,0 +1,4 @@
+# Built at package time by `helm package -u` (the file:// dependency is
+# bundled from ../agent); never commit the resolved artifacts.
+charts/
+Chart.lock
--- a/deploy/helm-chart/chart/hoopagent-ng/Chart.yaml
+++ b/deploy/helm-chart/chart/hoopagent-ng/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+# Clean image line (DEP-66). Thin wrapper over hoopagent-chart that defaults the
+# agent image to the clean-only, no-dirty-history hoophq/hoopdev-ng repo. All
+# templates and other values are inherited from hoopagent-chart (no duplication).
+name: hoopagent-ng-chart
+description: Clean-only (AGPL/SSPL-free) Hoop agent chart
+type: application
+version: 0.0.1
+dependencies:
+  - name: hoopagent-chart
+    version: "0.0.1"
+    repository: "file://../agent"
--- a/deploy/helm-chart/chart/hoopagent-ng/values.yaml
+++ b/deploy/helm-chart/chart/hoopagent-ng/values.yaml
@@ -1,0 +1,8 @@
+# hoopagent-ng: bright-line clean image line (DEP-66).
+#
+# This chart only overrides the default agent image repository of hoopagent-chart
+# to the clean-only hoophq/hoopdev-ng repo. Every other value is inherited from
+# hoopagent-chart and can be set here under the "hoopagent-chart:" key.
+hoopagent-chart:
+  image:
+    repository: hoophq/hoopdev-ng
--- a/gateway/api/connections/helpers.go
+++ b/gateway/api/connections/helpers.go
@@ -687,10 +687,20 @@ func getConnectionCommandOverride(currentConnectionType pb.ConnectionType, conne
 	case pb.ConnectionTypeCloudWatch, pb.ConnectionTypeDynamoDB:
 		return []string{"bash"}
 	case pb.ConnectionTypeMongoDB:
-		// Force the execution using the legacy mongo cli
-		// It avoids using any wrapper scripts (.e.g: /opt/hoop/bin/mongo) to perform system queries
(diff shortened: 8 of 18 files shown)


sandromello:
✅ Build Completed with Success, Version=1629.0.0-g1b2853c

racerxdl:
Thanks — good catch on the provenance-gate blind spot. Confirmed against the published image: the legacy MongoDB **server** tarball was left extracted at `/mongodb-linux-*/bin/` with `mongod`/`mongos` (SSPL) lingering undeclared, plus the AWS CLI zip + installer dir. Fixed in `2749d837`:

**Dockerfile.tools — remove the leftover trees (both cleanup paths qodo suggested):**
- legacy Mongo: `rm -rf` the extracted `mongodb-linux-*` dir (not just the `.tgz`), so only `/usr/local/bin/mongo` remains — the SSPL `mongod`/`mongos` no longer ship.
- AWS CLI: `rm -rf awscli-exe-linux.zip aws` after `aws/install`.
- session-manager-plugin: `rm -f` the `.deb` after `dpkg -i` (same hygiene).

**Gate — close the remaining coverage + robustness gaps:**
- Added `/usr/local/aws-cli` (the AWS CLI install tree) to `scan_dirs` with a covering prefix entry.
- Reworked ownership resolution: a fast file-list pre-filter now produces *candidates*, and every would-be violation is re-confirmed with authoritative `dpkg -S` before failing — so a pre-filter miss (e.g. a dpkg diversion) can’t cause a false CI failure, and it stays fast on large trees (legacy scan of 7.5k artifacts runs in ~0.5s).
- Hardened the embedded shell (paths via argv / `set --`) and made `scan_dirs` reject whitespace/glob.

Verified against the real image: legacy passes; the clean train fails on `/usr/local/bin/mongo` after `dpkg -S` confirms it’s unmanaged.

Automated by MisterMal

racerxdl:
### Reviewer note: legacy Mongo cleanup keeps the client, drops only server-side leftovers

The `Dockerfile.tools` cleanup in `2749d837` (addressing qodo's leftover-SSPL finding) is **intentionally scoped so the legacy-dependent customer is unaffected**. The legacy Mongo tarball is the full MongoDB **server** distribution, downloaded only to extract the shell; the shell is `mv`'d to `/usr/local/bin` *before* the extraction dir is removed.

What the **legacy** train keeps vs. removes:

| Item | Legacy train | Why |
|------|--------------|-----|
| `mongo` shell (`/usr/local/bin/mongo`, v5.0.31, SSPL) | ✅ **kept** | The client customers connect with — this is the whole reason the legacy train exists. |
| `mongod`, `mongos`, `install_compass` (`/mongodb-linux-*/bin/`) | ❌ removed | MongoDB **server** daemon, sharding **router**, and Compass **GUI installer** — a hoop agent is a connection client/proxy and never runs a Mongo server. These were accidental leftovers from extracting the server tarball. |
| aws CLI (`/usr/local/bin/aws` → `/usr/local/aws-cli`) | ✅ kept | Only the throwaway `/aws` *installer* dir + zip were removed. |
| `session-manager-plugin` | ✅ kept (`dpkg -l` = `ii`) | Only the post-install `.deb` file was removed. |

So the cleanup removes ~unused SSPL/bloat only; the SSPL **shell** that legacy customers require still ships on the legacy train (`hoophq/agent-tools` / `hoophq/hoopdev`). The **clean** train never installs any of it (the whole block is gated off).

Confirmed with the repo owner that no customer invokes `mongod`/`mongos`/`install_compass` from the agent image.

Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1629.0.0-g2749d83

gateway/api/connections/helpers.go
```diff
@@ -687,10 +687,20 @@ func getConnectionCommandOverride(currentConnectionType pb.ConnectionType, conne
...
+		// MongoDB's supported shell going forward.
+		// mongosh reads a piped script as a REPL and echoes "test>" prompts to
+		// stdout, which would corrupt the JSON the caller parses; --file
+		// /dev/stdin runs the same piped script in file mode, producing output
+		// byte-identical to the legacy shell (verified against a live server).
 		if len(connectionCmd) > 1 {
-			cmd = append(cmd, "/usr/local/bin/mongo")
+			cmd = append(cmd, "/bin/sh", "-c",
+				`if [ -x /usr/local/bin/mongo ]; then exec /usr/local/bin/mongo "$@"; else exec mongosh "$@" --file /dev/stdin; fi`,
+				"mongo")
```
sandromello: Remove this. It will break the schema listing.
We shouldn't deal with this for now, the user could create a standard connection passing the `mongosh` shell

rootfs/opt/hoop/bin/mongo
```diff
@@ -15,7 +22,14 @@ fi
...
 if [ "$HOOP_EXEC_MONGOSH_RUNTIME" == "1" ]; then
-    MONGO_BIN=/usr/bin/mongosh
+    MONGO_BIN="$(command -v mongosh || true)"
+fi
+
+# Fail fast with a clear message rather than a confusing "command not found"
+# if no usable client is present in this image.
+if [ -z "$MONGO_BIN" ] || [ ! -x "$MONGO_BIN" ]; then
+    echo "mongo: no mongo/mongosh client found in this image" >&2
+    exit 1
```
sandromello: Remove any changes on this file. It will break for other customers. We can't enforce mongosh for now.

deploy/helm-chart/README.md
```diff
@@ -44,6 +44,31 @@ helm upgrade --install hoopagent https://releases.hoop.dev/release/$VERSION/hoop
...
+# gateway
+helm upgrade --install hoop \
+  https://releases.hoop.dev/release/$VERSION/hoop-ng-chart-$VERSION.tgz \
+  -f values.yaml
+# agent
+helm upgrade --install hoopagent \
+  https://releases.hoop.dev/release/$VERSION/hoopagent-ng-chart-$VERSION.tgz \
+  --set 'hoopagent-chart.config.HOOP_KEY='
+```
+
```
sandromello, line 71: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

deploy/helm-chart/chart/hoopagent-ng/values.yaml
```diff
@@ -0,0 +1,8 @@
+# hoopagent-ng: bright-line clean image line (DEP-66).
```
sandromello, line 1: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

deploy/helm-chart/chart/hoopagent-ng/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
```
sandromello, line 1: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

deploy/helm-chart/chart/hoopagent-ng/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Built at package time by `helm package -u` (the file:// dependency is
```
sandromello, line 1: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

deploy/helm-chart/chart/hoop-ng/values.yaml
```diff
@@ -0,0 +1,12 @@
+# hoop-ng: bright-line clean image line (DEP-66).
```
sandromello, line 1: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

deploy/helm-chart/chart/hoop-ng/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
```
sandromello, line 1: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

deploy/helm-chart/chart/hoop-ng/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Built at package time by `helm package -u` (the file:// dependency is
```
sandromello, line 1: Why are you changing this. It only contains yaml files.
The image could be referenced inside the values.yaml

sandromello (changes requested):

gateway/api/connections/helpers.go
racerxdl: Reverted in `8c4dbb5f` — `getConnectionCommandOverride` is back to main (hardcoded `/usr/local/bin/mongo`). Clean-train Mongo schema listing is deferred as you suggested; customers use the existing mongosh runtime (`HOOP_EXEC_MONGOSH_RUNTIME`) on a standard connection. Automated by MisterMal

rootfs/opt/hoop/bin/mongo
racerxdl: Reverted in `8c4dbb5f` — the wrapper is restored to main, no mongosh enforcement. Automated by MisterMal

deploy/helm-chart/chart/hoop-ng/Chart.yaml
racerxdl: Per our follow-up chat, keeping the wrapper charts — they are the clean deploy artifact we point customers at ("use these charts instead"). They only default the images to the clean-only `hoophq/hoop-ng` / `hoophq/hoopdev-ng` repos (no dirty history) and inherit everything else from the base chart via the file:// dependency. Automated by MisterMal

sandromello:
✅ Build Completed with Success, Version=1629.0.0-g8c4dbb5

sandromello (approved):

PR #1651, opened 2026-07-29
[Feature] hoopinspect: protocol-aware inspection sidecar (v0.1.0)

matheusfrancisco:
## What this is

Good READ first
https://github.com/hoophq/hoop/blob/4870feeb17e875b098a3fcaf1fe49a9bdcbe1c21/hoopinspect/README.md

https://github.com/hoophq/hoop/blob/4870feeb17e875b098a3fcaf1fe49a9bdcbe1c21/docs/adr/hoopinspect-flow.md

`hoopinspect` turns raw wire-protocol bytes into structured statements, and
structured statements into allow/deny verdicts. It sits behind something that
already owns TLS and identity — typically Envoy, forwarding plaintext over
loopback or a unix socket.

The library opens no socket, terminates no TLS, and routes nothing. You hand it
bytes you already have, and whatever holds the connection keeps holding it. The
nested `cmd` module wraps it in a relay — a TCP port or a unix socket, per lane
— for deployments that want a process instead of a library.

**The root module has zero dependencies.** Standard library only, tests
included, no `go.sum`. Everything heavy (YAML parsing, PII detection, SQLite)
lives in a nested module you link only if you use it.

## Why not just Envoy

Envoy already does some of this, and pretending otherwise is a bad way to lose
a technical review. `envoy.filters.network.postgres_proxy` parses SQL from
`Query` and `Parse` messages and emits `table.db → [operations]` metadata the
RBAC filter can act on. Within Postgres, at table-and-verb granularity, that is
real.

Here is the boundary:

| | Envoy | hoopinspect |
|---|---|---|
| Postgres SQL parse | `postgres_proxy`, best effort | full statement text |
| Postgres granularity | `table.db` + operation verb | statement, operation, tables |
| Postgres **response** | ✗ | result columns, row count, masking by re-framing |
| HTTP request | `ext_authz`: method, path, headers, bounded body | same, plus normalized resource |
| HTTP **response** | ✗, `ext_authz` decides before the upstream is called | status, headers, body |
| Deny UX | RBAC/`ext_authz` drops or returns a bare 403 | operator-authored message |

The response side is the load-bearing gap. Envoy consults `ext_authz` *before*
calling the upstream, so it never sees the row that comes back. On HTTP request
authorization, Envoy is not blind and this adds little.

## Features

**Protocol decoding**
- Postgres wire protocol: `Query`/`Parse` on the request side; `RowDescription`,
  `DataRow` and the three result-set terminators on the response side. Stateful,
  because one `RowDescription` describes every `DataRow` after it and those land
  in different TCP reads.
- HTTP/1.x request and response decoding, plus two direct entry points
  (`InspectRequest`/`InspectResponse`) for use inside an existing HTTP pipeline
  with no re-parse and no second copy of the body.
- Normalized HTTP resources: `/users/12345/orders/98765` → `/users/*/orders/*`,
  so one rule replaces a regex per endpoint. Conservative on short slugs, so a
  policy comes out too narrow rather than too broad.
- Split-read safe. Each codec is tested by feeding the same message in two
  chunks at every possible boundary.
- Codecs are registered by factory, not instance — two connections sharing one
  codec would corrupt each other's reassembly and leak one tenant's SQL into
  another's audit trail.

**Policy**
- Local rules, evaluated in order, first match wins. SQL: `deny_words_list`,
  `pattern_match` (RE2), `operation`, `table`. HTTP: `http_resource`,
  `http_status`. Cross-protocol: `pii`.
- Operation classification runs a lexer that strips comments and string
  literals first, so `SELECT 'DROP TABLE customers'` classifies as `select`.
- One ordered rule set can mix protocols. An HTTP rule never matches a SQL
  statement and vice versa.
- OPA integration posting to a Data API endpoint. hoopinspect does not own
  policy; it owns the input document, a strict superset of what `ext_authz` and
  `postgres_proxy` metadata provide, in one shape for both protocols.
- Chainable (`policy.Chain{local, opa}`) so a statement the local rules already
  forbid costs no network round-trip. The chain short-circuits on a denial only;
  a fail-open error is accumulated and later evaluators still run, so an
  unreachable OPA cannot skip a rule that would have denied.
- Both evaluators fail closed. Unreachable OPA, a 500, or an undefined decision
  denies. `FailOpen` inverts it where availability outranks enforcement.
- Observe-only mode per lane (`enforce: false`, the default) so a misconfigured
  rule cannot take production down on first deploy.

**Response masking**
- Two mechanisms, picked by asking the codec rather than consulting a list of
  protocol names: **substitution** for HTTP (with `Content-Length` retagged
  after the rewrite) and **re-framing** for Postgres (each changed `DataRow`
  rebuilt around the new values).
- Four strategies: `redact`, `mask`, `partial` (keep last N), `hash`.
- Mask by detected entity, or by column name where the protocol names its
  columns — position beats detection when you have it.
- Requests are never rewritten. Changing the statement the upstream executes is
  a correctness change wearing a privacy label.
- A codec offering neither mechanism gets its `mask` section refused at startup,
  rather than accepting a masking config that can never fire.
- A body whose `Content-Length` cannot be corrected — the header block already
  went out in an earlier read — is forwarded **unmasked** and the skip is
  audited. Masking it anyway would grow the body past the length the client was
  told to read, and the client would stop mid-token and report a corrupt
  upstream.

**PII and secret detection**
- Backed by [alcatraz](https://github.com/hoophq/alcatraz): 45 entity types
  across 12 countries, 25 checksum-verified (Luhn on cards, ISO 7064 mod-97 on
  IBAN, Verhoeff on Aadhaar, mod-11 on the Brazilian schemes).
- Credential recognizers registered into the same engine: `AWS_ACCESS_KEY`,
  `JWT` (decodes the header rather than matching its shape), and `PRIVATE_KEY`
  (including a PEM block a size limit cut short).
- One detector drives both masking and the `pii` policy rule. The guardrail half
  matters independently: a national ID in a `WHERE` clause lands in the
  database's own query log, slow-query log and `EXPLAIN` output, and response
  masking never undoes that.
- No all-entities default, on purpose. Enabling all 45 recognizers masks
  ordinary numeric columns — `US_SSN` fires on about a third of random
  nine-digit business ids, because SSNs carry no checksum.

**Audit**
- JSONL sink to stdout or a file, async by default so a slow sink never blocks a
  user's query.
- Bounded queue with an explicit `ErrQueueFull`. Blocking would make a slow
  audit store stall the query this exists to avoid; dropping silently would take
  the decision away from the gate.
- `fail_closed: true` refuses a statement whose audit write failed, because an
  unrecorded statement is the one an attacker wants.
- `redact_statements` keeps a stable fingerprint instead of the text where
  recording the raw literal is the wrong trade.

**Storage and query API**
- `store.Store` is the read side: in-memory and SQLite backends, each
  implementing both `Sink` and the query interface so the two paths cannot drift.
- `GET /sessions`, `/sessions/{id}`, `/sessions/{id}/events`, `/events`,
  `/stats`.
- A deployment that only wants JSONL never links a database at all.

**Operations**
- One config file, YAML or JSON, extension picks the parser. Strict decoding, so
  a mistyped key fails startup instead of silently disabling a control.
- Per-listener config inheriting documented top-level defaults. `policy.rules`
  concatenate; `policy.opa`, `policy.enforce` and `mask` replace.
- YAML anchors for sharing a rule block across lanes; `x-`-prefixed keys are
  dropped before validation.
- `-validate` builds every lane offline and reports every problem in one run.
  It refuses: masking on a codec that cannot carry it, a `pii` rule naming an
  entity absent from `pii.entities` (otherwise the guardrail looks live while
  allowing everything), key typos, and bad regexes — each naming the lane.
- `GET /config` reports what each lane *resolved*, since inheritance happens at
  startup and reading the file cannot tell you. Rule names only: a
  `pattern_regex` can encode business logic.
- No build tags. Every capability is decided by config, so an operator turning
  on PII detection does not also have to swap the binary.
- Admin surface on its own listener: `/healthz`, `/stats`, `/config`, `/events`,
  `/api/*`.
- Idle timeouts and connection caps per lane.

## Running it

Two entry points, one relay.

**Through the hoop CLI**, linked into the binary the release pipeline already
builds — no CI change and no second artifact to ship:

```bash
hoop start inspect --config config.yaml --validate   # check the config and exit
hoop start inspect --config config.yaml              # run
```

`--config` also reads `HOOP_INSPECT_CONFIG`, which is the shape a Kubernetes
deployment wants: mount the ConfigMap, set the variable, pass no arguments.
The version reported at `/stats` comes from the Makefile's existing `-ldflags`.

**As a standalone binary**, when a sidecar container should carry the relay and
nothing else. `hoopinspect/cmd` is the nested module where the optional plugins
(YAML front end, alcatraz PII detection) get linked, so the root stays
dependency-free.

Both call the same `sidecar.Setup` and `sidecar.Run`.

## Transport: TCP or unix socket

Each lane binds a TCP port or a unix socket, chosen per listener. **TCP is the
default** — omit `network` and you get a port.

```yaml
listeners:
  - name: appdb-uds
    network: unix                      # the only line that changes it
    listen: /run/hoop-inspect/pg.sock
```

Nothing above the transport changes: policy, masking, audit and upstream TLS
read a `net.Conn` and never ask what kind it is. Lanes in one process can
differ.

A TCP listener is reachable by anything that can route to the host; a socket
opens no port at all, so reachability becomes a filesystem question.
`deploy/docker-compose/envoy-stack/uds/` is a runnable overlay — `netstat`
inside the sidecar then shows only the admin port, and the data lanes have no
TCP presence.

Which one is live is observable three ways: the `network` field in the startup
log, the `addr` in `GET /stats` (a path means unix, a `host:port` means TCP),
and the socket inode on disk.

A socket left behind by a SIGKILL or an OOM kill is reclaimed at startup —
dialed first, unlinked only if nothing answers, so a live socket is never
stolen from another process.

## Upstream TLS

The hop from the relay to the backend can be encrypted, and it costs no
inspection: the relay is the TLS *client* there, so it decrypts on read and the
gate still sees plaintext. Encryption protects the bytes crossing the network,
not the bytes the relay was built to read.

Three details that are not obvious:

- **Postgres negotiates in-band.** A TLS-on-connect dial fails against it — the
  server expects an 8-byte `SSLRequest` and a one-byte `S`/`N` reply before any
  handshake. The relay speaks that exchange.
- **A refusal is an error, never a downgrade.** If the server answers `N` the
  connection fails with a message naming the likely cause, rather than sending
  credentials in the clear.
- **Channel binding is stripped from the server's SASL offer.**
  `SCRAM-SHA-256-PLUS` cannot work across a proxy that terminates TLS, and libpq
  refuses a `-PLUS` mechanism over a link it knows is unencrypted, so relaying
  the offer would fail the connection outright. Plain `SCRAM-SHA-256` remains,
  authenticating the same password against the same verifier.

## The Envoy stack

`deploy/docker-compose/envoy-stack` runs the whole path in six containers: Envoy
terminating TLS and calling OPA for reachability, `hoop-inspect` behind it as an
ordinary upstream, Postgres (with `ssl=on`) and httpbin behind that. No hoop
gateway, no agent, no control-plane database — the sidecar reads one YAML file.

```bash
cd deploy/docker-compose/envoy-stack
./run.sh
./demo.sh
```

`./demo.sh` walks three tiers and prints the audit trail:

- **Tier 1** — OPA answers reachability and nothing else. `alice` gets 200,
  `bob` gets 403.
- **Tier 2, postgres** — Envoy forwarded these bytes as opaque TCP and consulted
  nobody, so the sidecar is the only thing that can judge them. A `DELETE` comes
  back refused in a real pgwire `ErrorResponse`; a `SELECT` comes back with the
  result set masked.
- **Tier 2, http** — a denial on the **response**, which no `ext_authz` config
  can express.

## Known limits

Read these before writing a policy against it.

- **`Tables` is best effort.** A lexer produces it, hitting the same ceiling
  Envoy's own docs acknowledge for `postgres_proxy`. Empty means "could not
  determine", never "touches nothing". Use `RequireTableMatch: true` on rules
  protecting something critical and accept the false positives.
- **A response batch can be truncated.** The Postgres codec stops decoding
  columns past 1000 rows in one result set, to keep the relay's memory out of
  the query's hands. It keeps counting and marks the batch `Truncated`. A policy
  MUST read that as inconclusive, never as proof a value is absent.
- **A response statement carries no verb.** A `FromServer` statement reports
  `OpUnknown`. Key response-side SQL rules on `Result`, not `Operation`.
- **PII detection is neither sound nor complete.** A checksum-verified
  identifier is solid; everything else is a pattern. Name detection needs NER,
  which is not wired. A caller can split a value across two responses. Masking
  raises the cost of accidental exposure — it does not replace withholding
  access to the table.
- **HTTP/1.x only for stream decoding.** HTTP/2 and HTTP/3 framing belongs to
  whatever terminated the connection.
- **Plaintext downstream only.** A client negotiating TLS end-to-end past the
  relay leaves nothing to parse; terminating that leg is Envoy's job. The
  UPSTREAM leg may be TLS — the relay originates it and still inspects.
- **HTTP masking needs the header and body in one read.** When the upstream
  sends them separately, `Content-Length` can no longer be corrected and the
  body is forwarded unmasked with an audit record saying so. Closing that gap
  needs response buffering across reads in the HTTP codec.
- **Channel binding is unavailable through the relay.** By construction, see
  Upstream TLS above.
- **Statements are not transactions.** Each is evaluated independently, with no
  cross-statement session state.
- MySQL, MSSQL and MongoDB codecs are **not shipped**. The `Codec` interface and
  the SQL classifier are protocol-agnostic, so adding one means a new
  `codec/<name>` package and nothing else. Earlier implementations were removed
  to keep the surface to what is exercised end to end.

## Scope

87 files, ~23.2k lines: ~9.6k of implementation, ~8.8k of tests across 314 test
functions. Additive under `hoopinspect/`,
`deploy/docker-compose/envoy-stack/` and `docs/adr/`.

The only existing code touched is `client/` (3 files, +133: the
`hoop start inspect` command and its module wiring) and `go.work` (+5). That
adds one real dependency to the CLI, `github.com/hoophq/alcatraz` — 316 KB with
no dependencies of its own.

## How to test

```bash
go build ./hoopinspect/...
go test ./hoopinspect/...
(cd hoopinspect/pii/alcatraz && go test ./...)
(cd hoopinspect/store/sqlite && go test ./...)
```

End to end:

```bash
cd deploy/docker-compose/envoy-stack && ./run.sh && ./demo.sh
```

Through the CLI rather than the standalone binary:

```bash
go build -o /tmp/hoop client/hoop.go
/tmp/hoop start inspect --config deploy/docker-compose/envoy-stack/hoopinspect/config.yaml --validate
```

```
config OK: 2 listener(s)
  appdb            postgres  enforcing 2 rule(s) + masking
  httpbin          http      enforcing 4 rule(s) + masking
```

Over unix sockets instead of TCP:

```bash
cd deploy/docker-compose/envoy-stack
./run.sh && docker compose down
docker compose -f docker-compose.yml -f uds/docker-compose.uds.yml up -d

# addr reads as a socket path, not host:port
curl -s localhost:19000/stats | python3 -m json.tool

# and the data ports are gone: only :::19000 (admin) is left
docker compose -f docker-compose.yml -f uds/docker-compose.uds.yml \
  exec -T hoop-inspect netstat -ltn
```

Verifying the upstream TLS hop, from a client session through the whole path:

```bash
docker compose exec -T client env PGPASSWORD=apppass PGSSLMODE=disable \
  psql -h envoy -p 5432 -U appuser -d appdb \
  -c "SELECT ssl, version FROM pg_stat_ssl WHERE pid = pg_backend_pid();"
#  t | TLSv1.3
```


sandromello:
✅ Build Completed with Success, Version=1651.0.0-g4264b95

sandromello:
✅ Build Completed with Success, Version=1651.0.0-g5130d36

sandromello:
✅ Build Completed with Success, Version=1651.0.0-g1778932

sandromello:
✅ Build Completed with Success, Version=1651.0.0-g07a9870

sandromello:
✅ Build Completed with Success, Version=1651.0.0-g67f8361

sandromello:
✅ Build Completed with Success, Version=1651.0.0-g4870fee

sandromello (approved):

p3rotto (approved):

PR #1653, opened 2026-07-30
EVL-115: PR 0.2 — scaffolding deletion + webapp_v2 doc reorg (single owner per topic)

rogefm:
> [!IMPORTANT]
> Every PR MUST have **exactly one** of these labels, the merge is blocked otherwise:
> - `major` / `minor` / `patch` publishes a new release on merge with the corresponding semver bump.
> - `skip-release` merges without publishing a release (use for docs, CI changes, refactors, etc.).

## 📝 Description

Executes roadmap **PR 0.2 — Scaffolding deletion + doc refresh** (`webapp_v2/MIGRATION_ROADMAP.md`, Phase 0), plus the migration-doc debt that accumulated while EVL-101/EVL-104/EVL-105 merged.

**Commit 1 — code (pure deletion, no behavior change):** removes the 11 unrouted, unreferenced 7-line page stubs from the initial shell commit — `pages/{Connections (+Setup), Dashboard, Sessions, Reviews, Guardrails (+Create), Resources (+Configure, +Create)}` and the orphaned `pages/Integrations/Authentication` (EVL-101 merged without touching it; roadmap B3.7 rebuilds that page fresh). Zero imports anywhere, no routes in `Router.jsx` (untouched — empty diff). Sidebar/CommandPalette reference these features by URL path only, so those routes keep landing on the CLJS catch-all exactly as before.

**Commit 2 — docs ("single owner per topic"):** each topic now lives in exactly one of the 7 `webapp_v2/*.md` files; other files carry one-line pointers (see the new **Documentation map** in `README.md`). The rot-prone store/service inventories (which listed 6 of 27 services and disagreed between files) are replaced by pointers to `src/stores/`/`src/services/` — the filesystem is the source of truth — plus non-obvious notes in `COMPONENTS.md`.

Staleness fixed along the way:
- EVL-101 (#1633), EVL-104 (#1638), EVL-105 (#1641) marked merged; B3.7/B4.4 unblocked; roadmap Top Risk 1 retired
- The deleted `useBridgeStore.showSnackbar` is no longer documented; the current bridge methods are (incl. the previously undocumented `syncPrimaryConnectionFromUrl`)
- Duplicate `react-shell` contract row removed; `/plugins/manage/*` redirect routes added to the routing table
- Phantom `VITE_GATEWAY_URL` env var replaced by the real `API_URL` proxy var (verified against `vite.config.js`); `SEGMENT_WRITE_KEY` documented in the README env table
[...]

PR #1656, opened 2026-07-30
EVL-119: Migrate /dashboard to webapp_v2

rogefm:
## 📝 Description

Migrates the admin-only `/dashboard` from ClojureScript to React and deletes the CLJS original. This is **B1.1** of `webapp_v2/MIGRATION_ROADMAP.md` — Track B, Wave 1, the roadmap's priority #1 page — and it closes [EVL-119](https://linear.app/hoophq/issue/EVL-119).

It is also the app's **first chart consumer**, so it establishes the pattern the rest of the migration will inherit:

- `@mantine/charts@8.3.18` + `recharts@2.15.4`, with the stylesheet imported as `styles.layer.css` so it joins the `mantine` cascade layer and CSS Modules keep winning.
- `components/BarChart` and `components/DonutChart` — thin wrappers carrying app-wide defaults only. The two bar charts on this page already need opposite configurations, so anything chart-specific stays at the call site.
- `components/SegmentedControl` — gains a **locked item** concept for Enterprise-gated options.
- `CHART_SERIES_COLORS` in `theme.js` — the categorical palette, consumed with `i % length`.
- The route is the only lazily-loaded one: recharts + d3 is ~150 kB gzipped for a page only admins can reach. It now builds into its own 436 kB / 120 kB-gzipped chunk instead of the main bundle.

### Deliberate deviations from the ClojureScript original

| Area | What changed and why |
|---|---|
| **Reviews bar colours** | Now green/red. The legacy chart config *declared* green and red, but a hard-coded `fill="var(--accent-11)"` / `"var(--accent-8)"` overrode it, so both bars rendered indigo and approved/rejected were indistinguishable. This restores the original intent. |
| **Redacted Data bar colours** | One indigo series with a visible x-axis, instead of a per-`info_type` indigo ramp with no axis. Mantine colours per *series*, not per data point (`getBarColor` only receives the numeric value). Showing the category on the axis makes bars identifiable without hovering — the alternative, `barProps.shape`, would desync the tooltip swatch from the bar. |
| **Card borders** | Cards use `withBorder`. The v2 page background is near-white (`#fcfcfd`), where the legacy borderless white card would be invisible; the CLJS page sat on a grey background. |
| **Loading state** | Adds `PageLoader` + `useMinDelay` on first load, matching every other React page. The legacy page rendered blanks while fetching. |

The two "Coming soon" placeholder cards are **ported as-is** — the 2×2 grid is unchanged.

### Data quirks fixed (one commit each, so every behaviour change is reviewable on its own)

1. **Mixed timezone bases** (`464e0ff`). cljs-time has two clocks: `today-at-midnight` is UTC midnight, `today` is the local calendar date. The Sessions card and Reviews count used the first; Redacted Data, every range and every label used the second. West of UTC the three headline figures on one card described different days. Everything now derives from the user's local day. Also fixes the review chart's bucket key, which was the same bug one level down — buckets keyed on the UTC date but labelled in local time.
2. **Ranges longer than 24h excluded today** (`b1d5cc`). The gateway compares `ended_at` against `TO_TIMESTAMP(end_date, 'YYYY-MM-DD')`, i.e. midnight *starting* that day, so `end_date=<today>` silently ended the window yesterday. `7d`/`14d`/`30d`/`3m` never showed anything redacted today, quietly contradicting the "Redacted Data" figure on the same page. All ranges now send `end_date=<tomorrow>`. Corrected in `08b1670a` after review: moving `end_date` without moving `start_date` stretched every range by a day, so `7d` covered eight dates. `rangeStart` now returns `today - (days - 1)` and an N-day range covers exactly N dates. (The earlier claim that the extra day was pre-existing was wrong — the legacy window was N days, just shifted one day into the past.)
3. **Empty review buckets** (`4295386`). The legacy `cond` had no `:else`, so a PENDING/REVOKED review wrote a literal `nil` key and left a `{approved: 0, rejected: 0}` bucket rendering as an invisible zero-height bar. Those buckets are no longer created.
4. **The Reviews chart only ever drew rejections** (`0e713a88`). A review carries one decision but its status keeps moving afterwards: `PENDING → APPROVED → PROCESSING → EXECUTED`, or `APPROVED → REVOKED` for JIT. `APPROVED` is transient and `REJECTED` is terminal, so matching the two literals — which the legacy `cond` did and the port reproduced faithfully — counts only approvals nobody has used yet while every rejection persists. The dev database shows it exactly: one `EXECUTED`, one `REJECTED`, one `PENDING`, zero `APPROVED`. The bucket rule is now derived (PENDING undecided, REJECTED the only rejection, everything else downstream of an approval) rather than enumerated, so a status added later surfaces instead of vanishing.

### Other correctness details in the port

- **`replaceAll`, not `replace`** — `clojure.string/replace` with a string pattern replaces *every* occurrence, so a single JS `.replace('_','-')` would have mislabelled and misbucketed every multi-underscore info type (`us-social_security_number`).
- **`/reviews` is fetched once.** It accepts no query parameters, so the legacy app's refetch on every range change re-issued an identical unbounded request. Both the today count and every chart range are now derived from one fetch.
- **Session count asks for `limit=1`.** `total` comes from a separate un-limited COUNT query, so this returns the same number for one row instead of twenty.
- **Bars and slices are sorted explicitly.** The report query has no `ORDER BY`, so server row order — and with it the positional colour assignment — varied between identical requests.
- **Per-slice error handling + a request token.** One failing endpoint degrades one card rather than the page (`/reviews` is unbounded and the most likely to time out), and fast range clicking cannot leave a stale chart.

### ClojureScript removed

`webapp/src/webapp/dashboard/` (5 files), `components/charts.cljs` (a Reagent port of shadcn's `chart.tsx`, dashboard-only), the `:dashboard` bidi entry, the `:dashboard-panel` defmethod, the sidebar entry, five `reports->` subscriptions, and the dashboard half of `events/reports.cljs`.

**Kept deliberately**: `:reports->get-report-by-session-id` and friends, the `:reports->session` sub and its db key — session details, the RDP data-masking analytics view and `events/audit.cljs` all still read the per-session report through them. This is the main regression surface; see "How to test".

Three dependencies fell out as dead weight, each verified with a repo-wide grep: `recharts`, `cljs-time`, and the `--chart-1..31` custom properties in `tailwind.css`. `package-lock.json` is regenerated in the same commit so `npm ci` doesn't fail on the drift.

### Doc correction worth a look

`MIGRATION_ROADMAP.md` asserted that `bidi/path-for` throws on an unknown handler, which is the stated reason Track A must prune sidebar constants in a *strictly earlier* PR. Reading `bidi/bidi.cljc` out of `bidi-2.1.6.jar` shows `path-for*` throws **only** when the handler is `nil`; an unknown keyword falls through `unmatch-pair` and returns `nil`. The ordering is still worth following — otherwise the nav item ships with `:uri nil` — but it's advisory, not a build-breaker, and a route plus its sidebar entry can land in one PR. This PR did exactly that.

### Known limitation — the Redacted Data card cannot honour a local day

"Today" is computed in the viewer's local calendar day, so São Paulo and Ottawa each read the dashboard in their own time. Two of the three figures honour that exactly: `/sessions` takes RFC3339 instants, and `/reviews` is filtered client-side on real timestamps.

`/reports/sessions` cannot. It accepts only `YYYY-MM-DD` and resolves it with `TO_TIMESTAMP(@start_date, 'YYYY-MM-DD')`, which lands in the database session timezone — `Etc/UTC` on our instances. So the Redacted Data figure and its chart are bucketed on **UTC day boundaries** regardless of the date string we send; the client can only choose *which* UTC day. For a UTC-3 viewer a redaction at 22:00 local falls into the next UTC day and won't appear under "today" until tomorrow.

Fixing this needs a gateway change — accept RFC3339 (or an explicit offset) on `/reports/sessions`, as `/sessions` already does. Out of scope here; flagged rather than silently absorbed.

## 📣 User-facing impact

The Dashboard is now part of the new interface: it loads faster, and its numbers are correct in three places they previously weren't — the "today" figures no longer disagree with each other outside UTC, the 7d/14d/30d/3m redaction views now include today, and approved vs rejected reviews are finally distinguishable at a glance.

## 🔗 Related Issue

[EVL-119](https://linear.app/hoophq/issue/EVL-119) — B1.1 Dashboard

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [x] 🎨 Style/UI update
- [x] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- Added `webapp_v2/src/pages/Dashboard/` — page, page-scoped store, date/aggregation utils, and four page components (`ChartCard`, `StatBlock`, `ComingSoon`, `RangeFilter`).
- Added `components/{BarChart,DonutChart,SegmentedControl}` and `CHART_SERIES_COLORS` in `theme.js`.
- Added `services/reports.js` and `services/reviews.js`.
- Registered `/dashboard` in `Router.jsx` behind `ProtectedRoute adminOnly`, lazily loaded.
- Fixed three data quirks in dedicated commits (timezone, range end date, empty buckets).
- Deleted the ClojureScript dashboard and its now-dead `recharts` / `cljs-time` / `--chart-*` dependencies.
- Updated `CONTEXT_MIGRATION.md`, `COMPONENTS.md` and `MIGRATION_ROADMAP.md`.

### Setup

```bash
# 1. Postgres (skip if you run your own)
make run-dev-postgres

# 2. Gateway + agent (needs .env — cp .env.sample .env first time)
make run-dev

# 3. Frontend. webapp needs Node 22 — re2 fails to build on 25.
nvm use 22
cd webapp_v2 && npm install && npm run dev:full
```

Open http://localhost:5173 and log in as an **admin**. `/dashboard` is
admin-only, and `ProtectedRoute` bounces admins with zero connections into
onboarding — so create at least one resource role first, otherwise you never
reach the page.

To see anything but empty charts you need some history: run a few sessions
(`/client`), approve and reject a couple of reviews, and have Live Data Masking
enabled on at least one connection.

Grab a token for the API checks below:

```bash
# In the browser console:
copy(localStorage.getItem('jwt-token'))
# Then:
export T=<paste>
```

---

### 1. The page renders in React, not ClojureScript

1. Navigate to http://localhost:5173/dashboard.
2. Open DevTools → Elements.

**Expected**
- The page shows: `Dashboard` title → `Today's overview` card with three figures
  (Sessions / Reviews / Redacted Data) → a 2-column row (`Sessions` "Coming
  soon" | `Reviews` chart) → full-width `Redacted Data` chart → a 2-column row
  (`Runbooks` "Coming soon" | `Resource Roles` donut).
- There is **no** `<div id="app">` in the DOM — that element only exists when the
  CLJS bundle is mounted. Exactly one sidebar, no duplicated chrome.
- Network tab shows a separate JS chunk (~436 kB / 120 kB gzipped) loading on
  first visit to this route and **not** on any other route — that's the lazily
  loaded recharts bundle.
- Sidebar "Dashboard" is highlighted. `⌘K` → "Dashboard" navigates without a full
  page reload (no white flash, no new document request).

### 2. The three headline numbers are correct

Compare against the API directly. Substitute today's local date for `$TODAY`
and tomorrow's for `$TOMORROW`.

```bash
export TODAY=$(date +%F)
export TOMORROW=$(date -v+1d +%F)          # GNU: date -d tomorrow +%F

# Sessions — read `.total`
curl -s -H "Authorization: Bearer $T" \
  "http://localhost:8009/api/sessions?start_date=$(date -v0H -v0M -v0S -u +%Y-%m-%dT%H:%M:%SZ)&limit=1" | jq .total

# Redacted Data — read `.total_redact_count`
curl -s -H "Authorization: Bearer $T" \
  "http://localhost:8009/api/reports/sessions?start_date=$TODAY&end_date=$TOMORROW" | jq .total_redact_count

# Reviews — count today's, any status
curl -s -H "Authorization: Bearer $T" http://localhost:8009/api/reviews \
  | jq --arg d "$TODAY" '[.[] | select(.created_at | startswith($d))] | length'
```

**Expected** the three cards match the three numbers. A card whose request
failed shows an em dash (`—`), never `0` — "unknown" must not read as "none".

### 3. The two range filters are independent

1. Set the **Reviews** range to `30d`.
2. Set the **Redacted Data** range to `24h`.

**Expected**
- Each card's subtitle tracks its own selector: Reviews reads
  `Jun 30th - Jul 30th`, Redacted Data reads `Jul 30th` (a one-day range shows a
  single date, not a bogus two-date span).
- Changing one range never moves the other.
- Changing the **Reviews** range fires **no** network request — the review list
  is fetched once and every range is derived from it. Watch the Network tab.
- Changing the **Redacted Data** range fires exactly one `/api/reports/sessions`
  call with the expected `start_date`/`end_date`.

### 4. Rapid clicking cannot leave a stale chart

Click across the Redacted Data ranges as fast as you can: `24h → 3m → 24h → 7d`.

**Expected** the chart settles on **7d** and its subtitle says
`Jul 23rd - Jul 30th`. If an earlier, slower response won, you'd see 3m's data
under a 7d label — that's the bug the request token prevents.

### 5. Chart specifics

**Reviews chart** — hover a bar.
- Approved bars are green, rejected are red (the legacy chart drew both in
  indigo; see the PR description).
- The tooltip label is a formatted date: `Jul 30, 2026`, not `2026-07-30`.
- No x-axis labels (identification is via the tooltip), no grid.

**Redacted Data chart** — this one **does** have an x-axis.
- Each bar is labelled with its info type on the axis, single indigo series.
- A multi-underscore type renders fully hyphenated:
  `us-social-security-number`, **not** `us-social_security_number`. This is the
  `replaceAll` fix — a single JS `.replace()` only swaps the first underscore.
- Bars are sorted largest-first, and the order is stable across reloads (the
  server query has no `ORDER BY`).

**Resource Roles donut**
- One slice per connection subtype; connections with an empty subtype fold into
  a slice named `others`.
- Hovering shows **only** the hovered slice, not every slice at once.

### 6. Empty and error states

- Pick a range with no data (a fresh org, or `24h` before running anything):
  the card body reads `No data found for the selected period` in grey.
- Simulate a failure — DevTools → Network → block request URL `*/api/reviews`,
  then reload.

**Expected** only the Reviews card degrades: it shows
`Failed to load reviews.` in red and the Reviews figure shows `—`. Sessions,
Redacted Data and Resource Roles still render normally. One dead endpoint must
not take the page down.

### 7. Free-license gating (negative path)

Either point at a free-license gateway, or force it in the console:

```js
// Temporarily flip the store
window.__z = undefined // no-op; use React DevTools → useUserStore → isFreeLicense = true
```

**Expected**
- A blue callout appears under the title: *"Organizations with Free plan have
  limited reporting. Upgrade to Enterprise to have unlimited access to the
  Dashboard."* with a working "Contact our Sales team ↗" link.
- On **both** range filters, `24h / 14d / 30d / 3m` are dimmed and `7d` stays
  selectable.
- **Hovering a dimmed option shows the tooltip** *"Available on Enterprise plan
  only."* — this is the specific regression to watch. Mantine's `disabled` sets
  `pointer-events: none`, which would silently suppress the tooltip and leave
  the user with a greyed-out option and no explanation.
- Clicking a dimmed option does nothing: the selection stays on `7d` and no
  request is fired.

### 8. Non-admin (negative path)

Log in as a standard user and go to `/dashboard` directly.

**Expected** immediate redirect to `/`. The sidebar and `⌘K` do not list
Dashboard at all.

### 9. Quirk fixes — the behaviour that intentionally differs from v1

- **Timezone.** In a non-UTC zone (`TZ=America/Sao_Paulo`), between local
  21:00 and midnight: the Sessions figure and the Redacted Data figure now
  describe the *same* day. On `main` they disagree, because the legacy code read
  Sessions off UTC midnight and Redacted off the local date.
- **Ranges include today.** Trigger a redaction now, then set Redacted Data to
  `7d`. The new redaction appears. On `main` it does not — `end_date=<today>`
  resolved to today-at-midnight, so every range except `24h` silently ended
  yesterday.
- **Empty buckets.** Create a review and leave it `PENDING`, with no
  approved/rejected review on the same date. That date does **not** appear on the
  Reviews chart. On `main` it renders as an invisible zero-height bar that still
  occupies a slot and still answers the tooltip.

---

### Regression check

The CLJS cleanup deleted the dashboard half of `events/reports.cljs` but kept
the per-session report. That's the part most likely to break:

1. **Session detail report** — go to `/sessions`, open a session that ran against
   a connection with Live Data Masking. Expected: the session detail renders and
   the data-masking analytics panel shows its redaction breakdown. This exercises
   `:reports->get-report-by-session-id`, `:reports->session` and
   `:reports->clear-session-report-by-id`, all deliberately retained.
2. **CLJS bundle still loads clean** — visit any CLJS route (`/sessions`,
   `/resources`, `/client`). Expected: page renders, and the browser console has
   no `No such namespace` / `Use of undeclared Var` errors from the removed
   `webapp.dashboard.*` and `webapp.components.charts` namespaces.
3. **Sidebar navigation elsewhere is unaffected** — the CLJS `main-routes` vector
   lost its Dashboard entry; confirm Resources / Terminal / Runbooks still
   navigate.

### Build gates

```bash
cd webapp_v2 && npm run build      # ✓ built; Dashboard emits its own ~436 kB chunk
cd webapp_v2 && npm run lint       # ✖ 16 problems (15 errors, 1 warning) — unchanged baseline
                                   # all in vite.config.js, CommandPaletteRoot,
                                   # Auth/Login, EventRouting/Form. None in touched files.

nvm use 22
cd webapp && rm -rf .shadow-cljs && npm ci && npm run release:hoop-ui
                                   # [:hoop-ui] Build completed. (2568 files, 2479 compiled, 0 warnings)
                                   # The 12 postcss-calc warnings in the CSS step are pre-existing
                                   # (verified identical before and after this branch's CSS change).
```

### Not covered locally

Nothing in this change requires enterprise libhoop, an IDP or cloud resources.
The only path that needs a specific environment is **step 7**, which needs
either a free-license gateway or a manual `isFreeLicense` override via React
DevTools.

---

## Rebase note (review pass, 2026-07-30)

This branch was cut from `6e753824`, before **EVL-115 / #1653** (`PR 0.2 — scaffolding deletion + doc reorg`) merged. It has been rebased onto current `main` and the four resulting conflicts resolved. The six commits are unchanged and still split one-per-behaviour-change.

| Conflict | Resolution |
|---|---|
| `webapp_v2/src/pages/Dashboard/index.jsx` (modify/delete) | EVL-115 deleted the 7-line unrouted stub; this branch modified it. Kept this branch's version — it is the real page now. |
| `webapp_v2/COMPONENTS.md` | The service **table** this branch wrote into no longer exists (EVL-115 replaced it with a filesystem pointer + non-obvious notes, because the old tables listed 6 of 27 services and disagreed across files). The two notes were re-targeted into the notes list, keeping the `end_date`-is-exclusive finding and the `/reviews` unbounded / no-query-params quirk. |
| `webapp_v2/CONTEXT_MIGRATION.md` | Same cause. The routing-table row for `/dashboard`, the removal from the CLJS route list, and the Done bullet all survive unchanged; the additions to the deleted stores/services tables were absorbed into `COMPONENTS.md` above. |
| `webapp_v2/MIGRATION_ROADMAP.md` | **This branch's rewrite of the bidi constraint has been dropped** — see below. Only the `B1.1 Dashboard — Done` row remains. |

### On the bidi paragraph

This PR and **EVL-116 / #1655** (Track A1) independently discovered the same thing — `bidi/path-for` returns `nil` for an unknown handler rather than throwing — and both rewrote the same roadmap paragraph, with opposite operational conclusions: #1655 instituted a mandatory grep gate and kept the PR ordering; this PR concluded the ordering is merely advisory.

Neither rewrite conflicted with `main`, so the collision was invisible to both: whichever merged second would have hit the conflict, and resolving it the wrong way would have silently dropped the grep gate right before A2/A3 execute.

The two positions were reconciled in #1655: **the gate, not PR ordering, is the safety mechanism**, so a route entry and its sidebar entry may ship in the same PR — citing this PR's `:dashboard` removal as the precedent. The finding here was correct and is preserved; only the duplicate prose was removed, so the paragraph has one author.

### Cross-PR note

`LayoutDashboard` in the `lucide-react` refer of `shared_ui/sidebar/constants.cljs` had two consumers: the `"Dashboard"` icon key (removed here) and `"infrastructure"` (removed by #1655). Each PR is correct alone; once both merge the refer is orphaned. It is recorded in the roadmap's A3 trap list and on EVL-118 to be swept there — deliberately **not** fixed in either PR, since removing it in one would break the other.

### Re-verified after the rebase

- `merge-tree` against `main`: zero conflicts
- `npm run build` in `webapp_v2`: green


sandromello:
✅ Build Completed with Success, Version=1656.0.0-g6e53425

sandromello:
✅ Build Completed with Success, Version=1656.0.0-g0e713a8

sandromello:
✅ Build Completed with Success, Version=1656.0.0-ga601e7c

rogefm:
Thanks — three of the four are real and are fixed in `08b1670a`.

**2. Redaction range off-by-one — fixed.** Correct, and it was mine rather than pre-existing: moving `end_date` to tomorrow so today would be included stretched every range without moving the start. The legacy window *was* N days, just shifted one day into the past. `rangeStart` now returns `today - (days - 1)`, which also removes the 24h special case — a window ending at tomorrow's midnight starts N-1 days back for every N.

```
 1d -> 2026-07-31 .. 2026-08-01  =  1 date   "Jul 31st"
 7d -> 2026-07-25 .. 2026-08-01  =  7 dates  "Jul 25th - Jul 31st"
30d -> 2026-07-02 .. 2026-08-01  = 30 dates  "Jul 2nd - Jul 31st"
```

**3. Review cutoff — fixed.** Also correct. The rolling `now - N days` cutoff was carried over from the ClojureScript original, but the subtitle now describes a calendar range, so the bars disagreed with their own label and "24h" reached into yesterday. It uses the same `[rangeStart, tomorrow)` bounds as the Redacted chart now, so both charts on the page agree with each other and with their labels.

**4. DST-unsafe day arithmetic — fixed.** Real, and worth the catch. Checked against Toronto's 8 March spring-forward:

```
start of DST day     : 2026-03-08, 00:00:00
addDays(+1)          : 2026-03-09, 00:00:00   ← setDate
+86400000 ms (before): 2026-03-09, 01:00:00   ← one hour off
```

`addDays` now uses `setDate`; `MS_PER_DAY` has no callers left.

**1. Relative imports — not taken.** `@/` is the convention for crossing the tree, which this code already does for stores, components and services. Importing colocated page modules and `./api` relatively is the established pattern here, not a deviation: all 26 existing files in `src/services/` use `import api from './api'`, and `JiraTemplates`, `Features/DataMasking`, `Agents` and `Settings/AuditLogs` all import their own `./components/` and `./sections/` relatively. There are 165 relative imports in `src/` today. Switching just this page would make it the outlier.

sandromello:
✅ Build Completed with Success, Version=1656.0.0-gba85d97

sandromello:
✅ Build Completed with Success, Version=1656.0.0-g08b1670

luanlorenzo (approved):

PR #1661, opened 2026-07-31
feat(mcp): protocol-aware MCP gateway as a first-class connection type

matheusfrancisco:
# Body

> [!IMPORTANT]

## 📝 Description

Adds a new connection type `mcpproxy` (ADR-0004, `docs/adr/0004-mcpproxy-connection-type.md`) that embeds the `github.com/hoophq/mcpproxy` gateway in the agent and parses every MCP JSON-RPC message instead of relaying MCP as opaque HTTP.

The existing `mcp` subtype stays exactly as it is — an httpproxy alias, byte relay, zero regression risk. The two coexist until the new path proves out.

Topology:

```
MCP client → hoop gateway (auth, RBAC, proxyproto) → gRPC packets
  → agent controller (packetQueue per sid:connID)
  → libhoop/agent/mcpadapter (byte stream ⇄ in-memory HTTP listener)
  → mcpproxy gateway (policy, budgets, rug-pull, S2C gate, guardrail/masking hooks)
  → backend: stdio child | client-hosted stdio | streamable-http | legacy SSE
```

Because the pipeline sees `tools/call` rather than bytes, three things become expressible that the relay cannot do: tool-level allow/deny (denied tools are stripped from `tools/list` before the model sees them), rug-pull detection (SHA-256 fingerprint over tool name + description + input schema), and a structured per-tool-call audit trail. Guardrails and masking now run on the free-text leaves — `tools/call` arguments, tool descriptions, `result.content[].text` — instead of a flattened method+URL+headers+body string, so the JSON-RPC envelope can never be corrupted by a redaction.

`MCP_TRANSPORT=client-stdio` is the notable transport: the MCP server process runs on the developer's own machine through `hoop connect` (their working tree, their SSH agent, their authenticated CLIs), while every inspection stage stays in the agent. Wire protocol is one packet pair — `MCPStdioRequest` / `MCPStdioReply` / `MCPStdioClose`. `Send` returns on the write ack, not on the MCP response: responses arrive asynchronously on `Recv` and are matched by JSON-RPC id, so waiting inside `Send` would deadlock on any server that interleaves notifications.

The whole resource card is gated behind `experimental.mcp_gateway`, default off. Creation only — an org that turns the flag back off keeps every connection it already created, and the role form, edit view and Connect modal stay reachable.

## 📣 User-facing impact

Hoop can now broker MCP servers with protocol awareness instead of relaying them as opaque HTTP: allow or deny individual tools, detect a server that silently changes a tool's description mid-session, block server-initiated sampling and elicitation, and review a session as a tool-call timeline rather than HTTP blobs. MCP servers that need a developer's own machine (filesystem, git, shell) run locally through `hoop connect` while every call is still inspected centrally. Gated behind the `experimental.mcp_gateway` flag.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

**Protocol / shared**

- `common/proto`: `ConnectionTypeMcpProxy`, the `MCPProxyConnectionWrite` packet pair, and the `MCPStdioRequest` / `MCPStdioReply` / `MCPStdioClose` trio with their spec keys (`mcp.event`, `mcp.stdio_backend`, `mcp.stdio_request_id`, `mcp.stdio_error`, `mcp.stdio_command`, `mcp.stdio_env`).
- `ToConnectionType` resolves the `mcpproxy` subtype under the `httpproxy`, `application` and `custom` parents — without the httpproxy arm an mcpproxy connection silently degraded to the byte relay and never reached the adapter.
- `common/featureflag`: `experimental.mcp_gateway`, default off, gateway component.

**Agent**

- `agent/controller/mcpproxy.go`: the MCP gateway per session, its own `mcpProxyQueues` FIFO (a tool call parked on a human review can hold a worker for minutes and must never delay an httpproxy connection multiplexed on the same agent), and `sessionCleanup` teardown.
- The gateway is memoised **per session**, not per `sid:connID`. The proxy listener mints a connection id per HTTP request, so the previous keying meant every message after `initialize` hit a gateway that had never issued the `Mcp-Session-Id` the client presented.
- `agent/controller/mcpstdio.go`: the client-hosted stdio backend swapped into `gateway.Options.Backends` (mcpproxy needs no change — `Backends` is a `map[string]backend.Factory` and `Factory` is a bare closure).
- `parseConnectionEnvVars` reads `MCP_TRANSPORT`, `MCP_AUTH`, `MCP_ALLOWED_TOOLS` / `MCP_DENIED_TOOLS` / `MCP_APPROVAL_TOOLS`, `MCP_MAX_CALLS`, `MCP_MAX_RESULT_KB`, `MCP_ON_RUG_PULL`, `MCP_BLOCK_SAMPLING`, `MCP_BLOCK_ELICITATION`, plus the `MCPENV_*` carve-out for the stdio child's environment (every hoop secret already arrives as `envvar:NAME`, so that prefix cannot distinguish the child's env from the connection's own settings).
- `_libhoop`: OSS stubs `NewMCPProxy` / `NewMCPHooks` and the `mcpadapter.Hooks` type, so `make libhoop-map` keeps building without the enterprise library — with the hooks absent rather than faked. **Both stubs fail closed on `guard_rail_rules`** (`CheckGuardRailEnforcement(rules, "mcpproxy")`), matching the enterprise implementation and ADR-0004 §DEP-48. This is a real behavior change for OSS builds: a connection carrying guardrail rules is now refused with the DEP-48 message instead of the generic missing-library one, and `NewMCPHooks` refuses rather than handing the gateway a pipeline whose `GuardInput` and `Redact` stages are silently absent. It is also what makes `TestFailedMCPGatewayBuildDoesNotEvictAHealthySuccessor` meaningful under `make test-oss` — with the stub returning `Hooks{}, nil` the build it asserts on could never fail.

**Gateway**

- `proxyproto/httpproxy`: packet types parameterized per session (`agentWriteType()` / `clientWriteType()`, zero value = httpproxy). Framing is identical raw HTTP bytes; only the agent-side dispatch differs.
- Proxy-token bootstrap now recognises the `mcpproxy-` key prefix alongside `httpproxy-`, otherwise a `/<token>` URL was read as a request path and 404'd.
- Structured protocol events (`mcp.event`) are recorded and dropped before the response stream — injecting one would corrupt the client's HTTP framing.
- Credential issuance: `mcpproxy` gets the `/mcp` endpoint and header-auth command instead of the browser bootstrap URL, which would have handed the user a link that 404s. Shared `httpProxyPublicOrigin` helper extracted from the two duplicated TLS/host resolutions.
- `plugins/audit`: `MCPProxyConnectionWrite` in both directions, with `mcp.event` marked in the event metadata.
- `GET /mcp-catalog` serves mcpproxy's embedded catalog (~32 hosted servers) so the create form can pre-fill `REMOTE_URL` / `MCP_TRANSPORT` / `MCP_AUTH`. Static build-time data, no tenant content, read-only role.

**Client**

- `hoop connect` gains an `mcpproxy` case: the HTTP front door (identical to httpproxy, different packet type) plus a second, port-less owner for client-hosted stdio that spawns nothing until the agent asks. Children run in their own process group and are reaped on `MCPStdioClose`, session close, or CLI exit.

**Webapp**

- `mcpproxy-role-form` in `resources/setup/roles_step.cljs`: server picker fed by `/mcp-catalog`, transport selector, and a tool-policy section.
- Auth mode is the admin's choice, not the catalog's. GitHub, Linear and Stripe all accept either an OAuth login or a PAT, so the catalog seeds the mode and the admin overrides it. Static tokens ride in the header the provider actually expects — Context7 wants `CONTEXT7_API_KEY`, Google Maps wants `X-Goog-Api-Key`; upper-casing or hyphenating the name authenticates as nobody.
- Env emission in `process_form.cljs`: stdio commands go to the connection's command array (not an env var), stdio secrets get the `MCPENV_` prefix rather than `HEADER_` (a subprocess never sees an outbound HTTP header), and the three UI-only keys (`mcp_server`, `mcp_auth_mode`, the header-name bookkeeping) never reach the wire.
- New `:resource-setup->remove-role-credential` event: switching auth mode has to *forget* the previous credential. Blanking is not the same — `config->json` filters blanks out of the payload but the key stays in app-db and reappears on the next form read.
- Connect modal renders `mcp-proxy-credentials-fields`; without it, mcpproxy fell through to the postgres default and showed empty host/username/password fields.
- `download-connection-metadata.sh` gains `SKIP_CONNECTION_METADATA_DOWNLOAD=1` so you can build against a `connections.json` that has not merged into `hoophq/documentation@main` yet. Opt-in only; CI and release builds keep pulling the published metadata. README updated to match.
- Deleted `webapp/test/webapp/core_test.cljs` (a `deftest fake-test` asserting nothing).

**MCP OAuth credential lifecycle**

A brokered OAuth login is frozen into `HEADER_AUTHORIZATION` at save time. On its own that credential dies when the provider expires it, so `mcp_oauth_flow_id` on the save payload asks the gateway to adopt the login into a durable grant it renews at every session open.

Adoption can legitimately refuse — the important case is an admin who authorizes one MCP server and then edits the URL before saving, where adopting would attach that server's refresh token to a different server and auto-renew it forever. The refusal must not fail the save: the connection row is already written and works exactly as before grants existed. So it rides back on the response as `mcp_oauth_warning`.

Nothing read that field. Every save path showed the same plain "created!" / "updated!" toast it shows for a healthy save, which is the worst possible outcome for this failure mode — the connection looks fine for days and then dies at the provider's TTL, with the failing session pointing at nothing. Fixed on all four save paths:

- `PUT /connections/:name` via `:resources->update-role-connection` and `:connections->update-connection` — both callbacks took `(fn [])` and discarded the response entirely.
- `POST /connections` via `:connections->create-connection`.
- `POST /resources` (the create wizard) — this one was broken on the **server** too: `ResourceRoleRequest` had no `mcp_oauth_flow_id` field, so Gin dropped it and no adoption was ever attempted for a role created through the wizard. Added the field, the adoption call (`apiconnections.AdoptMCPOAuthGrant`, exported for this), and `mcp_oauth_warnings []MCPOAuthRoleWarning` on `ResourceResponse` — per role, because the wizard creates several at once and the admin has to know *which* one is degraded.
- `:add-role->all-roles-processed` — the add-role flow creates roles one at a time and summarizes, so the warning has to survive the aggregation.

Warnings render as a distinct `:warning` toast carrying the gateway's reason. That level was already dispatched at two call sites (`parallel_mode/events/selection.cljs`, `add_role/events.cljs`) but `show-snackbar`'s `case` had no `:warning` clause **and no default**, so a cljs `case` fallthrough threw "No matching clause" and took down the caller's event handler. Added the level end to end (`toast-warning`, amber `AlertTriangle`) plus a default clause, so an unknown level can never again turn a toast into an exception.

### Test Configuration:
- **Browser(s)**: Chrome Headless 151 (karma)
- **OS**: macOS (darwin arm64)

### Commands

```sh
# agent — MCP gateway, session affinity, client-hosted stdio, webapp/agent env contract
cd agent && go test ./controller/ -run 'MCP|Mcp' -count=1

# gateway — catalog conversion, credential response, packet-type parameterization
cd gateway && go test ./api/connections/ -run 'MCP|Mcp' -count=1
cd gateway && go test ./proxyproto/httpproxy/ -count=1

# gateway smoke suite — MCP OAuth grant lifecycle end to end against a fake provider
make test-gateway

# webapp
cd webapp && npx shadow-cljs compile karma-test && npx karma start --single-run

# all three Go modules build (including the OSS libhoop stubs)
cd gateway && go build ./... && cd ../agent && go build ./... && cd ../client && go build ./...
```

All green: Go unit suites across `./gateway/... ./agent/... ./common/...` pass, `gateway/integration` `ok 29.5s`, karma `76/76 SUCCESS`, `BUILD_OK`.

### What the tests actually defend

- **Session affinity** (`mcpproxy_session_test.go`) — the bug that motivated the memoisation: `TestMCPSessionSurvivesAcrossRequests` drives `initialize` then a second request over a *different* connection id and asserts the `Mcp-Session-Id` is still valid. `TestMCPGatewayIsMemoisedPerSession` and `TestSessionCleanupClosesMCPGateway` pin the lifetime at both ends.
- **Client-hosted stdio** (`mcpstdio_test.go`) — round trip against a real stdio MCP server (`testdata/fakemcp.js`), `MCPENV_*` reaching the child's environment, a notification not blocking the pipeline, child reaping on close (including close-before-spawn), and spawn failure surfacing as a JSON-RPC error rather than a 30-minute timeout.
- **Webapp ⇄ agent env contract** (`mcpproxy_contract_test.go`) — the exact env maps the CLJS form emits, parsed by the agent's real `parseConnectionEnvVars`. A typo in a key name is not an error anywhere in the running system; it is a connection that fails at connect time, long after the admin saved it.
- **Header verbatim-ness** (`TestMcpBackendHeadersStripsOnlyThePrefix`, `a-provider-named-header-reaches-the-wire-verbatim`) — a rewritten header name is an unauthenticated request with no warning.
- **Credential URLs** (`connection_mcp_credentials_test.go`) — mcpproxy advertises `/mcp`, and httpproxy keeps its browser bootstrap URL untouched.
- **OAuth grant lifecycle** (`gateway/integration/mcp_oauth_grant_test.go`) — adoption against a fake provider, refusal on a cross-server flow, grant deletion when the provider answers `invalid_grant` (it used to be rolled back with the error, so a dead refresh token was replayed forever), a callback replay not downgrading a good flow, and — new here — the same adoption and the same named refusal for a role created through `POST /resources`.
- **OAuth warning surfacing** (`webapp/test/webapp/resources/mcp_oauth_warning_test.cljs`) — all four save paths, both branches each: a refusal produces a `:warning` toast carrying the gateway's reason and *no* success toast, a clean save produces exactly one `:success`. The negative half matters as much as the positive: a `:success` toast on a refused adoption is precisely the defect.
- **Flag gate** (`helpers_test.cljs`) — hidden when the flag is absent (the default state, so getting it wrong ships the feature to everyone), hidden when false, visible when true, string and keyword flag keys both accepted, and no other card affected.

Every new test above was mutation-checked: reverting the fix it defends (dropping `mcp_oauth_flow_id` from the binding, stubbing the adoption call, blanking the warning read) makes it fail, and only it.

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass (agent ⇄ mcpproxy gateway ⇄ real stdio MCP server over the tunnel)

## ✅ Checklist

## 📄 Additional Notes

**Known gaps, deliberately out of scope** (tracked in the ADR's build order):

1. **Reviews.** The `mcp-tool-call` review type is not built yet, so `MCP_APPROVAL_TOOLS` matches degrade to **deny** — fail closed, never silently allow.
2. ~~**OAuth refresh.**~~ **Done.** `MCP_AUTH=oauth` logins are adopted into a durable grant (`services.AdoptMCPOAuthGrant`) and the access token is renewed from its refresh token at every session open (`resolveMCPOAuthForSession`). Adoption is refused when the flow authorized a different server than the connection points at — otherwise editing the URL field after authorizing would durably attach one server's refresh token to another and auto-renew it forever. The refusal does not fail the save (the connection works on its frozen header), so it is reported on the response and surfaced in the UI; see "MCP OAuth credential lifecycle" below.
3. ~~**Edit renderers.**~~ **Done.** `credentials_tab.cljs` routes the `mcpproxy` subtype to `mcpproxy-edit-form`, so transport, tool policy and re-authorization are all reachable after creation. Covered by `webapp/test/webapp/resources/configure_role/mcpproxy_edit_test.cljs`.
4. **Session timeline.** The agent emits one JSON audit record per protocol event and the gateway records it, but the `mcp.event` marker is dropped between the WAL and both the SSE stream and the persisted blob (`eventbroker.Event` and `sessionStreamEvent` are three-field structs). Every event name is `mcp.`-prefixed, so the viewer can sniff the payload without a backend change.

**Where to focus review:**

- `gateway/proxyproto/httpproxy/httpproxy.go` — the packet-type parameterization touches the live httpproxy path. The zero value is httpproxy, so an unmodified session behaves identically, but this is the one file where a mistake affects existing connections.
- `agent/controller/mcpstdio.go` — the ack-not-response contract on `Send`, and process-group reaping.
- `webapp/.../process_form.cljs` — which form keys become env vars, which become the command array, and which are dropped.

**Dependency weight:** mcpproxy pulls prometheus/otel into the agent build. If footprint becomes a concern the telemetry package can be split into its own Go module; the seams already isolate it.


sandromello:
✅ Build Completed with Success, Version=1661.0.0-g33fd0a1

sandromello:
✅ Build Completed with Success, Version=1661.0.0-gfb6b01e

sandromello:
✅ Build Completed with Success, Version=1661.0.0-g84e8d9b

sandromello:
✅ Build Completed with Success, Version=1661.0.0-gb8655dd

sandromello:
✅ Build Completed with Success, Version=1661.0.0-gb862469

sandromello:
✅ Build Completed with Success, Version=1661.0.0-g19c8a43

EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1661.0.0-g9965f7a

sandromello:
✅ Build Completed with Success, Version=1661.0.0-gcc90d46

matheusfrancisco:
@copilot resolve the merge conflicts in this pull request

sandromello:
✅ Build Completed with Success, Version=1661.0.0-gba0eec7

sandromello:
✅ Build Completed with Success, Version=1661.0.0-g1dea618

sandromello:
✅ Build Completed with Success, Version=1661.0.0-g3a59020

luanlorenzo (approved):

sandromello:
✅ Build Completed with Success, Version=1661.0.0-gba32b66

PR #1664, opened 2026-08-03
EVL-159: add Features Indicator to the Terminal toolbar

rogefm:
## 📝 Description

Adds a persistent indicator to the Terminal toolbar showing which hoop features are active on the selected resource role, so users know what will happen before they hit Run.

This is the **Before Execution** half of [Major Features Visibility](https://linear.app/hoophq/project/major-features-visibility-3c3c56351db3). The design deliverables for it were already marked Done (`Before Execution - X` tickets in that project) but no implementation existed.

The core of the change is `webclient/features.cljs` — a catalog that is the single source of truth for what a "feature" is. Each entry carries name, description, icon, category and an `:active?` predicate over the same context map, so resolving the active set is a `filterv` instead of a growing `cond`. The real-time callouts (EVL-160) reuse this catalog, which keeps naming and ordering from drifting between the two surfaces.

Two decisions worth calling out for review:

- **The Access Requests signal is not new.** It reuses `(or :jit_access_duration_sec (seq :reviewers))` — exactly what `connections/native_client_access/events.cljs:114-115` already uses to decide a connection requires review. One question, one rule.
- **No backend work.** `openapi.Connection` already exposes `reviewers`, `guardrail_rules`, `redact_enabled`, `jira_issue_template_id` and `mandatory_metadata_fields`, and the webclient already dispatches `:ai-session-analyzer/get-role-rule` on connection change.

## 📣 User-facing impact

Before running a query, you can now see at a glance which protections are active on the resource role you selected — access approval, guardrails, required metadata, live data masking, AI analysis or Jira ticket creation. Hover the toolbar indicator for the list, click it for a description of what each one does in this session.

## 🔗 Related Issue

[EVL-159](https://linear.app/hoophq/issue/EVL-159)

## 🚀 Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## 📋 Changes Made

- **New** `webapp/src/webapp/webclient/features.cljs` — feature catalog and the `active` resolver
- **New** `webapp/src/webapp/webclient/components/features_indicator.cljs` — `HoverCard` with the active list plus the "Features Active" modal
- `webapp/src/webapp/components/notification_badge.cljs` — optional `:badge-color` (defaults to red, so both existing callers are unaffected) to reuse the component for the green dot instead of duplicating the button-plus-dot markup
- `webapp/src/webapp/webclient/components/header.cljs` — renders the indicator before Parallel Mode

### Test Configuration:
- **Browser(s)**: Chrome
[...]

PR #1665, opened 2026-08-03
EVL-172 + EVL-159: show which features are active on a resource role in the Terminal

rogefm:
## 📝 Description

Shows users which hoop features are active on the selected resource role **before** they run a query, and fixes the API gap that made the first attempt at this report nothing.

### Why the API had to change first

The connection payload reported **stored** configuration, not **effective** configuration.

Guardrails, data masking and access requests can be attached either directly to a connection or through an **attribute** — the mechanism Protection Profiles use — and the runtime enforces the union of both. The response only ever reflected direct associations, so a connection governed by a protection profile came back with empty `guardrail_rules` while guardrails were in fact being enforced on it.

Two of the fields were wrong **even with no profile involved**:

- `jit_access_duration_sec` used the name-only resolver and only ever looked up the `jit` access type. Anything that *executes* resolves the `command` rule instead (`gateway/transport/interceptors/accessrequest/accessrequest.go:193-196`), so a plain name-targeted command rule never surfaced.
- `redact_enabled` is read by **no runtime code in the gateway** — grepping `RedactEnabled` returns only the model field and `ToOpenApi`. Real masking is `services.GetDataMaskingRulesForConnection`, so the field also missed directly-attached masking rules.

A frontend-only fix was not viable: every endpoint that would close the gap is admin/auditor-gated, so a standard user sitting in the Terminal gets 403, and `GET /attributes` does not expose `access_type`.

## 📣 User-facing impact

Before running a query you can now see at a glance which protections are active on the resource role you picked — approval, guardrails, required metadata, live data masking, AI analysis or Jira ticket creation. The indicator sits next to the role name; hover it for the list, click it for what each one does in this session.

## 🔗 Related Issue

[EVL-172](https://linear.app/hoophq/issue/EVL-172) (backend) and [EVL-159](https://linear.app/hoophq/issue/EVL-159) (indicator). Supersedes #1664, which read the stored fields.

**Scope note:** this branch also built the post-Run callout stack ([EVL-160](https://linear.app/hoophq/issue/EVL-160)) and then removed it — three cards repeating what the indicator already says, on every Run, with less information than its hover popup. EVL-160 is cancelled. The intermediate commits are still on the branch; squash on merge.

## 🚀 Type of Change

## 📋 Changes Made

**Backend**
- `gateway/api/openapi/types.go` — `ConnectionEffectiveFeatures` and `ConnectionAccessRequestFeatures`; `Connection.effective_features` (read-only, detail endpoint only)
- `gateway/api/connections/connections.go` — `resolveEffectiveFeatures` composes the enforcement-path resolvers in an errgroup; `Get` fills the object and fixes `jit_access_duration_sec` to use the attribute-aware resolver
- Regenerated OpenAPI docs

**Frontend**
- **New** `webapp/src/webapp/webclient/features.cljs` — the feature catalog and the `active` resolver
- **New** `webapp/src/webapp/webclient/components/features_indicator.cljs` — `HoverCard` with the active list plus the "Features Active" modal
- `webapp/src/webapp/webclient/events/primary_connection.cljs` — selecting from the dialog now also fetches the connection detail
- `webapp/src/webapp/components/notification_badge.cljs` — optional `:badge-color` and `:radius` (both default to current behavior, so the two existing callers are unaffected)
- `webapp/src/webapp/webclient/components/header.cljs` — renders the indicator beside the resource-role pill

### Four things worth a reviewer's attention

**The backend cannot drift from enforcement.** The resolvers are the ones the runtime already calls, composed exactly the way `gateway/analytics/segment.go:264-299` already composes them. No new resolution logic was written.

**Failure is non-fatal but not silent.** On error `effective_features` is left `null` rather than filled with `false`, because "nothing is active" and "we could not tell" must not look the same to a caller deciding whether to warn a user. This replaces an `err == nil && rule != nil` check that made a real DB failure indistinguishable from "no rule".

**The dialog fix is the same bug through a different door.** `connection_dialog.cljs:22` handed `set-selected` a row from the paginated list, which carries only the stored columns. Two entry points were producing different shapes of "selected connection", and reading `effective_features` off a list row would have seen nothing. Both paths now converge on `set-from-details`. The dialog already filters to `access_mode_exec` enabled, so that path's guard never fires here.

**`access_request` is split by access type on purpose.** `command` gates exec verbs and `jit` gates connect verbs; they are not interchangeable, and collapsing them into one boolean is exactly what produced the original bug.

### Test Configuration:
- **Browser(s)**: Chrome
- **OS**: macOS

### Tests performed:
- [x] Unit tests pass

`make test-oss` green. `npx shadow-cljs compile hoop-ui` — 0 warnings. `make generate-openapi-docs` regenerated and the new schemas verified present in `openapiv3.json`.

Verified against the dev database, where connection `postgres-demo` carries the `protection-permissive` profile:

| source | guardrails | data masking |
|---|---|---|
| direct (`*_rules_connections`) | 0 rows | 0 rows |
| via attribute (`*_rules_attributes`) | 1 | 1 |

Running the resolver query verbatim returns a real rule — the `UPDATE`/`DELETE`-needs-`WHERE` pattern match — while the payload's own subquery for the same connection returns `NULL`. There is also an access-request rule `twtw` with `access_type='command'` targeting `postgres-demo` by name, which the old payload did not surface at all.

So that connection goes from showing nothing to showing **Guardrails + Live Data Masking + Access Requests**.

To exercise the UI, open `/client` and pick a resource role with at least one feature configured:

1. The indicator appears beside the resource-role pill, circular, with a green dot.
2. Hovering lists every active feature with an "Active" badge.
3. Clicking opens the "Features Active" modal with one card per feature.
4. Switching to a role with nothing configured hides the indicator entirely.
5. Switching roles **while the modal is open** updates its contents — the modal subscribes rather than taking a snapshot.
6. Selecting from the **dialog** behaves the same as arriving via `?role=` in the URL. This is the regression the dialog fix addresses.

_Pending._

## ✅ Checklist

## 📄 Additional Notes

`mandatory_metadata` and `jira_templates` in the new object are projections of existing columns — no attribute-based path exists for either (the profile spec carries only guardrails, masking, access rules and analyzers). They are included so a consumer does not need to know which features happen to be plain columns; they are computed in the same function from the same struct, so they cannot disagree with the fields they mirror.

The enrichment is on the **detail** endpoint only — the list endpoint does not get the four extra queries.

Because `effective_features` now carries the analyzer, the frontend no longer needs the `:ai-session-analyzer/role-has-rule?` subscription to decide this. The `get-role-rule` dispatch stays where it is; only this consumer dropped.

**The hover surface diverges from the Figma on purpose.** The Figma draws a plain dark tooltip with one `{Feature}: Active` line per feature; we ship a popover with the same tiles the modal uses, because it reads better beside them. The tooltip is implemented behind a `hover-variant` keyword in `features_indicator.cljs`, so switching back is a one-word change. The unused branch should be deleted once that settles — flagging it rather than leaving it to be discovered.

**One non-obvious constraint if you touch the trigger:** Radix `HoverCard.Trigger` forces `asChild` and calls `requireReactElement`, so it needs a real DOM element to clone. A Reagent component does not forward refs, which is why there is a `[:span]` wrapping the badge.

Copy says **Live Data Masking**, following RD-231. The Feature Specs document still says "AI Data Masking" and is out of date.

No new env vars, so no helm chart changes.


sandromello:
✅ Build Completed with Success, Version=1665.0.0-g2a1c7c8

sandromello:
✅ Build Completed with Success, Version=1665.0.0-gac765c6

sandromello:
✅ Build Completed with Success, Version=1665.0.0-g9301613

EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1665.0.0-gaba4357

sandromello:
✅ Build Completed with Success, Version=1665.0.0-g8f489b2

sandromello:
✅ Build Completed with Success, Version=1665.0.0-ga0df522

sandromello:
✅ Build Completed with Success, Version=1665.0.0-gafca09e

sandromello:
✅ Build Completed with Success, Version=1665.0.0-g04e915a

sandromello:
✅ Build Completed with Success, Version=1665.0.0-g6519843

sandromello:
✅ Build Completed with Success, Version=1665.0.0-g094bb43

sandromello:
✅ Build Completed with Success, Version=1665.0.0-g6c04fd9

PR #1668, opened 2026-08-04
EVL-145 + EVL-161: Sessions list in React, and the backend fields it needs

rogefm:
> [!WARNING]
> **Do not merge without the session-details modal.** Shipping the list alone
> changes a user-facing behaviour — a row click opens a modal in v1, and would
> navigate to a full page here — and adding the modal later would change it a
> second time. Two breaks where there should be none. The full port of
> `session_details.cljs` lands on this branch before this is mergeable; scope and
> progress below.

## 📝 Description

Ships **EVL-145** (React Sessions list) and **EVL-161** (the gateway fields it needs) together.

`/sessions` and `/sessions/filtered` are now React. `/sessions/:id` deliberately gets **no** React route — it keeps falling through the catch-all to the CLJS dedicated page, which is already in production via Share links and self-fetches its own data (`app.cljs:514-521`). EVL-132 flips it. No feature flag; exposure is route registration.

On the backend, `models.ListSessions` never selected `s.ai_analysis`, so it was always `null` on `GET /sessions` — even though `openapi.Session` has always declared it, because list and detail share that schema. The embedded `review` object had the same divergence. A list UI could not render a risk badge without N+1 calls to the detail endpoint.

**Zero CLJS files change** in this PR: `webapp/src/webapp/app.cljs` is owned by #1665, so the CLJS `:sessions-panel` stays registered. Harmless — React Router's exact route wins.

## 📣 User-facing impact

The Sessions list is now a React page: filters round-trip through the URL so a filtered view can be bookmarked and shared, pagination works past 100 sessions, rows are keyboard-navigable and open in a new tab, and revoked/processing/executed reviews are finally visible instead of rendering an empty cell.

## 🔗 Related Issue

- EVL-145 — B2.1 Sessions list
- EVL-161 — T2.1 Expose `ai_analysis` (and the missing review fields) on `GET /sessions`

## 🚀 Type of Change

- [x] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

**Backend (EVL-161)** — `gateway/models/session.go`, SQL only, two hunks in the rows query of `ListSessions`:

- add `s.ai_analysis` and `s.origin` to the SELECT
- add `time_window`, `access_request_rule_name`, `min_approvals`, `force_approval_groups` to the review `jsonb_build_object`

No Go changes are needed: `Session.AIAnalysis` already carries `gorm:"serializer:json"`, `SessionReview`'s custom `Scan` unmarshals the whole blob and all four keys already have json tags, and `parser.go` routes list rows through the same `toOpenApiSession`/`topOpenApiReview` mappers as the detail path. Both callers — the HTTP handler and the MCP `sessions_list` tool — scan by column name, so the extra columns are additive (MCP's `sessionToMap` already emitted `ai_analysis` conditionally; it just never had a value).

Two deliberate decisions worth a reviewer's attention:

- **`s.origin` is included for literal alignment only.** It has no API surface — `openapi.Session` has no `origin` field — and is otherwise read only on the write path.
- **`s.blob_stream_id` is left alone.** It is list-only, has no matching struct field, and GORM discards it. Removing it is an unrelated cleanup that would muddy this diff.

**Frontend (EVL-145)** — `webapp_v2/src/pages/Sessions/` plus an extended `services/sessions.js` and a new `utils/datetime.js`. Ports `audit/views/{main,sessions_list,session_item,audit_filters,sessions_filtered_by_id}.cljs` and ~150 of the 786 LOC in `events/audit.cljs`. **`events/audit.cljs` is not deleted** — the other ~636 LOC is Wave 6.

**Docs** — `MIGRATION_ROADMAP.md` reordered to Wave 1 → 2 → 6 → 3 → 4 → 5 → 7 → Endgame (Wave 6 pulled ahead for EVL-165), plus `CONTEXT_MIGRATION.md` and `COMPONENTS.md`.

### Two deliberate deviations from v1

1. **Pagination is offset+append with a dedupe by session id**, not v1's `limit = count + 20` refetch. The gateway clamps `limit` at 100 (`session.go:46`) and computes `has_next_page` as `len(items) == limit`, so past 100 rows v1's button renders forever and adds nothing. `hasMore` is derived from `total` and additionally requires the last append to have added something — with `ORDER BY created_at DESC` and sessions arriving mid-paging, the offset window shifts and `total` grows.
2. ~~Rows are a real table with a header~~ — **reverted after seeing it rendered.** v1 was never a table: `main.cljs` wraps the rows in a plain `rounded-lg border` container and each row is a Radix `Grid columns="4"` with no header. The column labels were inventing chrome the original never had, so the list is back to `Paper` + `Grid`. The real anchor in the user cell stays — it is what restores keyboard access, and it renders as plain text.

### Bugs fixed while porting

All of these are current v1 behaviour:

| Bug | v1 |
|---|---|
| A failed fetch left the spinner running forever | `:audit->get-sessions` registered no `:on-failure` |
| REVOKED / PROCESSING / EXECUTED / UNKNOWN reviews rendered nothing — a revoked session was indistinguishable from an unreviewed one | `session_item.cljs:31-32` `:default` returned `nil` |
| A null `user_name` threw and took the whole list down | machine-identity sessions can have one |
| The date filter throws a `RangeError` **on Safari** | `new Date("YYYY-MM-DD 00:00:00.000Z")` is not a valid ISO string (`audit_filters.cljs:59-63`) |
| `/sessions/filtered?id=` fetched everything twice | both `app.cljs:543-551` and the component dispatched |
| A `batch_id` containing `&` broke the fetch and the Share link | string-concatenated, never encoded |
| "Load more" appended `#` to the URL and jumped scroll | `<a href="#">` with no `preventDefault` |
| The session list was unreachable without a mouse | rows were `div`s with no `tabIndex`/`role`/key handler |
| Enter in the ID search cost 2N requests | fired immediately *and* left the debounce running |
| ID-search results appeared in HTTP-race order | each arrival was prepended |
| A mistyped session id silently produced no row | errors were collected only to detect completion |

Rather than `role="link"` on the `<tr>` (which strips row semantics), the user cell holds a real anchor — keyboard focus, accessible name, and open-in-new-tab.

### Visual-fidelity pass after manual testing

Running the page side by side with the CLJS original turned up four divergences, all fixed in `747b2079`:

- **Table → list.** See above.
- **Badge colours.** v1 renders `[:> Badge {:color "green" :size "2"}]`, and Radix's default Badge variant is `soft`. I had mapped them onto the semantic `active`/`warning`/`danger` variants, which are *filled* — solid yellow especially is near-illegible. Now `color` + `variant="light"`.
- **Filter label weight.** Not font-weight: v1 uses `font-semibold` too. Mantine's `default` button variant ignores `color` and paints the label with `--mantine-color-text` (near-black) where the Radix trigger was `color="gray"`. Fixed with `c="dimmed"`. **Blast radius:** `ValueFilter`/`AsyncValueFilter` are shared, so the Rulepacks and Data Masking filters lighten too. That matches v1 everywhere, but it is a change outside Sessions.
- **Page flashed on every request.** `lookupByIds` and `fetchBatch` reset their slice to `EMPTY_*` on entering `loading`, emptying the list and handing the content area back to `PageLoader`. Both now keep the previous rows mounted, matching v1 (`main.cljs:57-64` only shows its spinner when the list is empty). The main list already preserved its items.

**Session details stays a full-page navigation, not the v1 modal.** A bridge to the CLJS modal is not available: on a React-only route the CLJS tree is parked hidden by `ClojureApp`, so the modal would open invisible — the same constraint documented in EVL-123. It returns in EVL-131, in the DES-27 layout.

### Other notes for reviewers

- **The React list calls `/sessions`; CLJS called `/plugins/audit/sessions`.** Same handler (`server.go:805` is an alias) — flagged so nobody diffing network tabs files a parity bug.
- **`usePaginatedConnections` yields `{ value: connection.id }`, but the gateway matches `?connection=` against the connection NAME.** The Resource Role filter remaps so the option value is the name; passing the id writes a UUID and silently returns zero rows. Commented at the call site.
- **`FILTER_KEYS` is an allow-list.** v1 forwarded every URL param to the API and relied on the gateway ignoring unknowns. `/sessions?batch_id=` therefore no longer filters — nothing links to it; `/sessions/filtered?batch_id=` is the supported URL and is unaffected.
- **The ID search is capped at 50 ids** (new). Each id costs one request, so an unbounded paste was a self-inflicted fan-out.
- **`getFilteredByIds` drops `?event_stream=base64`.** No list row renders anything from the event stream; v1 downloaded a full base64 transcript per session and discarded it.
- **`sessionsService.list()` still returns the raw axios response** while every new method unwraps to `.data`. `useConfigStatusStore` reads `data.total` off the response — unwrapping would not throw, it would silently pin the sidebar checklist to `false`. Commented in the file and in `COMPONENTS.md`.
- **`downloadFile`/`downloadInput` follow v1** in using `GET /sessions/:id` with blob params rather than the dedicated `/download` routes (`server.go:814-815`). Wave 6 should choose deliberately rather than inherit this.
- No analytics events were added, removed or moved by this PR.

### Test Configuration:
- **Browser(s)**: Chrome, Safari (the date-filter fix is Safari-specific)
- **OS**: macOS

### Tests performed:
- [x] Unit tests pass — `make test-oss` green
- [ ] Integration tests pass

```bash
go build ./gateway/... && go vet ./gateway/models/...
make test-oss

# The openapi tree must stay untouched — #1665 owns those files
make generate-openapi-docs
git status --porcelain -- gateway/api/openapi      # verified EMPTY

cd webapp_v2 && npm run lint && npm run build      # lint at the 15-error baseline, build clean
```

Proof the SELECT fix reaches the wire — every review key in the detail response must also appear in the list response:

```bash
diff <(curl -s -H "Authorization: Bearer $T" "$GW/sessions?limit=1" | jq -S '.data[0].review|keys') \
     <(curl -s -H "Authorization: Bearer $T" "$GW/sessions/$SID"    | jq -S '.review|keys')
```

### How to test

1. **Pagination** — scroll to "Load more sessions": rows append, no duplicates, no `#` in the URL, and the button disappears at the true end. Worth testing an org with exactly 20/40/60 sessions (the `has_next_page` bug) and one with >100 (where v1 dead-ends).
2. **Filters** — each of user / resource role / type / access request / date range / Jira: selecting adds the param and refetches **once** (watch the Network tab for loops); clearing **deletes** the key rather than writing `key=`. Confirm Resource Role puts a *name* in `?connection=`, not a UUID, and that rows actually narrow. Deep-link a fully-populated filter URL in a fresh tab; hit Back after three filter changes.
3. **Date range in Safari** — no `RangeError`, params are `T`-separated ISO.
4. **ID search** — three ids produce exactly three requests with no `event_stream` param, rows in typed order; Enter mid-debounce still fires N not 2N; a bogus id raises a snackbar; clearing restores the filtered list with no extra fetch.
5. **Row click** — lands on the CLJS `/sessions/:id` page fully loaded, no blank flash. Back restores the React list with filters, rows and loaded pages intact. Hard-reload `/sessions/:id` and confirm CLJS boots cold. Also try the CLJS page's own back/close to confirm the handoff to React `/sessions` (this seam can't be covered by CI — `:sessions-panel` is still registered on the CLJS side because `app.cljs` is frozen by #1665).
6. **`/sessions/filtered?batch_id=X`** — sticky header, infinite scroll with no duplicates, Share copies a working URL (test a batch id containing `&`), Share is **absent** when `disable_clipboard_copy_cut` is on, search filters without a request. `?id=a,b,c` fires exactly 3 requests with StrictMode on. `?batch_id=X&id=a,b` → batch wins.
7. **Keyboard** — Tab to a row, Enter navigates; Tab to a workflow chip, Enter opens the workflow and does *not* also fire the row.
8. **Badges** — an open `connect` session pulses Live; a REVOKED review shows a visible badge; a machine-identity session with a null `user_name` renders.
9. **Error state** — stop the gateway and reload `/sessions`: an error with a retry, not an eternal spinner.
10. **Regression** — as an admin, the sidebar Config Status "session ran" check still resolves (proves `sessionsService.list` stayed raw).

## ✅ Checklist

## 📄 Additional Notes

Reviewer focus, in order:

1. **The `s.origin` / `s.blob_stream_id` calls** in the backend commit — both are judgement calls, both explained above.
2. **The connection-name remap** in `SessionsFilterBar.jsx`. It fails silently if wrong: the chip shows a plausible name and the list just returns nothing.
3. **The roadmap reorder.** Zero code risk, real planning risk — four coupled locations plus ~25 `Wave N` mentions were re-read in the new order.

Deliberately out of scope: `pages/Settings/AuditLogs` keeps its own copy of the date formatter (`utils/datetime.js` is now the canonical one, noted in `COMPONENTS.md` for whenever that page is next touched).


sandromello:
✅ Build Completed with Success, Version=1668.0.0-g9a16f8c

PR #1687, opened 2026-08-06
DEP-91 (3/5): attach SBOM and bundled-tool manifest to releases

racerxdl:
## Stack (DEP-91)

Split from #1662 for reviewability, then **rescoped**: `main` has since landed #1681 (dependency/toolchain CVEs) and #1693 (`hoophq/hoopagent`), which superseded two PRs of the original stack.

- ~~#1683 — Go dependency + toolchain CVE bumps~~ — closed, superseded by #1681
- ~~#1685 — opt-in `hoophq/hoopdev-minimal`~~ — closed, superseded by #1693

- #1684 — 1. agent-tools: drop kernel-header CVEs, modernize clean train
- #1686 — 2. Vulnerability gates for hoopagent + govulncheck
- #1687 — 3. SBOM + bundled-tool manifest on releases **← you are here**
- #1688 — 4. Weekly base-OS rebuild of hoopagent
- #1689 — 5. Image release integrity hardening

Merge in order, bottom-up. Each PR targets the one below it.

The remaining work is what `main` still lacks: the fat image's kernel-header CVEs, and any vulnerability gate, SBOM, or scheduled OS patching for `hoophq/hoopagent`.

---

## What
Generates CycloneDX/SPDX SBOMs and a per-tag release manifest for the published images, recording each tag's index digest, per-platform child digests, and (for the fat agent) bundled CLI versions.

Covers `hoophq/hoop`, `hoophq/hoopdev`, and both `hoophq/hoopagent` flavours.

## Why
Makes shipped image contents auditable, and prevents reported tool versions from drifting from what actually shipped.

## Note on hoopagent tags
`hoophq/hoopagent` deliberately publishes **no unsuffixed tag and no `:latest`** — every deployment names its flavour. The image list therefore carries an explicit tag suffix per entry (`("hoophq/hoopagent", "-minimal")`) instead of assuming `repo:tag` resolves. Assuming a bare tag would have made the SBOM job silently skip both flavours.

`hoopagent` is intentionally **not** in `REQUIRED_IMAGE_REPOS`, so a lagging publish warns rather than failing the release.

## How to test
```bash
python3 -m unittest scripts/ci/test_generate_sbom_manifest.py
```
Expected: tests pass. Registry visibility probes stay non-raising; Trivy failures propagate.

Part of DEP-91.

Automated by MisterMal

PR #1686, opened 2026-08-06
DEP-91 (2/5): gate hoopagent vulnerabilities and add govulncheck

racerxdl:
## Stack (DEP-91)

Split from #1662 for reviewability, then **rescoped**: `main` has since landed #1681 (dependency/toolchain CVEs) and #1693 (`hoophq/hoopagent`), which superseded two PRs of the original stack.

- ~~#1683 — Go dependency + toolchain CVE bumps~~ — closed, superseded by #1681
- ~~#1685 — opt-in `hoophq/hoopdev-minimal`~~ — closed, superseded by #1693

- #1684 — 1. agent-tools: drop kernel-header CVEs, modernize clean train
- #1686 — 2. Vulnerability gates for hoopagent + govulncheck **← you are here**
- #1687 — 3. SBOM + bundled-tool manifest on releases
- #1688 — 4. Weekly base-OS rebuild of hoopagent
- #1689 — 5. Image release integrity hardening

Merge in order, bottom-up. Each PR targets the one below it.

The remaining work is what `main` still lacks: the fat image's kernel-header CVEs, and any vulnerability gate, SBOM, or scheduled OS patching for `hoophq/hoopagent`.

---

## What
Adds the vulnerability gating `hoophq/hoopagent` currently has none of:

- **Build smoke check** for `Dockerfile.agent`, both flavours, both platforms. It was previously only built after a tag push, so a broken Dockerfile could reach `main` and fail post-merge.
- **Enforcing Trivy gate** on fixable CRITICAL/HIGH for both flavours.
- **Unprivileged-user assertion** — resolves `USER` to a real uid, handling `distroless` (`10001:10001`, numeric) and `minimal` (`hoop`, by name) correctly.
- **Report-only Trivy** on `agent-tools`, which bundles upstream CLIs whose CVEs cannot be rebuilt away.
- **govulncheck** (pinned v1.6.0) on every PR for the shipped Go binary.

## Why
`hoophq/hoopagent` is the low-CVE image, but nothing enforced that. `main` runs Trivy with `--scanners license` only — there is no vulnerability gate anywhere, and no source-level Go gate.

## Risk
Adds blocking checks. The fat image stays report-only, so existing releases are not newly gated.

## Note on the placeholder binary
[...]

PR #1685, opened 2026-08-06
DEP-91 (3/7): add opt-in hoophq/hoopdev-minimal agent image

racerxdl:
## Stack (DEP-91)

Split from #1662 at Sando's request — that PR was too large to review.

- #1683 — 1. Go dependency + toolchain CVE bumps
- #1684 — 2. agent-tools: drop kernel-header CVEs, modernize clean train
- #1685 — 3. Opt-in `hoophq/hoopdev-minimal` agent image **← you are here**
- #1686 — 4. Trivy vulnerability gate + govulncheck
- #1687 — 5. SBOM + bundled-tool manifest on releases
- #1688 — 6. Weekly base-OS rebuild of the minimal image
- #1689 — 7. Image release integrity hardening

Merge in order, bottom-up. Each PR targets the one below it.

---

## What
Adds `hoophq/hoopdev-minimal`: the agent binary plus only the runtime it needs (tini, ca-certificates, locales). No database or cloud CLIs, no Node, no OCR, no `-dev`/header packages.

Also adds the `image.minimal: true` opt-in toggle to the agent Helm chart and `docs/custom-agent-image.md`.

## Why
Gives security-sensitive deployments a low-CVE base without forcing the bundled toolchain on anyone.

## Risk
**Additive and opt-in.** `hoophq/hoopdev` remains the default everywhere; an explicit `image.repository` always wins over `image.minimal`.

Works out of the box: native DB proxy (postgres/mysql/mssql/mongo/oracle), HTTP, TCP. Exec-based sessions and connections that shell out to a client need that binary added in a derived image — documented.

## How to test
```bash
helm template t ./deploy/helm-chart/chart/agent | grep 'image:'
helm template t ./deploy/helm-chart/chart/agent --set image.minimal=true | grep 'image:'
helm template t ./deploy/helm-chart/chart/agent --set image.minimal=true --set image.repository=myorg/custom | grep 'image:'
```
Expected: `hoophq/hoopdev`, then `hoophq/hoopdev-minimal`, then `myorg/custom` (explicit repository wins).

Part of DEP-91.

Automated by MisterMal

PR #1684, opened 2026-08-06
DEP-91 (1/5): agent-tools: drop kernel-header CVEs and modernize clean train

racerxdl:
## Stack (DEP-91)

Split from #1662 for reviewability, then **rescoped**: `main` has since landed #1681 (dependency/toolchain CVEs) and #1693 (`hoophq/hoopagent`), which superseded two PRs of the original stack.

- ~~#1683 — Go dependency + toolchain CVE bumps~~ — closed, superseded by #1681
- ~~#1685 — opt-in `hoophq/hoopdev-minimal`~~ — closed, superseded by #1693

- #1684 — 1. agent-tools: drop kernel-header CVEs, modernize clean train **← you are here**
- #1686 — 2. Vulnerability gates for hoopagent + govulncheck
- #1687 — 3. SBOM + bundled-tool manifest on releases
- #1688 — 4. Weekly base-OS rebuild of hoopagent
- #1689 — 5. Image release integrity hardening

Merge in order, bottom-up. Each PR targets the one below it.

The remaining work is what `main` still lacks: the fat image's kernel-header CVEs, and any vulnerability gate, SBOM, or scheduled OS patching for `hoophq/hoopagent`.

---

## What
Removes build-only packages from the `agent-tools` base image (the base of `hoophq/hoopdev`) and modernizes the clean train.

- Drops `alien`, `elfutils`, `libelf-dev`, `python3-dev` — they pulled in `linux-libc-dev` (kernel headers, ~891 findings) and `libc6-dev` with no runtime use.
- Installs `groff` and `python3-pip` with `--no-install-recommends`.
- Declares `xz-utils` explicitly (previously transitive via the removed packages).
- Uses runtime `unixodbc` instead of `-dev` headers.
- Makes `KUBECTL`/`SQLCMD`/`MONGOSH`/`MONGODB_TOOLS` versions ARG-driven so the legacy and clean trains can diverge.

## Why
Kernel headers were the single largest CVE contributor in the fat image and were never needed at runtime — every bundled CLI is prebuilt.

`hoophq/hoopagent` (#1693) does not help here: `hoophq/hoopdev` is a separate image that many deployments still need for exec-based connections and RDP, and `main` still installs all four packages.

## Risk
Legacy-train pins are unchanged, so `hoophq/hoopdev` keeps its current tool versions. The workflow asserts train contents (legacy keeps the legacy mongo shell; clean ships only `mongosh`).

## How to test
```bash
docker build -f Dockerfile.tools -t agent-tools-test .
[...]

PR #1688, opened 2026-08-06
DEP-91 (4/5): weekly base-OS rebuild of hoophq/hoopagent

racerxdl:
## Stack (DEP-91)

Split from #1662 for reviewability, then **rescoped**: `main` has since landed #1681 (dependency/toolchain CVEs) and #1693 (`hoophq/hoopagent`), which superseded two PRs of the original stack.

- ~~#1683 — Go dependency + toolchain CVE bumps~~ — closed, superseded by #1681
- ~~#1685 — opt-in `hoophq/hoopdev-minimal`~~ — closed, superseded by #1693

- #1684 — 1. agent-tools: drop kernel-header CVEs, modernize clean train
- #1686 — 2. Vulnerability gates for hoopagent + govulncheck
- #1687 — 3. SBOM + bundled-tool manifest on releases
- #1688 — 4. Weekly base-OS rebuild of hoopagent **← you are here**
- #1689 — 5. Image release integrity hardening

Merge in order, bottom-up. Each PR targets the one below it.

The remaining work is what `main` still lacks: the fat image's kernel-header CVEs, and any vulnerability gate, SBOM, or scheduled OS patching for `hoophq/hoopagent`.

---

## What
Weekly scheduled rebuild of `hoophq/hoopagent:-minimal` against current OS security patches, republished only when the OS layer is free of fixable CRITICAL/HIGH vulnerabilities.

Published tags:
- `:latest-minimal` — moving, always the latest patches
- `:<release>-minimal-<YYYYMMDD>` — weekly patched snapshot

The immutable `:<release>-minimal` tag is left untouched.

## Why
Releases already rebuild every image; this closes the gap **between** releases so the moving tags do not accrue unpatched OS CVEs for weeks.

## Design note — this is not a plain rebuild
`Dockerfile.agent` pins its base (`ARG UBUNTU_IMAGE=ubuntu:noble-<date>`) for reproducibility, and does **not** run `apt-get upgrade`. A plain rebuild would therefore reproduce identical bytes and patch nothing.

This job instead resolves the current `ubuntu:noble` digest and overrides the `UBUNTU_IMAGE` build-arg. The pinned default in `Dockerfile.agent` is untouched — bumping it stays a deliberate PR, per the policy in that file.

Only the `minimal` flavour is rebuilt: `distroless` installs no OS packages, so an OS re-patch has nothing to fix there.

## Risk
Scheduled workflow only. Gated on the same enforcing Trivy check per arch before any push, so a vulnerable rebuild is never published. Pin by digest when immutable bytes matter.

## How to test
Trigger **Slim Agent Weekly Rebuild** via `workflow_dispatch`. Expected: resolves the latest release, stages both arch tarballs, builds against current noble, passes both gates, then publishes.

Part of DEP-91.

Automated by MisterMal

### DIFF
--- a/.github/workflows/hoopagent-weekly-rebuild.yml
+++ b/.github/workflows/hoopagent-weekly-rebuild.yml
@@ -1,0 +1,191 @@
+# Weekly base-OS re-patch of the slim agent image (hoophq/hoopagent).
+# Closes DEP-91 point 4: "weekly base-OS patches".
+#
+# hoophq/hoopagent is the low-CVE agent image, but Dockerfile.agent pins its
+# base (ARG UBUNTU_IMAGE=ubuntu:noble-<date>) so builds stay reproducible.
+# A plain rebuild would therefore reproduce the same bytes and patch nothing.
+#
+# Instead this job rebuilds against the CURRENT noble base by overriding the
+# UBUNTU_IMAGE build-arg, and republishes only if the result is free of fixable
+# CRITICAL/HIGH vulnerabilities. The pinned default in Dockerfile.agent is left
+# untouched — bumping it stays a deliberate PR, per the policy in that file.
+# This fills the gap BETWEEN releases so the flavour tags do not accrue
+# unpatched OS CVEs for weeks.
+#
+# Only the `minimal` flavour is rebuilt. The `distroless` flavour installs no OS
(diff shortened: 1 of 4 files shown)


sandromello:
✅ Build Completed with Success, Version=1688.0.0-g862482d

PR #1698, opened 2026-08-10
[feat] Kerberos (Postgres, MSSQL) 

matheusfrancisco:
> [!IMPORTANT]

**Label: `minor`.** New protocol support and two new listener capabilities. Existing configs keep working unchanged.

## 📝 Description

Adds an MSSQL lane to hoopinspect and closes a silent inspection bypass on the Postgres lane.

The MSSQL codec decodes T-SQL from TDS, denies statements in a real `0xAA` ERROR token, and masks response values by re-framing TDS row tokens. Kerberos crosses the relay without a line of Kerberos code in the tree: TDS gives the SSPI exchange its own packet type, so the relay forwards `0x10` and `0x11` verbatim and starts inspecting at the first `0x01`/`0x03`.

On the Postgres side, a client holding a Kerberos ticket asks for GSS transport encryption first, because libpq defaults `gssencmode=prefer`. Accepted, the whole session becomes ciphertext and the gate reports zero statements, silent about it. The relay now refuses that request, so the client falls back and keeps its Kerberos authentication while the transport stays readable. The Postgres lane also records the authenticated principal, which every lane previously logged as `anonymous`.

Two bypass guards land with it. A routing ENVCHANGE (MSSQL) and a GSS-encryption request (Postgres) both end the relay's visibility without a trace. Both now produce `ErrStreamUnsafe`, which the gate turns into a denial that no rule enables and none can switch off.

A compose overlay under `deploy/docker-compose/envoy-stack/mssql/` runs the whole thing locally: SQL Server, a Samba domain controller, a Kerberos client, and Envoy terminating TLS on the three lanes.

## 📣 User-facing impact

hoop can now inspect, mask and audit Microsoft SQL Server traffic, including sessions where the user authenticates with Kerberos. On PostgreSQL, connections from Kerberos-enabled clients are no longer able to slip past inspection unnoticed, and audit records for that lane now name the user who ran each statement.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [x] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [x] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

**MSSQL codec** (`codec/mssql`, new)

- Decodes `SQLBatch` (0x01) and `RPCRequest` (0x03), reassembled across packets on EOM.
- Forwards `PRELOGIN` (0x12), `LOGIN7` (0x10) and SSPI continuations (0x11) verbatim, which is what carries a Kerberos ticket through untouched.
- `DetectSSPI` reports that a login is integrated. It cannot report who: the name lives inside the ticket.
- Masks responses by parsing `COLMETADATA` (0x81), `ROW` (0xD1) and `NBCROW` (0xD2), rewriting values, and laying fresh packets over the result. Covers the USHORT, BYTE, PLP and LONG value encodings.
- Refuses a routing ENVCHANGE, which would otherwise move the client onto a socket the relay does not hold.
- Restores `MSSQLError` in `proxy/deny.go`, ported from libhoop's version so a long operator message cannot overflow the token's length fields.

**Postgres lane** (`proxy/downstream.go`, new)

- Answers the pgwire pre-startup exchange as the server: refuses `GSSENCRequest` with `N`, and terminates TLS when `downstream_tls` is configured.
- Uses `N`, not `E`: pgjdbc reopens the TCP connection on `E`, which doubles each login in the audit trail.
- Recovers the `user` parameter from the StartupMessage and records it as the session principal.
- `codec/postgres` returns `ErrStreamUnsafe` if GSS-wrapped bytes arrive anyway, so the lane fails closed.

**Shared**

- `hoopinspect.ErrStreamUnsafe`, a root sentinel for bytes whose forwarding would end the relay's visibility. `gate` maps it to a denial regardless of policy.
- `session.AnonymousPrincipal`, and `store` now treats it as unset. Without that, a principal learned after `session_start` stops short of the session row.
- `downstream_tls` on a listener, refused at startup on any protocol other than postgres. The keypair loads at startup, so a bad path shows up in the log and not in a user's first login.

**Demo stack** (`deploy/docker-compose/envoy-stack/mssql/`)

- SQL Server 2022, Samba AD DC (KDC plus the LDAP directory SQL Server needs to resolve a login), a Debian client carrying msodbcsql18, and Kerberos wiring for both databases.
- Envoy terminates TLS on the three lanes. Postgres needs `envoy-contrib` for that, which is why this lives in an overlay and the parent stack keeps its plain `tcp_proxy` and its zero-extensions claim.
- `scripts/dev/mssql-stack.sh` and `scripts/dev/mssql-kerberos-check.sh`.

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: macOS 15 arm64 (SQL Server runs emulated; no arm64 image is published)

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

**Unit:** 69 tests across `codec/mssql` and `proxy`. Full `go test ./...` green, `go vet` clean, gofmt clean.

**Integration:** `hoopinspect/scripts/dev/mssql-kerberos-check.sh` reports **19 passed, 0 failed** from a `down -v` and a clean rebuild.

```
codec/mssql tests green
SELECT returned rows through the relay
DELETE denied, operator message reached sqlcmd
both rows still present, so the DELETE never reached the database
email redacted by the entity rule
ssn masked by the column rule, which detection alone would miss
NVARCHAR(MAX) masked too, so the PLP encoding is covered
no raw address reached the client
audit recorded masked_count=6 for that session
postgres over TLS, terminated by Envoy's postgres_proxy filter
without channel_binding=disable it fails, and says why
envoy reports sessions_terminated_ssl=16
the OS holds a TGT for alice@HOOP.TEST
the KDC issued a service ticket for MSSQLSvc/mssql.hoop.test:1433
hoop-inspect opened an mssqldb session for the Kerberos attempt
a routing ENVCHANGE is refused rather than forwarded
```

**Kerberos on Postgres, end to end:**

```
gss_authenticated | encrypted
        t         |     f
```

Authentication by ticket, transport left readable. The audit row for a denied statement reads `principal=alice@HOOP.TEST rule=no-destructive-sql`.

**Mutation-tested guards.** Reverting the routing-redirect guard fails the two tests that defend it, and no others. Reverting both masking fixes reproduces the original bug:

```
FAIL  email was not redacted
FAIL  a raw address reached the client despite masking being reported
PASS  audit recorded masked_count=6
```

That last line is why the masking check asserts on what the client received. A check that verified the count alone would have passed while the client read cleartext.

**Manual:** psql, sqlcmd (ODBC 18) and pgjdbc 42.7.4 across the four `gssEncMode` values, with packet captures on the relay's listeners confirming the GSS refusal and the SSPI passthrough.

n/a

## ✅ Checklist

## 📄 Additional Notes

**Where I would like review attention**

`codec/mssql/rewrite.go`. Four bugs in it survived eight green unit tests and only surfaced against a real SQL Server: the rewriter gave up on packet one (the server's first reply is PRELOGIN option data, not a token stream), deadlocked the login by mis-measuring `FEATUREEXTACK`, discarded a completed rewrite on the `SESSIONSTATE` token that closes most SQL Server 2016+ result sets, and reported `masked_count` for values the client received in cleartext. Each is fixed with a test, and the density of them suggests the file deserves a careful read.

**Known gaps, documented in the overlay README**

- MSSQL statement extraction covers `sp_executesql` only. `sp_prepare`, `sp_prepexec` and the cursor family yield no statement at all, and `EXEC('DELETE ...')` classifies as `call`, which I reproduced live by deleting a row through the guardrail. libhoop has an 894-line extractor covering these; porting it is the largest follow-up, and it raises a question worth settling: whether the two trees share one TDS parser or carry two.
- The MSSQL Kerberos login fails at SQL Server error 18452. The check script runs the same login directly at the database, with Envoy and hoop-inspect out of the path, and it fails identically there. Reported as `GAP`, and it fails the suite if the two paths ever disagree.
- The relay reaches SQL Server in plaintext. `upstream_tls` performs a TLS-on-connect handshake, which is TDS 8.0, and the Linux build of SQL Server has no `forcestrictencryption`. Closing this means teaching `starttls.go` the 0x12-wrapped TDS 7.x handshake that libhoop's `agent/mssql/net.go` already implements.
- pgjdbc's `prepareThreshold=5` hides repeat executions of a prepared statement from the audit trail. Measured: 10 executions, 5 records. Policy still fires at Parse.

**Two client parameters on the Postgres lane of the demo overlay**

`channel_binding=disable` and `gssencmode=disable`. Both come from Envoy owning that lane's TLS, both fail loudly when omitted, and the check script asserts the first one's failure mode so they do not read as cargo cult later.


sandromello:
✅ Build Completed with Success, Version=1698.0.0-gbc1152c

p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1698.0.0-g139210b

sandromello:
✅ Build Completed with Success, Version=1698.0.0-gbabb11b

sandromello:
✅ Build Completed with Success, Version=1698.0.0-gf2b2d0b

PR #1701, opened 2026-08-11
EVL-185: onboarding "How did you hear about Hoop?" survey

p3rotto:
## 📝 Description

Adds the onboarding **"How did you hear about Hoop?"** survey — a floating launcher on the right edge of the viewport that expands into a non-blocking card, offered once per user for 7 days after their user record is created.

Business driver: we are shipping free standalone tools as a Product Led Growth strategy, and need to measure how many OSS sign ups those tools originate versus the other channels.

**Where the data lives.** Both places, each for a different job:

- **Postgres** (`private.users.signup_origin`, `signup_origin_other`) is the source of truth for *who answered what* — joinable to email, org and subject, and durable regardless of the org's analytics mode.
- **Segment** gets `hoop-onboarding-origin-answered` with `org-id` and the chosen option, so the acquisition funnel is queryable in PostHog/Mixpanel alongside the rest of the funnel.

The free-text "other" answer is deliberately **not** in the event: it is user supplied and may contain personal data, which would bypass the organization's `analytics_mode` contract. Read it from the database when reviewing answers.

**The 7-day window needs no new state.** It is derived from the existing `users.created_at`, and evaluated *in Postgres* (`created_at > NOW() - make_interval(days => 7)`) rather than in Go — `created_at` is `TIMESTAMP WITHOUT TIME ZONE` written by `NOW()`, so comparing it against the gateway's own clock would drift by the database's UTC offset. Both sides of the comparison now come from the same clock.

## 📣 User-facing impact

New users are asked how they heard about Hoop during their first week, through a small floating button that expands into a card and never blocks the product. Answering takes one click and the question does not come back.

## 🔗 Related Issue

[EVL-185](https://linear.app/hoophq/issue/EVL-185/onboarding-include-origin-question)

## 🚀 Type of Change

## 📋 Changes Made

**Gateway**

- `gateway/migrations/000108_users_signup_origin.{up,down}.sql` — two nullable columns on `private.users`. `NULL` means "never answered", which is what drives whether the survey is still offered.
- `gateway/models/user_signup_origin.go` — `ShouldShowOriginSurvey` and `SetUserSignupOrigin`, the only read/write path for these columns. The write is a conditional `UPDATE ... WHERE signup_origin IS NULL`, so the first answer wins even when two tabs submit concurrently.
- `gateway/api/user/originsurvey.go` — `POST /users/self/signup-origin`, with an allowlist of the 8 options and validation of the free text. The length check counts characters (`utf8.RuneCountInString`), matching what `VARCHAR(255)` enforces — `len()` would have rejected a 255-character non-ASCII answer that stores fine.
- `gateway/api/user/user.go` — `GET /userinfo` now returns `show_origin_survey`.
- `gateway/analytics/events.go` — new `EventOnboardingOriginAnswered`.
- OpenAPI regenerated.

**Frontend (`webapp_v2`)**

- `features/OriginSurvey/` — the launcher and the card, mounted globally in `App.jsx` so it also reaches the chrome-less onboarding routes that render outside the app `Layout`. The launcher is the positioning anchor: vertically centred while collapsed, dropping to 75vh when open, with the card growing upward from just above it.
- `services/originSurvey.js` — the POST call.

### Test Configuration:
- **Browser(s)**: Chrome
- **OS**: macOS

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

**Automated**

- `make test-oss` → **1579 pass / 0 fail**, including `TestHoopMigrations`, which applies the full migration chain (now including `000108`) against embedded Postgres.
- 10 new DB-backed subtests in `gateway/models/user_signup_origin_test.go`, running against real Postgres via the pglite harness so the interval arithmetic and the conditional update are actually exercised. They cover: fresh user, inside window, outside window, `NULL` `created_at`, already answered, unknown user → `gorm.ErrRecordNotFound`, free-text storage, `NULL` free text for non-`other` answers, and first-answer-wins → `ErrAlreadyExists`.
  - Mutation-verified twice: setting the window to 700 days correctly fails `user-outside-window`, and dropping the `COALESCE` correctly fails `user-without-created-at` with `couldn't convert <nil> into type bool`.
  - The boundary is bracketed at 6.95 and 7.05 days rather than sitting on 7, so the assertions cannot flake on clock resolution.
- Migration rollback verified explicitly: `migrate down 1` drops both columns and re-running `up` restores them.
- `npm run lint` — no new problems (the 16 reported are pre-existing on `main`, in `vite.config.js` and `EventRouting`); `npm run build` succeeds.

**How to test manually**

1. `make run-dev-postgres && make run-dev`, then `cd webapp_v2 && npm run dev:full`.
2. Sign in as a user created less than 7 days ago. The launcher appears on the right edge, vertically centred, with a red pending dot — on any page, including the post-resource-creation and Terminal screens.
3. Click it. The launcher drops to 75vh and the card opens above it. Click it again (or the **X**) to collapse back.
4. Pick an option and press **Submit** → the form is replaced by "Answer received" straight away, without waiting for the gateway, and the whole widget retires ~3s later. Pressing the **X** during that window retires it immediately; clicking the launcher does not reopen the form.
5. Confirm it is recorded and cannot be overwritten:
   ```sql
   SELECT email, signup_origin, signup_origin_other FROM private.users WHERE subject = '<your-subject>';
   ```
   Reload — the launcher must not come back. `GET /api/userinfo` should now report `"show_origin_survey": false`.
6. Answer twice (e.g. replay the POST with curl) → the second call returns **409** and the stored answer is unchanged.
7. Pick **Other**, leave the text empty → Submit stays disabled. Type something → it is stored in `signup_origin_other`.
8. Shrink the window height to ~600px with the card open → the card scrolls internally instead of running off the top of the screen.
9. Ageing a user past the window hides it:
   ```sql
   UPDATE private.users SET created_at = NOW() - INTERVAL '8 days' WHERE subject = '<your-subject>';
   ```

## ✅ Checklist

## 📄 Additional Notes

**Reviewers, the design decision worth your attention:** the two new columns are deliberately **absent** from the `models.User` struct. `models.UpdateUser` persists via `DB.Save`, which writes every mapped column, and `gateway/api/signup` builds a partial `models.User{}` literal before calling it — so mapping these columns would let a future signup path silently wipe a stored answer. Keeping them off the struct makes that class of bug structurally impossible. A reviewable alternative is GORM's read-only field tag (`gorm:"->"`), which would make the columns discoverable on the struct while still excluding them from every write; happy to switch if reviewers prefer that trade-off.

**Analytics contract:** `hoop-onboarding-origin-answered` is a brand-new event constant. No existing `Event*` constant was renamed, removed, or orphaned, so no current metric or dashboard is affected.

**Feature flag:** not gated. Per the `CLAUDE.md` policy the decision is stated explicitly — nothing here is experimental, the survey is inert for users outside the 7-day window, and the write path is idempotent.

**Deliberate deviations:**
- The ticket says the survey persists for 7 days "after the user sees the onboarding screen"; this anchors the window to `users.created_at` instead. Matching the ticket literally would need a new `onboarding_seen_at` column written on a render, and the two anchors only diverge for a user who signs up and then does not open the app for days. The survey is also offered on every page for the whole window, not just on the onboarding screen, so reaching that screen never gates it.
- The design's button reads **"Submit"**; the ticket text said "send". Followed the design.
- The mock leaves clearance for the Intercom launcher bubble, but both webapps boot Intercom with `hide_default_launcher: true`, so that corner is empty in the real product.

**Submit is fire-and-forget.** The acknowledgement shows on the click without waiting for the response, so a slow or lost reply can never hold the UI. It is feedback for the click, not a claim that the write landed. Recovery is free and needs no client state: if the write does not land, the next `/userinfo` still reports `show_origin_survey` and the user is asked again. Because there is nothing for them to act on in the meantime, a failure raises no toast — it goes to the console only.

**Dismissal semantics:** the **X** collapses the card back to the launcher rather than hiding the survey — the launcher persists until the user answers or the 7 days elapse. This is what the mock shows (the launcher stays visible next to the open card) and it removes the earlier `sessionStorage` snooze key entirely, along with the shared-machine caveat that key carried.

### DIFF
--- a/gateway/analytics/events.go
+++ b/gateway/analytics/events.go
@@ -71,6 +71,9 @@ const (
 	EventUpdateOrgAnalyticsMode    = "hoop-update-org-analytics-mode"
 	EventProtectionProfileSelected = "hoop-protection-profile-selected"
 
+	// onboarding
+	EventOnboardingOriginAnswered = "hoop-onboarding-origin-answered"
+
 	// search api
 	EventSearch = "hoop-search"
 
--- a/gateway/api/openapi/autogen/docs.go
+++ b/gateway/api/openapi/autogen/docs.go
@@ -10255,6 +10255,67 @@ const docTemplate = `{
                 }
             }
         },
+        "/users/self/signup-origin": {
+            "post": {
+                "description": "Record how the authenticated user heard about Hoop. Each user may answer only once; a second attempt returns 409.",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "User Management"
+                ],
+                "summary": "Answer Signup Origin Survey",
+                "parameters": [
+                    {
+                        "description": "The request body resource",
+                        "name": "request",
+                        "in": "body",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/openapi.UserSignupOriginRequest"
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "Answer recorded"
+                    },
+                    "400": {
+                        "description": "Bad Request",
+                        "schema": {
+                            "$ref": "#/definitions/openapi.HTTPError"
+                        }
+                    },
+                    "403": {
+                        "description": "Forbidden",
+                        "schema": {
+                            "$ref": "#/definitions/openapi.HTTPError"
+                        }
+                    },
+                    "404": {
+                        "description": "Not Found",
+                        "schema": {
+                            "$ref": "#/definitions/openapi.HTTPError"
+                        }
+                    },
+                    "409": {
+                        "description": "Conflict",
+                        "schema": {
+                            "$ref": "#/definitions/openapi.HTTPError"
+                        }
+                    },
+                    "500": {
+                        "description": "Internal Server Error",
+                        "schema": {
+                            "$ref": "#/definitions/openapi.HTTPError"
+                        }
+                    }
+                }
+            }
+        },
         "/users/self/slack": {
             "patch": {
                 "description": "Patch own user's slack id",
@@ -18551,6 +18612,10 @@ const docTemplate = `{
                     "readOnly": true,
                     "example": "standard"
                 },
+                "show_origin_survey": {
+                    "description": "Whether the \"How did you hear about Hoop?\" survey should still be offered.\nTrue only while the user has not answered it and is within 7 days of their\nuser record being created. Always false for anonymous users.",
+                    "type": "boolean"
+                },
                 "slack_id": {
                     "description": "The identifier of slack to send messages to users",
                     "type": "string",
@@ -18598,6 +18663,34 @@ const docTemplate = `{
                 "slack_id": {
                     "type": "string",
                     "example": "U053ELZHB53"
+                }
+            }
+        },
+        "openapi.UserSignupOriginRequest": {
+            "type": "object",
+            "required": [
+                "origin"
+            ],
+            "properties": {
+                "origin": {
+                    "description": "The acquisition channel the user picked",
+                    "type": "string",
+                    "enum": [
+                        "search-engine",
+                        "ai-discovery",
+                        "referral",
+                        "already-in-use-at-company",
+                        "tech-community",
+                        "social-media",
+                        "hoop-free-tools",
+                        "other"
+                    ],
+                    "example": "ai-discovery"
+                },
+                "origin_other": {
+                    "description": "Free text detail. Required when origin is \"other\", ignored and stored as\nnull for every other option.",
+                    "type": "string",
+                    "example": "Saw it in a conference talk"
                 }
             }
         },
--- a/gateway/api/openapi/openapiv3.json
+++ b/gateway/api/openapi/openapiv3.json
@@ -8076,6 +8076,10 @@
             "readOnly": true,
             "type": "string"
           },
+          "show_origin_survey": {
+            "description": "Whether the \"How did you hear about Hoop?\" survey should still be offered.\nTrue only while the user has not answered it and is within 7 days of their\nuser record being created. Always false for anonymous users.",
+            "type": "boolean"
+          },
           "slack_id": {
             "description": "The identifier of slack to send messages to users",
             "example": "U053ELZHB53",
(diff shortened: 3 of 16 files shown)


sandromello:
✅ Build Completed with Success, Version=1701.0.0-gd758f82

sandromello:
✅ Build Completed with Success, Version=1701.0.0-gb517252

sandromello:
✅ Build Completed with Success, Version=1701.0.0-g283eb42

sandromello:
✅ Build Completed with Success, Version=1701.0.0-gfdf09eb

sandromello:
✅ Build Completed with Success, Version=1701.0.0-g64163e2

p3rotto:
Thanks — went through all four. Two were real and are fixed in bb69eb8; two are staying as they are, with reasoning below.

### 2. NULL `created_at` breaks the survey — fixed ✅

True positive, and it fails loudly rather than silently: `created_at` has a default but no `NOT NULL`, so a row holding `NULL` there made the whole predicate `NULL` and the scan blew up with

```
sql: Scan error on column index 0, name "?column?": sql/driver: couldn't convert <nil> (<nil>) into type bool
```

which `GetUserInfo` swallows into `show_origin_survey = false`. Wrapped the predicate in `COALESCE(..., false)` so an unknown creation time reads as *outside the window* instead of as a broken `/userinfo`. Added a subtest that seeds a user with `created_at NULL`; verified it fails with the exact error above when the `COALESCE` is removed.

### 4. `origin_other` counts bytes — fixed ✅

Also true. Confirmed against Postgres:

```sql
SELECT length(repeat('ы',255)), octet_length(repeat('ы',255)), repeat('ы',255)::varchar(255) IS NOT NULL;
--  255 | 510 | t
```

255 Cyrillic characters store fine in the `VARCHAR(255)` column but `len()` saw 510 bytes and returned a 400. Switched to `utf8.RuneCountInString`, which is what the column actually enforces. The error message already said "characters", so it is now true.

### 1. `console.warn` in OriginSurvey — keeping, precedent points the other way

The cited precedent is [#1642](https://github.com/hoophq/hoop/pull/1642), where this exact finding was raised on `useConfigStatusStore.js:140` — and it was **declined**, not accepted; the finding is struck through on that PR and it merged with the `console.warn` in place. So "team rejects console.warn in production paths" reads the history backwards: what the team rejected was the finding.

Substantively, same reasoning as there. This is not a debug leftover; it is the only diagnostic on a deliberately silent path. Submitting is fire-and-forget — the user gets their acknowledgement without waiting on the gateway, and a failed write self-heals because the next `/userinfo` still reports `show_origin_survey`. There is nothing for the user to act on, so a toast would be noise, and an empty `.catch(() => {})` would make a failing endpoint completely invisible. It matches 11 existing call sites in `webapp_v2`, the closest being `stores/useUserStore.js:145` (`console.warn('[analytics] identify failed:', err)`) — the same prefix-tagged shape for the same class of failure. Happy to move it if the project adopts a real client logger, but that is a codebase-wide change, not this PR.

### 3. 7-day window anchored to `created_at` — intentional, not a gap

This one was decided up front rather than by omission. The ticket's "after the user sees the onboarding screen" would require a new `onboarding_seen_at` column plus a write path on a render, and the two anchors only diverge for a user who signs up and then does not open the app for days — for whom the survey is far less useful anyway. `users.created_at` gives the same window with no new state and no new write path.

Worth noting the "less than 7 days" concern does not really land: the survey is offered on *every* page for the whole window, not only on the onboarding screen, so the window is not gated on reaching that screen at all. I have made the trade-off explicit under *Deliberate deviations* in the PR description so it is a visible decision rather than an accident. Happy to add the column if reviewers want the ticket read literally.

sandromello:
✅ Build Completed with Success, Version=1701.0.0-gbb69eb8

luanlorenzo (approved):

PR #1704, opened 2026-08-11
feat(hoopinspect): defer rule verdicts to OPA, rewrite the SQL scanner

matheusfrancisco:

## 📝 Description

Two changes to `hoopinspect`, plus the docs for both.

**A policy can now rule on what a producer found.** A PII scanner knows which
entity classes a statement carries and a model knows how risky it looks;
neither should also decide the outcome, because the outcome depends on the
actor, the hour and the table. Any rule can now carry `action: defer`, which
records a `policy.Finding` and hands the judgement to Rego. The AI analyzer
uses the same channel through `high: defer`. OPA reads the result as
`input.findings`, and `policy.opa.gate: true` adds a second call before the
producers run so Rego also decides what is worth spending a model call on.

**The SQL classifier is a scanner with a stack instead of a counter.** The old
one tracked parenthesis depth as a single integer, so it could tell "inside
parentheses" but not "inside a CTE body". That is why
`WITH d AS (DELETE FROM customers RETURNING *) SELECT count(*) FROM d` read as
a `select`. The new `lexer/` package makes one pass over typed tokens and keeps
a labelled region stack. `Statement` gained `Effects` and
`Relations{Name, Access}`, so "nothing writes to customers" is a rule you can
write, and `Operation` now reports the most consequential effect rather than
the leading verb.

Along the way the scanner grew a `Complete` flag. A statement it cannot follow
reports `Operation: unknown` with the reason in `metadata["sql.incomplete"]`
instead of guessing in the permissive direction.

## 📣 User-facing impact

Policy decisions that used to live in hoop-inspect's YAML can move into the
Rego an InfoSec team already owns: a rule matches locally and defers the
verdict, so "block a taxpayer id unless the caller is on the finance team" is
now expressible. Separately, several SQL statements that slipped past
destructive-write rules are now caught, including data-modifying CTEs, `MERGE`
delete branches and multi-statement queries hidden behind an escape string.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [x] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [x] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

**Policy**

- `policy.EvalContext` threads findings, a gate request and the session context
  between evaluators in one `Chain` run. `Chain` and `Rules` implement
  `ContextualEvaluator`; the `Evaluator` interface is unchanged.
- `action: defer` on every rule type. Matching stays in the local engine and
  only the determination moves. First-match-wins still applies to denials, so a
  deferring rule records and evaluation continues.
- `access: read|write` on `table` rules. Unset matches either, so deployed
  rules behave as before. Closes a false positive:
  `INSERT INTO staging SELECT * FROM customers` no longer trips a rule meaning
  "nothing writes customers".
- `policy.opa.gate: true` adds a gate-phase call. Both phases hit one URL and
  carry `input.phase`. An undefined gate decision allows, because a gate is an
  optimization over a policy someone already wrote.
- OPA input gained `phase`, `findings`, `effects` and `relations`. A lane with
  no producers sends a byte-identical document.

**Fixed: `input.context` arrived empty on every sidecar lane.** The gate
stamped session facts only onto a bare `*policy.OPAClient`, and `buildPolicy`
always returns a `Chain`, so the assertion never matched. Any Rego keyed on
`input.context.principal` read nothing. Covered by
`gate/sessioncontext_test.go`.

**Fixed: `action` on an `ai_analysis` rule was accepted and ignored.**
`splitAnalyzerRules` lifts those rules out before `policy.newRules`, so the
refusal there was unreachable from a config file. `validateAIRules` now refuses
it by name.

**SQL scanner**

- New `lexer/` package: dialect-parameterized one-pass scanner, typed tokens,
  labelled region stack. `Dialect` exists because `[` opens a quoted identifier
  in T-SQL and is an array subscript in PostgreSQL.
- `lexer.Split` replaces the codec's own statement splitter. That splitter was a
  second, weaker lexer: it cut `SELECT E'a\';b'; DELETE FROM customers` at the
  semicolon inside the literal and produced two fragments, neither of which
  parsed. The codec lost 80 lines.
- `Statement.Effects`, `Statement.Relations`, and `Operation` as the most
  consequential effect. `Tables` stays as the flattened names.
- New `lexer/conformance/` module: differential tests against PostgreSQL's own
  grammar through `wasilibs/go-pgquery` (pure Go under wazero, `CGO_ENABLED=0`).
  It is never a runtime dependency, so the root module still has zero requires
  and no `go.sum`.

**Docs**

- `docs/adr/hoopinspect-flow.md` rewritten for the new model, plus a section on
  the conformance module.
- `hoopinspect/README.md`, the envoy-stack README, `authz.rego` and
  `config.yaml` updated.
- Docs site: `config-file`, `risk-analysis`, `components` and `get-started`
  corrected, plus a new **Policy Rules** page covering every rule type,
  deferring to Rego and four paste-ready lane recipes.

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: darwin/arm64, go1.26.5, OPA 1.19.0, PostgreSQL 17.5 corpus

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

**Module suite:** 13 packages green, `go vet` clean, `go build ./client/...`
clean. About 2,900 lines of new tests across `lexer`, `policy`, `analyzer`,
`gate`, `sidecar` and a permanent bypass corpus at `hoopinspect/bypass_test.go`.

**Conformance against PostgreSQL's own parser** (`lexer/conformance`, run with
`CGO_ENABLED=0 go test ./...`):

| Measure | Result |
|---|---|
| Curated fixtures | 72/74 exact, 2 conceded (`DO`, `CALL`), **0 wrong** |
| PostgreSQL 17.5 regression suite | 20,391 statements, 20,240 complete (99.3%) |
| The 151 concessions | 94 `PREPARE`, 36 `CALL`, 18 `DO`, 3 unbalanced-paren, all failing closed |
| Fuzz | 4.2M execs, zero findings |

The assertion is asymmetric on purpose: the scanner must agree with PostgreSQL
**or** report `Complete=false`. It never claims equal precision. The oracle
found 11 real bugs in the scanner during review, every one of them failing
open; all 11 are fixed and pinned.

**End to end against a live OPA**, two producers and one policy:

```
DENY   cpf lookup     alice@example.com     BR_CPF may not appear in a query from this account
ALLOW  cpf lookup     treasury@example.com
DENY   mass delete    alice@example.com     blocked: rated high risk
ALLOW  mass delete    oncall@example.com    (break-glass in Rego's data)
ALLOW  select logs    alice@example.com     0 model calls, gate said no
DENY   drop table     oncall@example.com    local rule, OPA never consulted
DENY   mass delete    (model outage)        ai_status=error, Rego failed closed
```

**Bypasses closed**, measured through the real pgwire path:

| Statement | Before | After |
|---|---|---|
| `UPDATE audit SET n=E'O\'Brien'; DELETE FROM customers` | `update`, DELETE invisible | `delete`, both written |
| `WITH d AS (DELETE FROM customers RETURNING *) SELECT …` | `select` | `delete` |
| `WITH x AS (SELECT $$a)b$$) DELETE FROM customers` | `select` | `delete` |
| `MERGE INTO customers … WHEN MATCHED THEN DELETE` | `unknown` | `delete` |
| `COPY (DELETE … RETURNING *) TO STDOUT` | `other` | `delete` |
| `EXPLAIN ANALYZE DELETE FROM customers` | `other` | `delete` |

**False positives removed:**

| Statement | Before | After |
|---|---|---|
| `/* a /* b */ DELETE FROM customers */ SELECT 1` | `delete` | `select`, no relations (comments nest) |
| `EXPLAIN DELETE FROM customers` | `other` | `explain`, customers **read** (plans, does not execute) |
| `INSERT INTO staging SELECT * FROM customers` | both tables, no access | staging **write**, customers **read** |
| `CREATE FUNCTION f() … $$ DELETE FROM customers $$` | phantom write | no relations |

**Docs verified rather than reviewed:** 15 complete config examples across the
six hoop-inspect pages validate against `hoop-inspect -validate`; 3 Rego
policies pass `opa check --strict` and were evaluated for the break-glass,
outage and single-call-lane cases; Mintlify's `broken-links` reports one broken
link site-wide and it predates this branch.

**Reproduce:**

```bash
cd hoopinspect && go test ./... -count=1
cd lexer/conformance && CGO_ENABLED=0 go test ./... -count=1
cd deploy/docker-compose/envoy-stack && ./run.sh && ./demo.sh
```

n/a

## ✅ Checklist

## 📄 Additional Notes

**Breaking change, and it is the only one.** `CALL` and `EXECUTE` report
`unknown` instead of `call`. Their bodies live in the catalog or in an earlier
message, so no parser can say what they touch, and reporting a verb claimed
knowledge the relay does not have. **A rule written `operations: [call]` stops
matching.** Write `operations: [call, unknown]`. Documented in a warning on
`config-file.mdx`, `get-started.mdx` and the new Policy Rules page.

Everything else is additive. `ClassifySQL` keeps its signature, no exported
symbol was removed, and a consumer written against the previous API compiles
and runs unmodified; there is a check for that in the branch notes.

**Where reviewers should look:**

1. `lexer/analyze.go`, the region stack and `access()`. That function decides
   read versus write, and position is the only thing separating a target from a
   source in `DELETE FROM a USING b`.
2. `policy/policy.go`, `EvalContext` and `Finding`. `Reason` is deliberately not
   `omitempty`: an undefined reference makes a whole Rego rule undefined, so
   `sprintf(…, [f.reason])` inside a fail-closed rule deletes that rule when the
   key is absent. Two smoke rows caught it.
3. `sidecar/config.go`, `buildPolicy`. Chain order decides cost, and a lane that
   defers without a gate pays for the model before OPA sees the statement.

**Three cases stay unresolvable by any parser**, PostgreSQL's own included:
`DO $$…$$` interprets its body at runtime, `CALL` and `EXECUTE` keep theirs
elsewhere, and a function call inside a SELECT list can do whatever the function
does. The first two set `Complete=false`; the third does not, because flagging
every `SELECT count(*)` would leave the flag meaning nothing. So
`operations: [unknown, other]` is a permanent posture, and it is the reason a
model stays useful beside a scanner. MSSQL runs the scanner permanently, since
no credible Go T-SQL parser exists.

**Not built:** `require_review` is still refused at startup. Rego can express a
review decision today, so the blocker sits below the policy layer: holding a
pgwire connection open while a person reads the statement needs a review
backend, a notice channel so psql explains the wait, and cancellation the
`policy.Evaluator` interface cannot carry.

---


sandromello:
✅ Build Completed with Success, Version=1704.0.0-gba08a70

sandromello:
✅ Build Completed with Success, Version=1704.0.0-gad54417

p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1704.0.0-g2fde1eb

PR #1709, opened 2026-08-12
fix(onboarding): compute the setup checklist server side (DEP-136)

luanlorenzo:
## 📝 Description

The sidebar setup checklist derived its 9 booleans in the browser, firing 8 requests on every route change, on focus, and on two timers. One of them was `GET /sessions?limit=1`, which runs an unbounded `COUNT(*)` — minutes of work on a large deployment.

This moves the checklist to the gateway. `GET /orgs/onboarding` answers all 9 checks in a single query, all `EXISTS`, no counting. **8 requests per trigger → 1, and 0 once onboarding completes.**

It also fixes a real bug. The checks aren't monotonic — an agent going offline unticks "Deploy Hoop Agent", deleting a rule unticks "Explore AI Data Masking" — so progress ran backwards and a finished checklist could reappear months later. Progress is now recorded as milestones: each step latches the first time it's satisfied, and the checklist as a whole latches once all 9 are.

`GET /sessions` itself is untouched. Its `COUNT(*)` is tracked separately.

## 📣 User-facing impact

None, other than steps no longer unticking themselves.

## 🔗 Related Issue

DEP-136

## 🚀 Type of Change

- [x] ⚡ Performance improvement

## 📋 Changes Made

- Migration `000111`: `orgs.onboarding_steps` (jsonb, per-step latch) and `orgs.onboarding_completed_at` (whole-checklist latch)
- `GET /orgs/onboarding`, admin only — 9 checks plus the two connection-name shortcuts in one query, reported as live OR latched, stamping whatever is newly satisfied
- Setting a protection level now completes "Explore Guardrails" and "Explore AI Data Masking". Applying a profile materializes both rule kinds but stamps them `managed_by`, which the checks exclude — so an org on Maximum had 12 guardrails and full masking and was still asked to go explore both
- `show_setup_checklist` on `/userinfo` — server-owned visibility, same contract as `show_origin_survey`, read off the org row the handler already loads
- Webapp: `useConfigStatusStore` consumes the single endpoint. Removed the 8 probes, both timers, and the effects that ran while the widget was hidden. The store resets on logout and every read is scoped by `forUserId`
- "Assign people to groups" now links to `/organization/users` instead of sharing a destination with "Create more groups"

### Tests performed:
- [x] Unit tests pass — `make test-oss` green, including `org_onboarding_test.go` (one subtest per step, plus latch-survives-deletion and completion-is-terminal)

**How to test:** as an admin on a fresh org the widget shows 0/3. Tick a step, then delete whatever satisfied it — the step stays ticked. Complete every step and the widget disappears, `onboarding_completed_at` is stamped, and `/userinfo` returns `show_setup_checklist: false`. Stop the agent afterwards and the widget stays gone. In the network tab: one `GET /orgs/onboarding` per navigation, none after completion.

## 📄 Additional Notes

**Query plan.** The check subqueries filter on the bound `@org_id`, not the correlated `o.id`. That's deliberate — with `o.id` the planner can't see the value, assumes an early match and picks a `Seq Scan` on `private.sessions`, and an org with no sessions has nothing to stop that scan. Measured against 2M sessions:

| | correlated `o.id` | bound `@org_id` |
|---|---|---|
| org with 2M sessions | 0.2 ms | 0.2 ms |
| org with 0 sessions | 135 ms | 3.3 ms |

The zero-session case is the one that matters, since the only orgs calling this endpoint are the ones still onboarding. For reference, the `COUNT(*)` this replaces was 89 ms and 26,667 buffers on the same data.

Remaining cost is two `Seq Scan`s on `private.user_groups`, which has no `org_id` index — about 2.5 ms at 50k memberships. Not addressed here; an index on `(org_id, name)` would remove it.

**Other notes.**
- Latching happens on the read path, because the checks span features that have no reason to know a checklist exists. Steady state is zero writes, and a failed stamp retries on the next call.
- `main` is merged in. Its `000109_session_list_indexes` and `000110_connection_credentials_expired_sessions_index` took the numbers this branch was using, so the onboarding migration is now `000111`. Contents unchanged.
- `guardrails_explored` is true on a brand-new org because the gateway seeds a default guardrail rule with `managed_by IS NULL`. That matches the previous client-side behavior, but the step arrives pre-ticked — flagging in case that isn't wanted.


sandromello:
✅ Build Completed with Success, Version=1709.0.0-g067eda9

sandromello:
✅ Build Completed with Success, Version=1709.0.0-g664a588

sandromello:
✅ Build Completed with Success, Version=1709.0.0-g8f3d054

sandromello:
✅ Build Completed with Success, Version=1709.0.0-gaf71a1f

gateway/api/openapi/openapiv3.json
```diff
@@ -4334,6 +4334,74 @@
...
+          },
+          "completed": {
+            "description": "Whether onboarding is finished. Terminal: stays true even if an\nindividual check later regresses",
+            "type": "boolean"
+          },
+          "exec_connection_name": {
+            "description": "First web-terminal capable connection, for the \"Run your first session\"\nshortcut; null when the organization has none",
+            "example": "pgdemo",
+            "type": "string"
+          },
```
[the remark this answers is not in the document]
luanlorenzo: Not taking this one — it's accurate, but it's a repo-wide convention gap rather than something this PR introduced.

`gateway/api/openapi/types.go` has 56 `*string` fields and **zero** nullable annotations; the entire generated `openapiv3.json` contains one `"nullable"`. Existing nullable responses (`default_protection_profile`, `managed_by`, `deactivated_by`, `subtype`, …) are all emitted as plain `"type": "string"` today.

So annotating only `exec_connection_name` and `first_connection_name` would make them the sole exception in the spec, without fixing the class of problem for any client generator. Worth doing across the API surface in its own PR — happy to open a ticket if we want it — but not as a one-off here.

gateway/migrations/000111_orgs_onboarding_completed.down.sql
p3rotto, line 1: 109 and 110 are already taken. You should use 111+ instead.

p3rotto (changes requested):

sandromello:
✅ Build Completed with Success, Version=1709.0.0-g1d217b4

gateway/migrations/000111_orgs_onboarding_completed.down.sql
luanlorenzo: done

sandromello:
✅ Build Completed with Success, Version=1709.0.0-g8bf7a74

sandromello:
✅ Build Completed with Success, Version=1709.0.0-g4a165e3

p3rotto (approved):

EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1709.0.0-g327dec5

PR #1722, opened 2026-08-18
DEP-142: list tunnel connections with their fixed credentials

racerxdl:
## 📝 Description

`hsh tunnel ls` now lists each connection's fixed `noop`/`noop` credentials with a ready-to-paste client command.

Those credentials could not have worked before. The tunnel relayed every database as raw TCP (`pbagent.TCPConnectionWrite`), which lands on the agent's byte relay and copies bytes verbatim — so the client was handed the database's own auth challenge and `noop`/`noop` was rejected. Listing them would have been a false claim.

Each subtype now rides its own protocol packet family, reaching the agent's libhoop proxy that terminates client auth locally and re-authenticates upstream with the connection's stored secrets. `tcp` (opaque upstream) and `httpproxy` keep their existing families and report no credentials.

Side effect worth noting: the raw relay **refuses** guarded sessions (DEP-48), so guarded database connections were previously unusable over the tunnel. They now work, with guardrails and DLP enforced.

The companion CLI change is hoophq/hsh#33.

## 🎁 User-facing impact

`hsh tunnel ls` now shows the credentials and the exact command to connect to each resource, so users no longer need to look up credentials anywhere. Database connections with guardrails or data masking configured now work through the tunnel, with those rules enforced.

## 🔗 Related Issue

DEP-142

## 🔄 Type of Change

## 📋 Changes Made

- **`tunnel/client/pipe.go`** — route each subtype over its protocol packet family; re-frame the client stream into whole protocol packets (the agent's proxies decode one packet per write, so raw TCP chunks desync them); MySQL gets an init write because its proxy is server-speaks-first.
- **`tunnel/tunnelmgr/credentials.go`** (new) — the placeholder credentials and the per-subtype rule for which connections have them.
- **`common/mysqltypes`** (new) + `CopyBuffer` in `mongotypes`/`mssqltypes` — the framing the pipe needs, alongside the `pgtypes.CopyBuffer` that already existed.
- **`tunnel/client/connections.go`** — filter Oracle out unless `beta.oracle_native` is on; the agent refuses the session otherwise, so listing it would advertise a resource that can never connect.
- **`tunnel/ipc`** — `username`/`password` on `/v1/connections`, spec + openapi.yaml in lockstep.
- **`tunnel/README.md`** — documents the credentials and the tunnel's single-user trust model.

`Decode` in `mongotypes` and `mssqltypes` also gained a length-underflow guard. That is not drive-by hardening: the pipe feeds the user's raw TCP stream straight into those decoders, so a malformed length field is now reachable input where it previously wasn't. An undersized value underflowed the unsigned subtraction into a huge allocation or an indefinite blocking read.

### Test Configuration:
- **OS**: macOS (darwin/arm64), Docker for testcontainers
- Requires enterprise `libhoop` symlinked at `./libhoop` for the protocol-proxy tests; they skip cleanly on the OSS stub.

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

```bash
make libhoop-map                     # or symlink your enterprise libhoop checkout
go test ./common/... ./tunnel/... ./client/...
go test -tags integration -count=1 ./gateway/integration/transport/
```

The defect and the fix are both pinned by tests against a real gateway + agent + Postgres container:

| Test | What it proves |
|---|---|
| `TestTunnelRawTCPRejectsPlaceholderCredentials` | the old path forwards the backend's SASL challenge (type 10) to the client |
| `TestTunnelAcceptsFixedCredentials` | a real `lib/pq` client authenticates with `noop`/`noop` through `client.DialAndPipe` and runs two queries |
| `TestTunnelIgnoresClientCredentials` | bogus client credentials still connect, and `SELECT current_user` returns the **connection's stored user** — i.e. injection, not passthrough |
| `TestTunnelEnforcesGuardrails` | a denied query is blocked with the rule's own message |

Also run: full `gateway/integration/transport` suite, and `agent/integration` Postgres/MySQL/MongoDB/MSSQL against real database containers (the framing migration's main risk).

## 📌 Additional Notes

**Where I'd like review attention:**

1. **`pumpBytes` concurrency.** It now returns as soon as *either* direction finishes, instead of waiting for both. Protocol proxies don't echo a close packet, so the reader parks in `transport.Recv()` forever and the old `wg.Wait()` deadlocked — caught by the e2e test. `TCPConnectionClose` is sent exactly once via a `sync.Once` so it can't be skipped when the gateway side ends first.
2. **Scope.** The routing rewrite and shared framing are larger than "print two strings", but listing credentials that don't work isn't the ticket. Everything here is reachable from the tunnel; opportunistic cleanups that touched `client/proxy`, `hoop connect` and the agent's Oracle controller were dropped from this PR to keep the diff to the change under review.

**Known gap, deliberately not fixed here:** the daemon runs as root and authorises flows by destination address alone — it never identifies the calling OS user, so on a shared host any local account inherits the logged-in user's tunnel access. This predates the change (`tcp`/`httpproxy` were already credential-free over the same routes, and `hoop connect` serves `noop`/`noop` on loopback), but this PR widens it to databases. Documented in `tunnel/README.md` and tracked in DEP-144.

Automated by MisterMal

### DIFF
--- a/common/mongotypes/packet.go
+++ b/common/mongotypes/packet.go
@@ -49,6 +49,11 @@ func (p *Packet) Encode() []byte {
 
 func (p *Packet) Dump() { fmt.Println(hex.Dump(p.Encode())) }
 
+// MaxMessageSize bounds a single wire-protocol message. MongoDB's own limit
+// for a reply is 48 MB; anything beyond that means the stream desynchronised
+// and the length prefix is garbage, so we refuse rather than allocate it.
+const MaxMessageSize = 48 * 1024 * 1024
+
 func Decode(r io.Reader) (*Packet, error) {
 	var header [16]byte
 	_, err := io.ReadFull(r, header[:])
@@ -62,7 +67,13 @@ func Decode(r io.Reader) (*Packet, error) {
 		ResponseTo:    binary.LittleEndian.Uint32(header[8:12]),
 		OpCode:        binary.LittleEndian.Uint32(header[12:16]),
 	}
-	pktLen := int(p.MessageLength - 16)
+	// MessageLength counts the 16-byte header itself. Validate before
+	// subtracting: an undersized value would underflow the unsigned length
+	// into a multi-gigabyte allocation.
+	if p.MessageLength < 16 || p.MessageLength > MaxMessageSize {
+		return nil, fmt.Errorf("invalid mongodb message length: %d", p.MessageLength)
+	}
+	pktLen := int(p.MessageLength) - 16
 	frame := make([]byte, pktLen)
 	_, err = io.ReadFull(r, frame)
 	if err != nil {
@@ -70,6 +81,29 @@ func Decode(r io.Reader) (*Packet, error) {
 	}
 	p.Frame = frame
 	return &p, nil
+}
+
+// CopyBuffer re-frames the MongoDB stream from src onto dst, issuing exactly
+// one dst.Write per wire-protocol message.
+//
+// dst is a packet-stream writer whose Write boundaries become hoop packet
+// boundaries, so this cannot be an io.Copy: the agent-side proxy decodes one
+// message per write it receives, and arbitrary TCP chunking would desync it.
+//
+// It returns nil when src ends cleanly on a message boundary.
+func CopyBuffer(dst io.Writer, src io.Reader) error {
+	for {
+		pkt, err := Decode(src)
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
+		if _, err := dst.Write(pkt.Encode()); err != nil {
+			return err
+		}
+	}
 }
 
 // DecodeOpMsgToJSON with return json content separated by break line for
--- a/common/mongotypes/packet.go
+++ b/common/mongotypes/packet.go
@@ -100,8 +100,16 @@ func CopyBuffer(dst io.Writer, src io.Reader) error {
 			}
 			return err
 		}
-		if _, err := dst.Write(pkt.Encode()); err != nil {
+		encoded := pkt.Encode()
+		n, err := dst.Write(encoded)
+		if err != nil {
 			return err
+		}
+		// io.Writer permits a short write with a nil error. Forwarding a
+		// truncated message would desynchronise the peer's decoder, which
+		// frames on the length prefix we just cut in half.
+		if n != len(encoded) {
+			return io.ErrShortWrite
 		}
 	}
 }
--- a/common/mongotypes/packet_test.go
+++ b/common/mongotypes/packet_test.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/hex"
+	"errors"
+	"io"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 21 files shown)


sandromello:
✅ Build Completed with Success, Version=1722.0.0-gb5f8004

sandromello:
✅ Build Completed with Success, Version=1722.0.0-gc1de60f

sandromello:
✅ Build Completed with Success, Version=1722.0.0-g70db36e

sandromello:
✅ Build Completed with Success, Version=1722.0.0-g2724579

sandromello:
✅ Build Completed with Success, Version=1722.0.0-g70db36e

sandromello:
✅ Build Completed with Success, Version=1722.0.0-g9d0025c

racerxdl:
Thanks Qodo — pushed 9d0025c1.

**Half-close dropping responses (#4) was a real bug I introduced.** A clean EOF from the local side is a TCP half-close: the client shut its write direction after sending a request and is still waiting to read. Returning there closed the socket while the gateway's response was in flight, so any client that half-closes after its query lost the reply — `psql` and `mysql` do exactly this.

Confirmed with a failing test first (`TestRunPipe_HalfCloseStillDrainsResponse`): the existing `fakeLocal` can't express half-close because its `Read` blocks until `Close`, so the case was invisible. The reader now owns termination; a clean write finish only stops writing.

**Short writes (#1) fixed** in all three copiers. `io.Writer` may accept fewer bytes with a nil error, which would forward a truncated packet and desync the peer's decoder — one test per package.

**Ticket IDs removed from Go comments** entirely, rather than reprefixed. A comment should explain the reasoning without needing a ticket lookup; the ticket lives in the PR.

**Not taking #2** (`Logger.Printf` → `common/log`). `tunnelmgr` takes an injected `Logger` interface and uses it at 12 existing call sites; that seam is what keeps the manager testable without a global logger. Switching just this one line would be inconsistent with the package and break that. Happy to migrate the whole package separately if we want it on structured logging.

CI is green (33/33). `TestMySQL_ParallelSessions` failed once on the first run — unrelated flake, passes on rerun and 4/4 locally; that path touches no code in this PR.

racerxdl:
Right on the ordinals — my numbering was off, and my "12 call sites" was wrong too: it is **18** across `manager.go` and `handlers.go`. Thanks for the correction.

**Please dismiss finding 3** (`Logger.Printf` → `common/log`).

`tunnelmgr` is deliberately decoupled from any concrete logger: `Options.Logger` is a one-method interface (`Printf`) that every one of those 18 sites goes through, and `hsh-tunneld` injects a stdlib `log.Logger` writing to stderr with a `hsh-tunneld ` prefix (`cmd/hsh-tunneld/main.go:69`). That is the daemon's log format — it is a foreground/systemd process whose output is captured by `journalctl`, not the gateway's structured pipeline. Tests inject `log.New(io.Discard, ...)` to stay silent.

So routing one warning through `common/log` would emit a differently-formatted line into the middle of the daemon's log stream, bypass the injected sink that tests rely on, and leave 17 neighbours inconsistent. The rule is aimed at gateway runtime logging; this package predates and intentionally sits outside it.

If we do want the tunnel daemon on structured logging, that is a whole-package migration (`tunnelmgr` + `hsh-tunneld` + the `Logger` seam and its test injection) and belongs in its own change, not smuggled in behind a feature PR. Happy to open a ticket for it.

sandromello (approved):

PR #1726, opened 2026-08-19
ATR-225: Metabase + hoop-inspect stack, and the lexer fix it surfaced

p3rotto:
## 📝 Description

Two things, and the second one exists because of the first.

**A runnable Metabase + hoop-inspect stack** under `deploy/docker-compose/metabase-stack/`. It demonstrates that `hoopinspect` enforces masking, policy and a per-statement audit trail on the wire, so the control sits on the warehouse rather than inside the BI tool. No Metabase plugin, no driver, no patched image, no agent in the container: the integration is a hostname and a port in Metabase's connection form.

```
metabase :3000 -> hoop-inspect :15432 -> appdb :5432
client ------------------------------> appdb :5432   (unmasked ground truth)
```

**A lexer fix, found by running it.** PostgreSQL allows a reserved word as a column or table alias exactly when `AS` is written, so `SELECT 1 AS delete` is a legal select. The analyzer treated every keyword in `headAfter` as unconditional, so that `delete` became a statement head and the select was classified as a delete. Metabase's schema sync asks the catalog which privileges it holds and names each column after the privilege it tested:

```sql
select ... has_table_privilege(..., 'delete') as delete from pg_tables
```

A read-only lane refused that on every table, so the sync failed and Metabase reported an empty database. Any BI tool that introspects privileges writes some version of it. `AS` is now conditional: `analyzer.headFollows` takes a statement after `AS` only under a DDL head, which keeps `CREATE TABLE x AS SELECT` and `PREPARE p AS SELECT` working. CTE bodies are unaffected, because head position inside one comes from the parenthesis push labelled `regCTE`, not from the `AS` in front of it.

## 📣 User-facing impact

A read-only `hoopinspect` policy no longer refuses catalog queries that name a column after a SQL verb, which unblocks schema introspection from BI tools such as Metabase. The repo also gains a runnable Metabase demo of wire-level masking, policy and audit.

## 🔗 Related Issue

ATR-225 (parent ATR-215).

## 🚀 Type of Change

- [x] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [x] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

- `hoopinspect/lexer/keywords.go`: `headAfter["as"]` is no longer unconditional, and the comment says why.
- `hoopinspect/lexer/analyze.go`: new `analyzer.headFollows`, which resolves `AS` against the enclosing statement's verb.
- `hoopinspect/lexer/lexer_test.go`: `TestReservedWordAliasIsNotAStatementHead` covers the false positive; `TestASStillHeadsAStatementUnderDDL` covers CTAS, `CREATE VIEW`, `CREATE MATERIALIZED VIEW`, `PREPARE` and a data-modifying CTE, so the fix cannot be over-applied.
- `hoopinspect/bypass_test.go`: `TestPrivilegeIntrospectionIsNotAWrite` runs Metabase's real catalog statement through the wire path and asserts a read-only rule allows it. It sits beside the bypasses because the consequence is a policy verdict.
- `hoopinspect/lexer/conformance/oracle_test.go`: four statements added to the `regression` corpus, so PostgreSQL's own parser judges the fix rather than only the unit tests. The suite reports 76/78 exact, 2 conceded, 0 wrong; both concedes predate this branch.
- `deploy/docker-compose/metabase-stack/`: new stack. `run.sh` (cert, build, up, provision), `demo.sh` (asserting walkthrough), `docker-compose.yml`, `hoopinspect/config.yaml`, `hoopinspect/read-audit.py`, `metabase/provision.py`, `metabase/query.py`, `upstream/seed.sql`, `README.md`.
- Policy in that stack is scoped **per table**, not lane-wide, because the config is meant to be copied and edited: `payroll` is unreachable in either direction, `customers` refuses writes with the owning system named, and the `operation` rule sits last as the backstop for the table nobody listed. `seed.sql` carries a third table so the scoping is demonstrated rather than described, and `demo.sh` asserts the refusal from the SQL editor, through a join where `payroll` is not the first table named, and from the query builder.
- `hoopinspect/CLAUDE.md`: new. Records the zero-dependency invariant and how to check it, the per-nested-module test loop, and that a `lexer/` change is unverified until the conformance suite passes. Written because this PR nearly shipped without running it.
- `CLAUDE.md`: the root file claimed six modules under `go.work` and never mentioned `hoopinspect`. There are thirteen entries, seven of them under `hoopinspect/`. Corrected, with a pointer to the new file.

The sidecar image is built from the local `hoopinspect` tree with the same Dockerfile `../envoy-stack` uses, so the two stacks share one image and cannot drift.

### Test Configuration:
- **Browser(s)**: n/a for the assertions; Metabase UI checked manually in Chrome
- **OS**: macOS 15.5 (darwin/arm64), Docker Desktop, Metabase `v0.63.13`, Postgres 17.6

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass

```
cd hoopinspect && go test -count=1 ./...     # all packages ok
go vet ./...                                 # clean

# separate module, not reached by the line above
cd lexer/conformance && CGO_ENABLED=0 go test -count=1 ./...
# agreement: 76/78 exact (97.4%), 2 conceded, 0 wrong
```

The conformance suite runs PostgreSQL's real parser against the scanner and fails on a confident disagreement. It is the check that matters for a `lexer/` change, and it needs its own invocation because it is a nested module.

Nothing outside `hoopinspect/` changed in Go, so `make test-oss` is unaffected by this PR.

To reproduce the stack:

```
cd deploy/docker-compose/metabase-stack
./run.sh          # about 2 minutes on a cold cache, Metabase migrates H2 on first boot
./demo.sh         # exits nonzero if any assertion fails
./run.sh down
```

`./demo.sh` issues every query **through Metabase's own API**, not through `psql`, because the claim under test is that redaction survives Metabase's JDBC driver, result pipeline and download endpoint. Last run on this branch, exit 0:

```
PASS 5000 rows exported
PASS 5000 rows redacted, masking did not stop at row 1000
PASS 0 addresses survived the export
PASS the customers rule answered, not the backstop
PASS the backstop answered for the table nobody listed
PASS CREATE is refused, with alter, grant, revoke and call
PASS customers still has 3 rows
PASS events still has 5000 rows
PASS no zzz_probe table exists, so the statement never landed
PASS payroll refused with the message its own rule carries
PASS the join is refused too, because the rule reads every relation
PASS the query builder path is refused as well

99 statements recorded (41 set, 37 select, 10 show, 6 begin, 2 rollback, 2 delete)
92 allowed, 7 denied, 38 result sets
10034 values masked across 23 result sets (EMAIL_ADDRESS, BR_CPF, IBAN_CODE, column:ssn, US_SSN)
verified: none of the 13 seeded values appears in the audit trail
```

Those counts were measured twice from a clean `./run.sh` and were identical both times.

Also verified: `make test-hoopinspect` (16 packages, exit 0), `shellcheck run.sh demo.sh`, `bash -n`, `python3 -m py_compile` on the three scripts, `docker compose config -q`, and `go run . -validate` against `hoopinspect/config.yaml`.

## ✅ Checklist

## 📄 Additional Notes

**Where to focus review.**

1. `analyzer.headFollows` in `hoopinspect/lexer/analyze.go`. The fix narrows one entry in a keyword table, and the risk is narrowing it too far. `TestASStillHeadsAStatementUnderDDL` is the guard; if there is a form of `AS` that heads a statement under a non-DDL verb, that test is where it belongs.
2. `headAfter["as"] = false` reads oddly as a map entry. It is deliberate: keeping the key documents that `AS` was considered and is handled elsewhere, rather than looking like an omission. Say so if you would rather it were deleted from the map.

**Known gaps, stated in the README rather than left to be discovered.** An aliased column escapes a column rule; a reshaped value escapes an entity rule; bind parameters (`'B'`) are not decoded, so request-side rules never see a query-builder filter value. Response masking is unaffected by the last one. **Masking is also per lane while policy is per table**: `alcatraz.Rule` has no `Tables` field, so a mask rule applies to every result set the lane carries. The table OID is on the wire in `RowDescription` and the codec already reads past it, so closing that gap is a `pg_class` lookup plus a cache rather than a redesign, and it is not in this PR. The README also states plainly where Metabase Pro is the better answer, and quotes Metabase's own documentation for the cases where it is not.

**Label.** `skip-release`. Nothing outside `hoopinspect/` imports `hoopinspect`, and nothing builds or publishes it into the released binaries, so a version bump here would tag a gateway, agent and client release whose binaries contain none of this change. The new test target below does not change that: it runs the tests, it does not ship the code.

**`hoopinspect/` now runs in CI, and did not before this PR.** `make test-oss` runs `go test github.com/hoophq/hoop/...`, which does not match module `github.com/hoophq/hoopinspect`, and no workflow referenced it, so the tests added here would have passed locally and never run in the pipeline, including on future PRs that touch the lexer. This PR adds `make test-hoopinspect`, which walks every `go.mod` under `hoopinspect/` and runs each module's tests, and makes `test-oss` depend on it. The list is discovered rather than written down, so a nested module added later is covered the day it is added. Cost measured on this machine: 16 packages, under 3 seconds, no services and no credentials. `hoopinspect/CLAUDE.md` and the root `CLAUDE.md` were updated to match.

**No analytics events were added, renamed or removed.**


sandromello:
✅ Build Completed with Success, Version=1726.0.0-g4ac2b94

sandromello:
✅ Build Completed with Success, Version=1726.0.0-g1ba7924

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
...
+Full documentation is `README.md` here (68 KB, read the section you need, not
+the whole thing).
+
+## The invariant: zero dependencies
+
+`github.com/hoophq/hoopinspect` depends on the standard library and nothing
+else. Test dependencies included. The reason is in `go.mod`: the library
+exists to be vendored without a supply-chain review, read end to end in an
+afternoon, and compiled to `wasip1` for an Envoy filter. A dependency is a
+breaking change to the product pitch, not a convenience.
```
matheusfrancisco: I would not add this expression about the Envoy filter in the CLAUDE.md

matheusfrancisco: It can become an Envoy filter with FFI, but it is not the reason I created the go.mod  :)

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
...
+Two checks, both cheap:
+
+```bash
+GOWORK=off go list -m all        # must print exactly one line: the module itself
+ls go.sum                        # must not exist
+GOOS=wasip1 GOARCH=wasm go build ./...
+```
+
+**Anything that needs a dependency gets its own `go.mod`.** That is why these
+exist as nested modules rather than packages:
```
matheusfrancisco: It is not necessary to be anything other than a nested module dependency "go.mod"; the major nested go. mod was part of the software, I think, and does not depend on the implementation. By this I mean we can remove SQLite or add a new nested and plug. 
But i do not think this was the best decision 🤔

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
...
+for m in cmd config/yaml pii/alcatraz store/sqlite analyzer/vertex lexer/conformance; do
+  (cd "$m" && CGO_ENABLED=0 go test ./...)
+done
+
+# the oracle: run this for ANY change under lexer/
+(cd lexer/conformance && CGO_ENABLED=0 go test ./...)
+
+# validate a sidecar config without starting anything
+(cd cmd && go run . -validate -config /path/to/config.yaml)
+```
```
matheusfrancisco: Does this mean the hoop inspect test isn't running in our CI?

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
...
+## Gotchas
+
+- **`go test ./...` at the root reaches no nested module.** Separate `go.mod`
+  files are what keeps the root at zero dependencies, and the cost is that
+  every nested module needs its own invocation. See the loop above.
+
+- **CI never runs any of this.** `make test-oss` runs
+  `go test github.com/hoophq/hoop/...`, which does not match module
+  `github.com/hoophq/hoopinspect`, and no workflow in `.github/workflows/`
+  references it. Tests here pass or fail on your machine only. Run them
```
matheusfrancisco: https://github.com/hoophq/hoop/pull/1726/changes#r3814489731 answered..

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
...
+- A masking or policy change needs a test that would fail without it. `gate/`
+  and `bypass_test.go` are where behaviour that an operator sees is pinned.
+- `bypass_test.go` holds statements that evade classification. A new evasion
+  belongs there, phrased as the verdict an operator would get.
+
+## Running it
+
+Two compose stacks exercise the relay end to end, both outside this directory:
+
+- `deploy/docker-compose/envoy-stack/` : Envoy plus OPA plus the sidecar.
```
matheusfrancisco: I would add that there is more inside envoy-stack, but we need to split this to run without envoy them..

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
...
+| `codec/{postgres,mssql,http}` | wire decode and response rewrite, one package per protocol |
+| `policy/` | statement to verdict; local rules, then OPA |
+| `analyzer/` | the model-backed evaluator, third in the policy chain |
+| `gate/` | orders inspection, policy, audit and masking into one decision |
+| `proxy/` | TCP relay that pumps both directions through a Gate |
+| `sidecar/` | assembles the relay from config: sinks, evaluators, listeners |
+| `session/` | one inspected connection and the identity behind it |
+| `audit/` | the write side of the trail |
+| `store/` | the read side |
+| `pii/` | detectors and maskers |
```
matheusfrancisco, line 123: hehehe this made me realize the folders organization are really bad :(

hoopinspect/CLAUDE.md
```diff
@@ -0,0 +1,143 @@
+# CLAUDE.md: hoopinspect
```
matheusfrancisco, line 1: I would change this for AGENTS.md I don't use claude so might be better for all type of agents... but also idk if is true hehe

matheusfrancisco (commented):
Nice PR so basically
Metabase is using Postgres and hoopinspect in the middle.

But I have a quick question: if I am using Metabase with my current database setup 
Metabase -> Pg

and I would like to add the hoopinspect, would I have to change the connection on metabase?

sandromello:
✅ Build Completed with Success, Version=1726.0.0-g6290dbc

deploy/docker-compose/metabase-stack/run.sh
```diff
@@ -0,0 +1,111 @@
...
+# refuses to start unless it is 0600.
+c_step "TLS certificate for appdb"
+if [[ ! -f upstream/certs/server.crt ]]; then
+    mkdir -p upstream/certs
+    openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
+        -keyout upstream/certs/server.key -out upstream/certs/server.crt \
+        -subj "/CN=appdb" \
+        -addext "subjectAltName=DNS:appdb" \
+        2>/dev/null
+    chmod 600 upstream/certs/server.key
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** in `1bbb7f84`.

Confirmed real. PostgreSQL requires the key be 0600 and owned by the database user or by root; a key written by the host user is neither, and `chmod 600` alone does not satisfy it. Docker Desktop on macOS hides this by remapping bind-mount ownership, plain Linux does not, and there `appdb` never becomes healthy.

Rather than `chown` on the host, which would need a uid hardcoded in this file, `run.sh` now mints the pair inside the postgres image that reads it, taking the uid from that image. The tag is read out of `docker compose config --format json` so it cannot drift from the compose file. Dropping the host `openssl` prerequisite is a side benefit.

Verified from a clean `./run.sh`:

```
key owner=postgres group=postgres mode=600
crt owner=root group=root mode=644
```

`./run.sh down` still removes the root-owned files, because the directory is created on the host first.

deploy/docker-compose/metabase-stack/README.md
```diff
@@ -0,0 +1,392 @@
...
+  by both. Enumerate the columns you know about *and* run the detector, and
+  accept that an analyst with a native SQL editor can reshape a value out of
+  both. If that matters more than the editor does, deny the operation instead
+  of masking the response: policy runs on the request, where there is nothing
+  to reshape.
+
+- **Bind parameters are invisible to request-side rules.** The postgres codec
+  reads `'Q'` Query and `'P'` Parse; it does not decode `'B'` Bind. A query
+  builder filter arrives as `WHERE "customers"."ssn" = $1` with the value in
+  a later message, so a rule that denies on statement content never sees it.
```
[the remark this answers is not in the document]
p3rotto: ⏭️ **Deferred**.

Accurate, and already stated in the README's known gaps rather than left to be discovered: "bind parameters escape request-side rules". Two reasons it is not fixed here.

Decoding `Bind` means correlating a parameter list with the `Parse` that named the statement, inside `codec/postgres`. That is a change to the library's wire handling, with its own tests and its own failure modes, and it does not belong in a PR whose subject is a compose stack.

Response masking is unaffected. Values still come back masked, because `Codec.Rewrite` runs on the response and never sees the request. The exposure is scoped to request-side policy evaluation of a query-builder filter value, which is what the README says it is.

Worth its own ticket. Say so if you would rather this PR block on it.

deploy/docker-compose/metabase-stack/hoopinspect/config.yaml
```diff
@@ -0,0 +1,106 @@
...
+      enforce: true
+      rules:
+        # BI is a read. Unlike a GRANT, this refusal carries a message the
+        # analyst can read. type: operation is lexer-derived, so it keys on
+        # the effect and not the leading verb: a DELETE inside a CTE is
+        # refused, and SELECT 'DROP TABLE customers' is not.
+        - name: read-only-lane
+          type: operation
+          operations: [insert, update, delete, drop, truncate]
+          message: this Metabase connection is read-only; writes are not permitted
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** in `fa908ee6`.

Confirmed by running `CREATE TABLE zzz_probe (id int)` through Metabase's API against the lane as it stood: it succeeded. `appuser` owns this database, so nothing underneath was going to stop it.

`read-only-lane` now names `create`, `drop`, `alter`, `grant`, `revoke` and `call` alongside the four write verbs, since operation matching is exact. `./demo.sh` gained two assertions: the `CREATE` is refused, and `pg_tables` is asked afterwards to prove the statement never landed.

I did not take the second focus area, a case in `hoopinspect/bypass_test.go`. That file pins library behaviour; asserting a config file under `deploy/` from it would couple the library's test suite to a demo stack. The assertion lives in `demo.sh`, which is the thing that owns the config.

deploy/docker-compose/metabase-stack/README.md
```diff
@@ -0,0 +1,392 @@
...
+  Inspection is unaffected either way, because the sidecar is the TLS client
+  on that hop and decrypts what it reads.
+
+- **One shared credential.** See [Identity](#identity). The per-human signal
+  is Metabase's own comment, which is attribution rather than authentication.
+
+- **Postgres only, in this stack.** hoopinspect ships postgres, http and
+  mssql codecs. Metabase's other warehouses (BigQuery, Snowflake, Redshift,
+  MySQL) need a codec for their wire protocol; that is a new `codec/<name>`
+  package and nothing else, but it does not exist today.
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** in `d5e7e248`.

The factual error was real: Redshift speaks the PostgreSQL wire protocol on 5439, so the existing pgwire codec is the starting point, not a new codec. The known-gaps bullet now splits three cases instead of one:

- **Redshift**: pgwire, existing codec is the starting point, untested here. The risk is the Amazon JDBC driver and its lz4 negotiation, not framing.
- **MySQL / MariaDB**: a genuine codec gap, `codec/mysql`.
- **Snowflake / BigQuery**: HTTPS clients against a vendor REST API, so the problem is reading JSON payloads rather than framing a socket.

I did not add the "MySQL-first expansion plan" the prompt asks for. A roadmap commitment in a README is not mine to make in a demo PR, and an invented one is worse than none.

hoopinspect/CLAUDE.md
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** in `4a93ea2e`.

`make test-hoopinspect` walks every `go.mod` under `hoopinspect/` and runs that module's tests, and `test-oss` now depends on it, so CI runs them. The list is discovered rather than written down, so a nested module added later is covered the day it is added.

Verified: 16 packages, exit 0, under 3 seconds. Every module there is pure Go with no services and no credentials, so this adds nothing to the pipeline's requirements.

`hoopinspect/CLAUDE.md` and the root `CLAUDE.md` were updated to match, and the PR body's "not in CI" paragraph was rewritten, since it is no longer true.

deploy/docker-compose/metabase-stack/hoopinspect/read-audit.py
```diff
@@ -0,0 +1,116 @@
...
+counts them and shows only the interesting rows.
+"""
+
+import json
+import sys
+from collections import Counter
+
+# Values that exist in upstream/seed.sql and must never appear here. The audit
+# trail recording, in the clear, a value that masking removed would un-mask it.
+FORBIDDEN = ("ada@example.com", "user00001@example.com", "555-12-3456")
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** in `ff67a0c2`.

I did not extend the list. A hardcoded copy of another file's values is what went stale in the first place, and extending it to 12 leaves the same trap for whoever adds the next row.

`read-audit.py` now reads the values out of `upstream/seed.sql`: it parses each `INSERT ... VALUES` tuple, aligns the fields to the column list, and keeps the values sitting in the columns `config.yaml` masks. That finds the 12 the seed carries: 3 emails, 3 SSNs, 3 CPFs, 3 IBANs, with `payroll.bank_iban` reusing the customers IBANs. `events` is filled by a `SELECT` rather than literal tuples, so one address from the generated series stays hardcoded with a comment saying why. Thirteen in total.

It now fails loudly if the seed cannot be read, if a row misparses, or if the set comes back empty. A leak check that silently covers nothing is the same bug wearing a different hat.

The success line names what it checked: `verified: none of the 13 seeded values appears in the audit trail`. Each of the 13 was injected into a synthetic audit event on its own, and every one is caught.

p3rotto:
## Qodo Fix Summary - Round 1

5 fixed, 1 deferred.

| # | Issue | Decision | Commit |
|---|---|---|---|
| 1 | Postgres key has wrong owner | ✅ Fixed | `1bbb7f84` |
| 2 | *bind* messages remain invisible | ⏭️ Deferred | |
| 3 | Read-only policy allows mutations | ✅ Fixed | `fa908ee6` |
| 4 | *redshift* assigned a new codec | ✅ Fixed | `d5e7e248` |
| 5 | hoopinspect excluded from test-oss | ✅ Fixed | `4a93ea2e` |
| 6 | Audit leak check misses PII | ✅ Fixed | `ff67a0c2` |

Two of these were real defects a reviewer would have had to run the stack to catch. The TLS key ownership one only fails outside Docker Desktop, whose bind-mount ownership remapping hides it. The read-only lane really did let `CREATE TABLE` through, verified by running it through Metabase before fixing it.

Three of the six agent prompts were not followed as written, each explained on its own thread: no `bypass_test.go` case for a config file under `deploy/` (backwards coupling), no invented MySQL roadmap in a README, and the PII list is read out of the seed rather than extended by hand.

Counts in the README and `config.yaml` moved with the policy fix. Re-measured twice from a clean `./run.sh`, identical both times: 99 statements, 92 allowed, 7 denied, 38 result sets, 11 connections. `e3ef6065` carries that.

Also re-verified after the fixes: `./run.sh down && ./run.sh && ./demo.sh` green (12 assertions, exit 0), `make test-hoopinspect` (16 packages, exit 0), `shellcheck`, `bash -n`, `py_compile`, `docker compose config -q`, `go run . -validate`.

---
*Generated by Qodo PR Resolver skill*

sandromello:
✅ Build Completed with Success, Version=1726.0.0-ge3ef606

hoopinspect/CLAUDE.md
p3rotto: Removed. I kept `GOOS=wasip1 GOARCH=wasm go build` in the checks, restated as a portability canary rather than an Envoy claim. Say the word and that goes too.

hoopinspect/CLAUDE.md
p3rotto: Taking your framing: the section now leads with each optional component being its own module so it plugs in or drops out on its own, and the zero-dependency root falling out of that rather than causing it. Whether it was the right call is yours; the file follows the tree.

hoopinspect/CLAUDE.md
p3rotto: It was not. `4a93ea2e` adds `make test-hoopinspect`, which discovers every nested `go.mod` and runs it, and makes `test-oss` depend on it: 16 packages, under 3 seconds, no services.

hoopinspect/CLAUDE.md
p3rotto: Fixed in `4a93ea2e`.

hoopinspect/CLAUDE.md
p3rotto: Added that line to the file. The split itself is yours to scope since you know what is in there. Want a ticket?

hoopinspect/CLAUDE.md
p3rotto: Claude Code reads `CLAUDE.md` but not `AGENTS.md` (with no fallback), so renaming it would just stop it from loading. 

IMHO, we should keep `CLAUDE.md` for now, and we can organize ourselves later with a repo-wide PR with `@AGENTS.md` imports, not this one.

p3rotto:
@matheusfrancisco Yes, and that is the whole integration: the host and port on the connection you already have. Metabase points at `hoop-inspect:15432`, the sidecar points at your `postgres:5432`, same credentials and same JDBC driver, no plugin and no agent in the container.

hoopinspect/CLAUDE.md
p3rotto: vai corinthians

sandromello:
✅ Build Completed with Success, Version=1726.0.0-gb94e43d

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1726.0.0-g42d6afd

PR #1728, opened 2026-08-19
DEP-147: correct stale claims in CLAUDE.md

racerxdl:
## Ticket
DEP-147

## What & why
CLAUDE.md had drifted from the code. Agents read it as authoritative, so a wrong path or env var name sends every session down the wrong road.

Three parts:

**1. Correct 19 stale claims.** Each verified against the source — wrong env var names (`MS_PRESIDIO_*`, `GCP_DLP_JSON_CREDENTIALS`), a directory that no longer exists (`gateway/aiclients/`), the wrong dev CLI output path, the wrong PR preview tag format, an interceptor chain that listed four when the code chains three, and `storagev2` described as storage when it is a request-scoped API context. Also documents `hoopinspect/`, which was absent despite being 7 of 13 workspace modules.

**2. Replace "No Hacks" prose with four decision tests.** The old rule was a synonym list — hack, workaround, monkey patch, partial solution — with no test separating them, plus "do not commit code that can break later," which is unfalsifiable. In practice it got ignored or used to block work someone disliked. Each new test has a visible failure mode a reviewer can point at:

- Compatibility shims are allowed when isolated, commented with the upstream cause and version, and covered by a test that fails once upstream fixes the defect.
- Do not depend on a condition the repo does not enforce. A comment describing the assumption does not enforce it.
- Incomplete work is fine when the incomplete path fails loudly. No empty struct, zero value, or `nil, nil` standing in for an unimplemented path. A log line is not loud enough.
- Do not expand scope to reach a fix. Report the missing piece instead.

**3. Fix the feature-flag model.** Flags were documented as always defaulting `false`, which contradicted both the catalog (6 of 13 ship `true`) and practice. The default now follows blast radius: a feature that only adds a new code path is born `true`; one that changes existing behavior is born `false` and flips after proving out. Names Settings → Experimental as the customer-facing kill switch, and states that a flag is not a way to ship unfinished work.

Adds two guidelines: gateway↔agent backward compatibility (agents run on customer infrastructure and may lag auto-deployed gateways), and referencing tickets by ID rather than URL, since these repos are public.

## How to test
Docs only — no build impact. To verify the corrected claims:

```
# renamed integration
ls gateway/aianalyzer && ! ls gateway/aiclients 2>/dev/null

# env var names as now documented
grep MSPRESIDIO_ANALYZER_URL gateway/appconfig/appconfig.go
grep GOOGLE_APPLICATION_CREDENTIALS_JSON gateway/appconfig/appconfig.go

# the Makefile target is swag-fmt, not `swag fmt`
grep -c '^swag-fmt:' Makefile

# 13 workspace modules, not six
grep -c $'\t\\./' go.work

# the flags cited as born-false-then-flipped
git log -S beta.oracle_native -- common/featureflag/featureflag.go

# the kill switch the docs now name
ls webapp_v2/src/pages/Settings/Experimental/index.jsx

# no duplicate headings
grep '^## ' CLAUDE.md | sort | uniq -d
```

Expected: every path exists, all greps hit, `uniq -d` prints nothing.

## Risks
Low for the diff itself — documentation only, no code, config, or build changes.

Worth knowing: the new rules make four existing spots visible violations. None are touched here.

- `gateway/storagev2/core.go:22` warns and returns an empty auth context — exactly what test 3 bans.
- `gateway/transport/plugins/audit/audit.go:267` is the repo's only shim comment, with no upstream cause and no test, so it fails test 1.
- `models.DB` global is read by 44 model files though the convention says pass `*gorm.DB`.
- `gateway/models/connections.go` swallows `gorm.ErrRecordNotFound` as `nil, nil` in six places.

The doc names the last two as non-patterns rather than hiding them. Fixes belong in the Go code, in a separate ticket.

Automated by MisterMal

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,14 +2,19 @@
 
 ## Project Overview
 
-This is a Go workspace for the hoop gateway, agent, and CLI. The codebase follows a monorepo-style structure using `go.work`, with six modules: `gateway/`, `agent/`, `client/`, `common/`, `libhoop/`, and `tunnel/`. There is also a Rust companion binary (`agentrs/`) for RDP/TLS proxy workloads, a legacy ClojureScript SPA (`webapp/`) - see `webapp/CLAUDE.md` for its own conventions, and a new React frontend (`webapp_v2/`) that is actively replacing it.
+A Go workspace (`go.work`) for the hoop gateway, agent, and CLI.
 
- `_libhoop/` is a symlink target; the build uses `ln -s _libhoop libhoop` (`make libhoop-map`) so Go sees the `libhoop` import path.
+- Product modules: `gateway/`, `agent/`, `client/`, `common/`, `libhoop/`, `tunnel/`.
+- `hoopinspect/` adds seven more workspace entries.
+- `agentrs/` — Rust companion binary for RDP/TLS proxy workloads.
+- `webapp/` — legacy ClojureScript SPA; see `webapp/CLAUDE.md` for its own conventions.
+- `webapp_v2/` — React frontend that is replacing it.
+- `_libhoop/` is a symlink target. The build runs `ln -s _libhoop libhoop` (`make libhoop-map`) so Go sees the `libhoop` import path.
 
 ## Toolchain & Prerequisites
 
 - Go >= 1.26, Rust + `cross` (for cross-compiled `agentrs`), Docker, Node/npm, Clojure/Java.
-- PostgreSQL is **mandatory** for the gateway (`POSTGRES_DB_URI`).
+- PostgreSQL is the default state store (`POSTGRES_DB_URI`). `POSTGRES_DB_URI=pglite:///<data-dir>` starts the embedded PGlite instead (`gateway/pglite`).
 - `golang-migrate` CLI for creating new SQL migration files.
 - `swag` (v1.16.3) for regenerating OpenAPI docs.
 - See `DEV.md` for full setup walkthrough.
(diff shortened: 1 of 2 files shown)


sandromello:
✅ Build Completed with Success, Version=1728.0.0-gf840f20

CLAUDE.md
```diff
@@ -258,16 +297,6 @@ When merging `main` into a feature branch:
 | Frontend migration context | `webapp_v2/CONTEXT_MIGRATION.md` |
 | Frontend coding rules | `webapp_v2/CLAUDE.md` |
 
-## Coding
```
matheusfrancisco, line 261: 🎉

CLAUDE.md
```diff
@@ -163,21 +185,30 @@ When implementing a new feature, behavior change, or non-trivial code path, **as
...
 ## Coding Conventions
 
+### No Hacks
+Code quality comes before immediate results.
+
+- Do not add a hack, workaround, monkey patch, or partial solution.
+- Do not commit code that can break later.
+- If the only way to ship a feature is a hack, stop. Either fix the cause in a robust, production-ready way, or say the request needs support the repository does not have.
+
+"I could not do X because the repository has no support for Y" is a good answer. Adding the missing support correctly is a better one. A workaround that breaks things later is the worst one.
```
matheusfrancisco: idk if it helps or makes the old no-hacking blocks a lot for me heheheheh

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1728.0.0-g9cd6922

PR #1729, opened 2026-08-19
EVL-211: in-app TTFV survey (frontend)

feraio:
## 📝 Description

Frontend half of TTFV (time-to-first-value) measurement: the in-app survey that asks an admin whether they've gotten what they needed out of hoop, and what it was.

TTFV is the gap between an org being created and the first moment an admin confirms value. The product is self-hosted, so there is no way to observe that moment — we have to ask. This is the asking; **#1730** ([EVL-212](https://linear.app/hoophq/issue/EVL-212/ttfv-measurement-backend)) is the gateway side that decides *when* to ask, records the answer, and emits the analytics event server-side.

**Ships together with #1730.** This side fails closed, so it is technically safe alone — visibility comes entirely from `show_ttfv_survey` on `GET /userinfo`, and a gateway that doesn't know the field leaves it undefined, which reads as "don't ask". But #1730 alone would make the gateway report `show_ttfv_survey: true` with no widget to render it, so the metric would read zero and look like "shipped, nobody answers". If they ever do land separately, **this one goes first.**

## 📣 User-facing impact

Admins may see a small, dismissable survey in the corner of the app asking whether hoop has done what they needed. Answering it once positively retires it for the whole organization; declining it quiets it for a few days. Organizations with analytics disabled are never asked.

## 🔗 Related Issue

Linear: [EVL-211 — TTFV Measurement Frontend](https://linear.app/hoophq/issue/EVL-211/ttfv-measurement-frontend)
Backend counterpart: #1730 — [EVL-212](https://linear.app/hoophq/issue/EVL-212/ttfv-measurement-backend)

## 🚀 Type of Change

- [x] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [x] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

Four commits, in the order they're best reviewed:

- **`a87d0e81` — extract `components/SurveyWidget`.** The origin survey (#1701, EVL-185) had the launcher, card, acknowledgement and lifecycle all inline. TTFV needs every one of those, so the shell moved out and `OriginSurvey` now renders through it — **−180 lines, no behaviour change**. `SurveyWidget` owns presentation and the collapsed → open → closed lifecycle; the feature owns visibility (`due`) and submission (`answered`). `SurveyWidget.ChoiceQuestion` is the single-choice question with a free-text follow-up on one designated option, and it's *controlled* — the card unmounts when collapsed, so the answer has to live in the feature or it's lost on every collapse.
- **`380c4820` — the survey itself.** `features/TtfvSurvey`, `services/ttfvSurvey.js` (`POST /orgs/ttfv-survey`), and `features/Surveys`, a coordinator that picks the one survey to show. `App.jsx` now mounts the coordinator instead of a survey.
- **`6d01c7d5` — stop naming the gateway's eligibility clauses in frontend prose.** Comments used to enumerate them (feature flag, cooldown, terminal Yes…). The frontend re-implements none of that policy, so naming it just creates a second place to update.
- **`439c1fc9` — rewrite the survey copy and the activity options.** See below.

### Bug fixed on the way through

The **already-shipped** origin survey card had:

```css
max-height: calc(50vh - var(--mantine-spacing-xxl));
```

There is no `--mantine-spacing-xxl`. `theme.js` adds the 4/8/16/24/32/48/64 scale under `*Alt` keys and leaves Mantine's own `xs`–`xl` defaults in place. An undefined custom property doesn't fall back — it invalidates the whole declaration — so the card had **no `max-height` at all** and ran off the top of short viewports instead of scrolling. Now `xxlAlt`.

`webapp_v2/CLAUDE.md` had the same wrong table, which is how the bug got written in the first place; it's corrected, with a warning about this specific silent-failure mode.

### Copy: asking about value received, not setup completed

The original wording asked *"Did you get done what you came here to do?"* and *"What did you get done **today**?"*. Both scope the answer to the current visit — but TTFV is the first time value was **ever** reached. An admin whose first value happened last week answers "No" truthfully, and a "No" costs a cooldown before we may ask again. The copy was delaying the measurement it exists to take.

```
Step 1  Have you gotten what you needed yet?
Step 2  What have you gotten done so far?      (only after a Yes)
```

"yet" and "so far" are load-bearing — there's a comment in `constants.js` saying so. The product is deliberately not named: naming it invites a verdict on the product, where what we need is whether the admin accomplished their own goal.

### Activity options — shared enum with the gateway

| `value` | Label |
| -- | -- |
| `saw-guardrail-applied` | Saw a guardrail apply to a live session |
| `saw-data-masked` | Saw sensitive data masked in a session |
| `approved-or-denied-access-request` | Approved or denied an access request |
| `reviewed-recorded-session` | Reviewed a recorded session |
| `opened-ai-analyzed-session-report` | Opened an AI-analyzed session report |
| `other` | Other (please specify) |

**Every option is a moment the product did something *for* the admin, never one where they configured it** — that's the bar for anything added later. Connecting a resource, creating a policy and setting up a masking rule are setup, costs paid up front, so a "Yes" backed by one of them would stamp TTFV at the moment configuration finished, which is the exact thing this metric exists to be distinguished from. Hence no resource option, and the guardrail/masking entries describe the rule being *seen to apply* rather than created.

These `value`s are the analytics contract, duplicated in `validTTFVActivities` on the gateway. **Never repurpose one** — it silently reinterprets historical rows and reports.

### Test Configuration:
- **Browser(s)**: Chrome, macOS
- **OS**: macOS 15 (darwin 25.4.0)
- **Gateway**: local `hoopdev` built from `felipe/evl-212-ttfv-measurement-backend` at tip (OIDC auth); `POST /api/orgs/ttfv-survey` returns 401 unauthenticated, confirming the route is present

### Tests performed:
- [ ] Unit tests pass — *`webapp_v2` has no test runner (no vitest/jest/testing-library in `package.json`), so there is nothing to run*
- [ ] Integration tests pass — *none exist for this surface*
- [x] Manual testing completed — **partial; scope recorded below**

**Verified in the browser against a live EVL-212 gateway:**

| Scenario | Result |
| -- | -- |
| Widget appears for an admin of an eligible org (1 session, `analytics_mode = identified`) | ✅ |
| Non-admins see nothing | ✅ |
| "Other" keeps Submit disabled until the free text is non-blank; input caps at 255 | ✅ |
| The X collapses to the launcher and POSTs nothing; widget returns on reload | ✅ |
| Short viewport (~500px tall) — card caps its height and scrolls instead of running off the top | ✅ **this is the `max-height` regression fixed here** |

**Not yet exercised**, and honestly flagged rather than assumed:

- The **write path end to end** — no answer has been submitted, so no row has been written and the analytics event has never been observed. `private.ttfv_survey_responses` is empty.
- **Terminal Yes / 409** and the **decline + cooldown** path.
- **Origin-vs-TTFV priority** when both surveys are due (only TTFV was present).
- **Fail-closed against a gateway without EVL-212** — reasoned, not tested: the field is simply absent from `/userinfo` and deserialises to falsy.
- **`analytics_mode = 'disabled'` suppression.**

**`npm run lint`** — no new problems. 15 errors remain, all pre-existing in four files this branch never touches (`vite.config.js`, `CommandPaletteRoot.jsx`, `pages/Auth/Login/index.jsx`, `pages/EventRouting/Form/index.jsx`), verified against `git diff --stat`.

**`npm run build`** — green.

### Contract cross-checks against #1730

The POST is fire-and-forget (see below), so a 400 the client fails to prevent is invisible — no row, no event, no error, just a quietly low response rate. These were therefore verified statically rather than left to runtime:

1. **`ACTIVITY_OPTIONS` matches `validTTFVActivities` element for element, order included.** Compared programmatically by parsing both lists, not by eye. `ACTIVITY_OTHER === 'other'` and it is the last element on both sides.
2. **`reached_value` is never `undefined` or `null`** — the Yes path sends a literal `true`, the decline path a literal `false`. The gateway 400s on a missing value, and this is the path that would have hit it.
3. **`activity` is sent as `''` on a decline**, which the gateway ignores and stores as NULL.
4. **`activity_other` can never exceed the server's 255-*rune* cap.** The DOM's `maxLength` counts UTF-16 code units and an astral character is 2 units but 1 rune, so the browser is strictly stricter — never looser. Non-blank is enforced client-side before Submit enables.
5. **`POST /orgs/ttfv-survey`'s role (`AdminOnlyAccessRole`) agrees with the role gating `show_ttfv_survey` on `/userinfo`** — a mismatch would offer the survey to someone the POST then rejects.
6. **409 already behaves as the gateway wants it to.** It is swallowed like any other rejection: the acknowledgement shows, the widget closes, nothing is retried, no error is surfaced, and `/userinfo` won't offer it again on the next boot — which is exactly "treat it like a successful terminal Yes, another admin answered first".

`TestTTFVActivityContract` on #1730 pins the server side, but it cannot see the JS file — cross-repo-file drift is caught by review only.

### How to test

Requires the #1730 gateway. Against `main`'s gateway the widget simply never appears, which is itself the fail-closed case worth confirming first.

```bash
make run-dev-postgres                       # container hoopdevpg, host port 5449
HOOP_RS_BUILD=0 make run-dev                # run from the EVL-212 worktree, not this branch —
                                            # the image snapshots source at build time, and a
                                            # frontend-only build makes the endpoints 404
cd webapp_v2 && npm run dev:full            # Vite :5173 + shadow-cljs :8280
```

Open **http://localhost:5173** (not :8009 — only Vite proxies both `/api` and the CLJS assets). Sanity check the gateway first: `curl -o /dev/null -w '%{http_code}\n' -X POST localhost:8009/api/orgs/ttfv-survey` → **401** means the backend is there, **404** means it was rebuilt over.

1. **Fail-closed.** Point at a gateway without #1730. No launcher appears. `GET /api/userinfo` has no `show_ttfv_survey` field.
2. **It appears.** Admin of an org that has run at least one session — a round hoop-symbol launcher with a red dot on the right edge, vertically centred. Non-admins see nothing.
3. **Only one survey at a time.** An admin of a brand-new org qualifies for both. Only the **origin** survey shows — its window is the one that closes (7 days, once per user). After answering it, TTFV should *not* pop up until the next page load, because `/userinfo` is fetched once per boot.
4. **The No path.** Open → "Have you gotten what you needed yet?" → **No**. "Answer received" for 3s, then the widget disappears. One row with `reached_value = false`. Reload: gone, for the 3-day cooldown.
5. **The Yes path.** **Yes** advances to "What have you gotten done so far?" rather than submitting — a bare "reached value" with no activity isn't worth a row, and the first Yes is terminal. Pick an option → Submit → row with `reached_value = true` and the `activity`.
6. **Other requires text.** Submit stays disabled until non-blank; the input stops accepting at 255 characters.
7. **Terminal Yes.** Reload after a Yes — no launcher, ever again for that org. A hand-rolled POST returns **409**.
8. **X is not an answer.** Open, click X. Collapses to the launcher, nothing POSTed. Reload — it's back.
9. **Short viewport.** ~500px tall, open the card — it caps and scrolls rather than running off the top. Worth checking the origin survey too.
10. **Analytics off.** `analytics_mode = 'disabled'` → never appears.

Order matters for 4/5/7: a Yes is terminal and a No starts a 3-day cooldown, so reset between runs with `DELETE FROM private.ttfv_survey_responses;` and reload.

To be added before this leaves draft.

## ✅ Checklist

- [ ] New and existing unit tests pass locally with my changes — *no test runner in `webapp_v2`*

## 📄 Additional Notes

Draft until the write path has been exercised end to end and screenshots are attached. Neither is a code concern — the remaining risk is entirely "has anyone watched a row get written", and the answer is currently no.

### What reviewers should push back on

- **Fire and forget.** The POST is not awaited and its rejection is swallowed to a `console.warn`; the acknowledgement shows on the click. This matches the origin survey, and the reasoning is that a failed write leaves `/userinfo` still reporting the survey, so the admin is asked again and has nothing to act on meanwhile — a failure toast would be pure noise. It also gives the 409 semantics the gateway wants for free. **The cost is that a 400 is indistinguishable from a 204 in the UI**, which is what the six static cross-checks above are compensating for. If you think that trade is wrong, this is the place to say so.
- **No policy on the client.** Not the admin check, not org eligibility, not the terminal Yes, not the cooldown — and `6d01c7d5` removed even the *enumeration* of those clauses from comments. This already paid off: #1730 subsequently added a one-session precondition and cut the cooldown from 7 days to 3, and this branch needed zero edits. **If a duration, threshold or eligibility rule ever appears in `webapp_v2`, treat it as a bug.**
- **`due` is only checked while collapsed** — a card the user already opened stays open even if the flag changed underneath it.
- **The extraction touches shipped code.** `OriginSurvey` is live; worth a look on its own to confirm nothing regressed beyond the `max-height` fix.

### DIFF
--- a/webapp_v2/CLAUDE.md
+++ b/webapp_v2/CLAUDE.md
@@ -292,8 +292,12 @@ CSS Modules are allowed **only** for complex selectors that Mantine props cannot
 Available Mantine CSS variables (set by the theme in `src/theme.js`):
 
 ```css
-/* Spacing — xs=4px sm=8px md=16px lg=24px xl=32px xxl=48px xxxl=64px */
-var(--mantine-spacing-xs | sm | md | lg | xl | xxl | xxxl)
+/* Spacing — TWO scales live side by side, and the `*Alt` suffix is part of the
+   variable name. `theme.js` only overrides the *Alt keys; Mantine deep-merges
+   them with its own defaults, which stay untouched. There is NO
+   --mantine-spacing-xxl / -xxxl — those exist only with the Alt suffix. */
+var(--mantine-spacing-xs | sm | md | lg | xl)                     /* Mantine: 10 12 16 20 32px */
+var(--mantine-spacing-xsAlt | smAlt | mdAlt | lgAlt | xlAlt | xxlAlt | xxxlAlt)  /* ours: 4 8 16 24 32 48 64px */
 
 /* Font sizes — xs=12px sm=14px md=16px lg=18px xl=20px */
 var(--mantine-font-size-xs | sm | md | lg | xl)
@@ -318,6 +322,14 @@ var(--mantine-color-{name}-light-color)  /* light variant text */
 /* ✅ Correct — always reference the theme */
 .label { font-size: var(--mantine-font-size-xs); margin-bottom: var(--mantine-spacing-sm); color: var(--mantine-color-indigo-8); }
 ```
+
+**A misspelled variable fails silently — check the name against the list above.**
+An undefined custom property doesn't fall back to the default, it invalidates the
+whole declaration, so the property is simply dropped. Inside a `calc()` this is
+especially quiet: `max-height: calc(50vh - var(--mantine-spacing-xxl))` had no
+`max-height` at all in the shipped origin survey, and the card ran off the top of
+short viewports instead of scrolling. If a rule looks like it isn't applying,
+inspect it in DevTools — an invalid declaration shows struck through.
 
 ## Text color — use Mantine tokens, not raw names
 
--- a/webapp_v2/COMPONENTS.md
+++ b/webapp_v2/COMPONENTS.md
@@ -818,7 +818,7 @@ Non-obvious notes only:
 - `originSurvey.js` / `ttfvSurvey.js` — the in-app surveys. Both are write-only:
   there is no GET, because whether to ask is a boolean on `/userinfo`
   (`show_origin_survey`, `show_ttfv_survey`) and the whole policy — admin check,
-  feature flag, cooldown, already-answered — lives in the gateway. Both POST
+  org eligibility, cooldown, already-answered — lives in the gateway. Both POST
   return 204, and 409 on a repeat answer, so callers fire and forget rather than
   surfacing an error the user can't act on. On `ttfvSurvey.answer`, `activity` is
   only read when `reachedValue` is true and `activityOther` only when `activity`
--- a/webapp_v2/COMPONENTS.md
+++ b/webapp_v2/COMPONENTS.md
@@ -620,6 +620,39 @@ Props: `onAdd()`, `selectMode`, `onToggleSelectMode()`, `allSelected`, `onToggle
 
 ---
 
+### `SurveyWidget`
+The shell every in-app survey is rendered in: a floating launcher pinned to the right edge of the viewport that expands into a non-blocking card, plus the "Answer received" acknowledgement that replaces the question and retires the widget. It owns presentation and the collapsed/open/closed lifecycle only — the survey feature owns visibility (`due`) and submission (`answered`).
+```jsx
+import SurveyWidget from '@/components/SurveyWidget'
+import { EMPTY_CHOICE_ANSWER } from '@/components/SurveyWidget/constants'
+
+const due = useUserStore((state) => !!state.user?.show_origin_survey)
+const [answer, setAnswer] = useState(EMPTY_CHOICE_ANSWER)
+const [answered, setAnswered] = useState(false)
+
+<SurveyWidget label={QUESTION} due={due} answered={answered}>
+  <SurveyWidget.ChoiceQuestion
+    question={QUESTION}
+    options={OPTIONS}
+    answer={answer}
+    onAnswerChange={setAnswer}
+    otherValue="other"
+    otherMaxLength={255}
+    otherLabel="Specify how you heard about Hoop"
+    onSubmit={handleSubmit}
+  />
+</SurveyWidget>
+```
+Props: `label` (announced on both launcher and card — keep it stable across a multi-step survey), `due`, `answered`, `children`.
+
+`SurveyWidget.ChoiceQuestion` is a single-choice question with a free-text follow-up on one designated option plus its Submit button; Submit stays disabled until the free text is filled in for that option. It is **controlled** — the card unmounts when the user collapses it, so the `{ value, otherText }` answer must live in the feature, which stays mounted. Seed it with `EMPTY_CHOICE_ANSWER`.
+
+Two rules when adding a survey:
+- Mount it through `features/Surveys`, never in `App.jsx` directly — every survey anchors to the same spot, so only one may be on screen at a time and that coordinator is what picks.
+- Read visibility from a `/userinfo` boolean (`show_origin_survey`, `show_ttfv_survey`). A gateway that doesn't know the field yet leaves it undefined, which reads as "don't ask" — the widget fails closed.
+
+---
+
 ### `FullBleed` (`src/layout/FullBleed/`)
 Lets a page render edge-to-edge and exactly one viewport tall **inside** the padded `PageLayout` — cancels the page padding (single-sourced from `PageLayout`'s `PAGE_PADDING`) and fills the `AppShell.Main` height. Use for hero/promotion panels.
 ```jsx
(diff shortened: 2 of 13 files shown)


sandromello:
✅ Build Completed with Success, Version=1729.0.0-g439c1fc

feraio:
## Closed unmerged — project frozen 2026-09-01

Closing because the work is being shelved, not because anything is wrong with it. The Gateway is no longer a priority: we are moving to a new architecture in which the webapp becomes a simpler **Control Plane** with a different way to connect and configure.

Worth recording for whoever finds this later: measuring TTFV by survey was a consequence of hoop being **self-hosted** — the gateway cannot phone home, so the only way to observe the moment value was reached was to ask. A hosted control plane removes that constraint entirely, which means this approach would most likely be replaced rather than ported.

**The commits are preserved.** Tag `archive/evl-211-ttfv-frontend` → `439c1fc9`, so they stay reachable even if the branch is pruned. This PR body remains the design record.

**Closing rather than merging was deliberate.** This side fails closed, so merging it alone would have been harmless — and pointless without #1730. #1730 alone is worse than either: the gateway would advertise `show_ttfv_survey: true` with no widget to render it. Both close together, as they were always going to ship together.

**Knowingly left on the table.** The `max-height` fix inside the extracted `SurveyWidget` CSS also fixes a live bug in the *shipped* origin survey on `main`: `--mantine-spacing-xxl` does not exist, an undefined custom property invalidates the entire `calc()`, so the card has no `max-height` and runs off the top of short viewports instead of scrolling. No user reports and no repro, so we are accepting it rather than opening a separate PR against a product being wound down. If it is ever reported, the one-line fix and the reasoning are in `a87d0e81` and in this branch's `webapp_v2/CLAUDE.md`.

**If this is ever revived:** the write path was never exercised end to end — no row was ever written from this UI and no Segment event was ever observed. What *was* verified is listed in the description.

Related: #1730 · EVL-211

PR #1730, opened 2026-08-19
EVL-212: measure time to first value with an in-app survey

feraio:
## 📝 Description

Backend half of TTFV (Time To First Value) measurement. TTFV is defined as the
duration between `orgs.created_at` and the moment an **administrator first
confirms they got what they came for** — a perceived outcome, not a completed
setup step. Nothing in the product records that moment today, so this PR adds
the survey that asks for it and the storage/analytics path that measures it.

Two pieces:

- `show_ttfv_survey` on `GET /api/userinfo` — the full ask policy, resolved
  server-side. True for an administrator of an organization that collects
  analytics, has run at least one session, has never confirmed value, and has
  not declined within the last 3 days.
- `POST /api/orgs/ttfv-survey` (admin only) — records the answer and emits
  `hoop-ttfv-survey-answered` with the measured duration.

A **"yes" is terminal** — the organization is never asked again, and a second
attempt gets a 409. A **"no" is not**: it is recorded as a real data point and
followed by another ask after a 3-day cooldown. Dismissing the widget is not an
answer and is not submitted.

Design notes worth a reviewer's attention:

- **No feature flag.** The survey is on for every organization and closes itself
  on the first confirmed value. Analytics mode is the opt-out that already
  exists — an organization with `analytics_mode = 'disabled'` is never asked,
  and the clause is evaluated in SQL rather than through `analytics.GetMode`,
  which is a process-local cache that is stale in the consent direction.
- **Organizations that never ran a session are never asked.** The question is
  meaningless before there is a product to have gotten value from, and a decline
  from an empty org would add a cooldown's worth of noise to the metric.
- **The cooldown is the metric's resolution.** Value is reached somewhere
  between the last decline and the yes, but only the yes carries a timestamp, so
  every measurement is inflated by up to one cooldown. 3 days keeps that error
  small without putting the question in front of a daily administrator more than
  about twice a week.
- **Concurrency is arbitrated by the database.** A partial unique index
  `ON ttfv_survey_responses (org_id) WHERE reached_value` is what decides between
  two simultaneous confirmations; the loser's 23505 surfaces as
  `gorm.ErrDuplicatedKey` and is reported as a 409. The write does not rely on a
  read-then-check.
- **The free text never reaches an analytics destination.** It is user-supplied
  and may contain personal data, so it is stored in
  `private.ttfv_survey_responses` and deliberately excluded from the event
  payload.
- **The event is emitted from the gateway, not the browser.** A client-side call
  is blocked by the same ad blockers that ruled out measuring this through
  Intercom, which would bias the metric towards the users least likely to run
  one. The row is the system of record; the event is the copy dashboards read,
  and the duration is recomputable from SQL if the event is ever dropped.

## 📣 User-facing impact

Administrators are asked once in a while whether they have gotten what they
needed from hoop yet, and the question stops for good once the answer is yes.
Organizations with analytics disabled never see it.

## 🔗 Related Issue

[EVL-212 — TTFV Measurement Backend](https://linear.app/hoophq/issue/EVL-212/ttfv-measurement-backend)

Pairs with **EVL-211 (frontend, `webapp_v2`)**, which opens its own PR. The two
are independent at the code level — the widget fails closed if `show_ttfv_survey`
is absent — but they should be **released together**, since neither is useful
alone. The activity enum is a shared contract between this PR's
`validTTFVActivities` and `webapp_v2/src/features/TtfvSurvey/constants.js`;
please check the two lists agree element-for-element at merge time.

## 🚀 Type of Change

## 📋 Changes Made

- **Migration `000114_ttfv_survey_responses`** — new table, an
  `(org_id, created_at DESC)` index for the "latest answers" reads, and the
  partial unique index that enforces one confirmed value per organization.
- **`gateway/models/ttfvsurvey.go`** — `ShouldShowTTFVSurvey` (the whole policy
  as one SQL predicate, with the time arithmetic done in Postgres so both sides
  of every comparison come from one clock) and `CreateTTFVSurveyResponse`
  (conditional insert, computes `ttfv_seconds` in the same statement, maps
  duplicates to `ErrAlreadyExists` and distinguishes unknown-org from
  already-answered).
- **`gateway/services/ttfvsurvey.go`** — cooldown constant, caller-dependent
  short-circuits (`!isAdmin || isAnonymous`), and the answer type.
- **`gateway/api/orgs/orgttfvsurvey.go`** — handler, the activity enum,
  request validation, and the pure `ttfvEventProperties` builder.
- **`gateway/api/user/user.go`** — resolves `show_ttfv_survey`; a failure only
  means the survey is not offered on that request, never a failed `/userinfo`.
- **`gateway/api/server.go`** — route registration under `AdminOnlyAccessRole`.
- **`gateway/analytics/events.go`** — new `EventTTFVSurveyAnswered` constant
  (`hoop-ttfv-survey-answered`). No existing event is removed or renamed. It is
  emitted from the handler rather than via `TrackRequest`, because that
  middleware runs before the body is validated and could carry neither the
  answer nor the measured duration.
- **OpenAPI** — annotations plus regenerated `autogen/docs.go` and
  `openapiv3.json`.

### Test Configuration:
- **Browser(s)**: n/a — API only
- **OS**: macOS (darwin), gateway in Docker against the shared dev Postgres

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass — n/a, no integration suite covers this layer

**Automated.** `CGO_ENABLED=0 go test github.com/hoophq/hoop/...` is green.
`make test-oss` could not be run on this machine — its `generate-wasm`
prerequisite needs `cargo`, which is not installed here — so please confirm it
is green in CI. New tests:

- `gateway/models/ttfvsurvey_test.go` — the ask policy as a table against real
  Postgres (pglite), 8 cases including both sides of the cooldown boundary
  (declined 2d23h ago → suppressed, 3d1h ago → asked) and an organization that
  never ran a session.
- `gateway/api/orgs/orgttfvsurvey_test.go` — request validation, the binding
  behaviour that lets an explicit `false` through, the event payload key by key
  (including a test that the free text cannot leak into it), and
  `TestTTFVActivityContract`, which pins the enum literally and asserts the
  published `enums` struct tag agrees with it.

**Manual end-to-end**, against a gateway built from this branch talking to the
dev Postgres, driving the real HTTP API with a throwaway admin:

| Scenario | Expected | Result |
|---|---|---|
| Org with zero sessions | `show_ttfv_survey: false` | ✅ |
| After one real session (`POST /api/sessions`) | `true` | ✅ |
| Declined 2d23h ago | `false` | ✅ |
| Declined 3d1h ago | `true` | ✅ |
| `analytics_mode = 'disabled'` | `false` | ✅ |
| Anonymous user | `false` | ✅ |
| `{"reached_value": false}` sent with empty-string activity fields | 204, `activity`/`activity_other` stored as NULL | ✅ |
| `{"reached_value": true, "activity": "other", ...}` with accents and emoji | 204, stored intact | ✅ |
| `/userinfo` after a confirmed value | `false` | ✅ |
| Second confirmed value | 409 | ✅ |
| Unknown activity identifier | 400 | ✅ |

Test artifacts (the throwaway admin, its group and token, the survey rows, the
analytics-mode change) were cleaned up. The one real session row was left in
place — it is genuine audit data, and it leaves the org in the state the EVL-211
session needs in order to see the widget.

## 📄 Additional Notes

**The one thing a reviewer cannot catch by reading the diff:** the
Segment → Mixpanel leg is **unverified**. `segmentApiKey` is injected only at
build time (`Makefile:39`) and `scripts/dev/run.sh:65` omits it, so
`analytics.New()` returns a client with a nil `Client` and every `Track` call is
a no-op locally. The event payload is asserted by unit test and the call site is
in place, but nobody has watched `hoop-ttfv-survey-answered` actually land. If
you have Segment access, please confirm the event and its properties (`org-id`,
`reached-value`, `activity`, `org-created-at`, `ttfv-seconds`) arrive once this
ships.

**Migration rollback was not exercised.** The dev Postgres is shared with the
concurrent EVL-211 session, so running `migrate down 1` would have been visible
to them. The `.down.sql` is a plain `DROP TABLE` of a table nothing else
references.

**Worth reviewing specifically:** the SQL in `models.ShouldShowTTFVSurvey`. The
whole policy lives there, and a wrong clause is the difference between a survey
nobody sees and one that nags.

---

### DIFF
--- a/common/featureflag/featureflag.go
+++ b/common/featureflag/featureflag.go
@@ -117,13 +117,6 @@ var catalog = map[string]Flag{
 		Stability:   StabilityExperimental,
 		Components:  []Component{ComponentGateway, ComponentAgent},
 	},
-	"experimental.ttfv_survey": {
-		Name:        "experimental.ttfv_survey",
-		Description: "Offer the in-app TTFV survey (\"Did you get done what you came here to do?\") to admins, so the time between the organization being created and the first confirmed value can be measured. Answers are stored server-side and the event is emitted from the gateway, not the browser. When off, /userinfo reports show_ttfv_survey false and no widget renders anywhere.",
-		Default:     false,
-		Stability:   StabilityExperimental,
-		Components:  []Component{ComponentGateway},
-	},
 	"experimental.mcp_gateway": {
 		Name:        "experimental.mcp_gateway",
 		Description: "Offer the MCP Gateway (mcpproxy) resource type in the webapp catalog, so admins can create protocol-aware MCP connections: tool-level allow/deny, per-tool approval, rug-pull detection and structured tool-call audit, over remote (streamable-http/sse) or stdio backends run either on the agent or on each user's own machine. When off, the card is hidden and no new MCP Gateway connection can be created; connections that already exist keep working end to end, and their role form, edit view and Connect modal stay reachable.",
--- a/common/featureflag/featureflag.go
+++ b/common/featureflag/featureflag.go
@@ -117,6 +117,13 @@ var catalog = map[string]Flag{
 		Stability:   StabilityExperimental,
 		Components:  []Component{ComponentGateway, ComponentAgent},
 	},
+	"experimental.ttfv_survey": {
+		Name:        "experimental.ttfv_survey",
+		Description: "Offer the in-app TTFV survey (\"Did you get done what you came here to do?\") to admins, so the time between the organization being created and the first confirmed value can be measured. Answers are stored server-side and the event is emitted from the gateway, not the browser. When off, /userinfo reports show_ttfv_survey false and no widget renders anywhere.",
+		Default:     false,
+		Stability:   StabilityExperimental,
+		Components:  []Component{ComponentGateway},
+	},
 	"experimental.mcp_gateway": {
 		Name:        "experimental.mcp_gateway",
 		Description: "Offer the MCP Gateway (mcpproxy) resource type in the webapp catalog, so admins can create protocol-aware MCP connections: tool-level allow/deny, per-tool approval, rug-pull detection and structured tool-call audit, over remote (streamable-http/sse) or stdio backends run either on the agent or on each user's own machine. When off, the card is hidden and no new MCP Gateway connection can be created; connections that already exist keep working end to end, and their role form, edit view and Connect modal stay reachable.",
--- a/gateway/analytics/events.go
+++ b/gateway/analytics/events.go
@@ -74,6 +74,11 @@ const (
 	// onboarding
 	EventOnboardingOriginAnswered = "hoop-onboarding-origin-answered"
 
+	// ttfv
+	// Emitted server-side only: a browser-side call would be blocked by the
+	// same ad blockers that ruled out measuring this through Intercom.
+	EventTTFVSurveyAnswered = "hoop-ttfv-survey-answered"
+
 	// search api
 	EventSearch = "hoop-search"
 
--- a/gateway/api/openapi/autogen/docs.go
+++ b/gateway/api/openapi/autogen/docs.go
@@ -5873,6 +5873,67 @@ const docTemplate = `{
                 }
             }
         },
+        "/orgs/ttfv-survey": {
+            "post": {
+                "description": "Record whether an administrator got done what they came to do, and measure the time since the organization was created. A confirmed value is terminal — the organization is never asked again and a second attempt returns 409 — while a decline is recorded and followed by another ask after a cooldown. Dismissing the widget is not an answer and must not be submitted.",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "Server Management"
+                ],
+                "summary": "Answer TTFV Survey",
+                "parameters": [
+                    {
+                        "description": "The request body resource",
+                        "name": "request",
+                        "in": "body",
+                        "required": true,
+                        "schema": {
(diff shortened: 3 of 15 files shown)


sandromello:
✅ Build Completed with Success, Version=1730.0.0-g1a62553

feraio:
## Closed unmerged — project frozen 2026-09-01

Same reason as #1729: the work is complete and was verified end to end, but the Gateway is deprioritised in favour of a new **Control Plane** architecture. Nothing here failed review.

**Commits preserved** at tag `archive/evl-212-ttfv-backend` → `1a625532`.

**Why this half in particular must not merge alone.** It would make `/userinfo` report `show_ttfv_survey: true` to admins with no widget in the webapp to render it. Every eligible org would be counted as asked, none would answer, and `ttfv-seconds` would read zero — indistinguishable from "shipped, and nobody reaches value". It would also land a migration and a new table on a gateway being wound down.

**No analytics contract is broken by closing this.** `hoop-ttfv-survey-answered` was never merged, so no emission site is being removed and no metric on `main` goes to zero. One thing to pass on, per the repo's analytics-as-contract rule: if anyone pre-built a Mixpanel/PostHog dashboard against that event name in anticipation, it will read zero permanently — whoever requested TTFV should hear that from us rather than discover it.

**Nothing to clean up.** The migration only ever ran against the local dev database, which no longer exists.

**If revived:** the migration was `000114` and that number will have been reused by then — it needs renumbering before the branch will build.

Related: #1729 · EVL-212

PR #1736, opened 2026-08-21
feat: enhance token validation to include client_id in audience checks

### LINKED ISSUE #1734: [Bug]: MCP OAuth static-client validation ignores the client_id claim
## Description

This is a follow-up to #1645 and the implementation in #1646.

Static client registration now works for IdPs without DCR, but the resource-server validation added by #1646 only recognizes a configured static client through `aud` or `azp`. Some IdPs, including JumpCloud Hydra, emit the OAuth client identity in the `client_id` access-token claim while leaving `aud` empty and omitting `azp`.

As a result, a correctly signed token is rejected even when its `client_id` exactly matches the configured MCP static client.

The comment above `VerifyAccessToken` already says that both `azp` and `client_id` are checked, but the MCP validation path does not read `client_id`.

## Environment

- Self-hosted Hoop gateway 1.144.0
- Also verified that the relevant implementation remains unchanged in 1.147.0 and current `main`
- OIDC issuer: JumpCloud, `https://oauth.id.jumpcloud.com/`
- Public static OAuth client with PKCE

## Reproduction

1. Enable MCP OAuth with a static client ID.
2. Obtain a valid JumpCloud JWT whose `client_id` is the same configured value.
3. Present the JWT to `/api/mcp`.

Relevant sanitized claims:

```json
{
  "aud": [],
  "client_id": "<configured-static-client-id>",
  "iss": "https://oauth.id.jumpcloud.com/",
  "sub": "<subject>"
}
```

Result:

```text
401 invalid_token: audience mismatch, got aud=[] azp=<nil>
```

The same result occurs when Hoop is temporarily configured with the client ID that minted a real `hoop login` token, proving that a matching `client_id` is not consulted. Reusing the web client is not the proposed production configuration; a dedicated MCP client should still be required.

## Root cause

`gateway/idp/oidc/oidc.go`, `tokenBoundToClient`, reads only `azp` and `aud`:

```go
azp, _ := claims["azp"].(string)
...
if azp == clientID || audienceContainsExact(claims["aud"], clientID) {
    return true
}
```

With JumpCloud tokens, `azp` is absent and `aud` is empty. The matching `client_id` claim is ignored.

## Proposed fix

Read `client_id` as a string and compare it exactly and case-sensitively against the explicitly configured static client IDs:

```go
azp, _ := claims["azp"].(string)
tokenClientID, _ := claims["client_id"].(string)
...
if tokenClientID == clientID || azp == clientID || audienceContainsExact(claims["aud"], clientID) {
    return true
}
```

Also include `client_id` in the audience-mismatch diagnostic.

This should remain limited to explicitly configured static client IDs. Since the IdP supplies no resource audience, the configured public PKCE client must be dedicated to MCP access for that gateway, matching the existing security requirement for the `azp` fallback.

## Regression tests

Please add cases to `TestTokenBoundToClient` for:

- empty `aud`, absent `azp`, and matching `client_id` is accepted
- unrelated `client_id` is rejected
- `client_id` comparison is exact and case-sensitive
- empty configured client IDs never match
- existing `aud` and `azp` behavior remains unchanged

## Additional issue

JumpCloud groups are nested under `ext`, and the MCP `extractGroups` path only performs a top-level lookup. That is independent of token binding and will be reported separately.

EmanuelJr:
## What & Why

Fixes #1734 

MCP OAuth static-client validation only recognized a configured static client through the aud or azp claims. Some IdPs (e.g. JumpCloud Hydra) emit the OAuth client identity in the client_id access-token claim (RFC 9068 §2.2) while leaving aud empty and omitting azp — so a correctly signed token whose client_id exactly matched the configured MCP static client was still rejected with 401 invalid_token: audience mismatch.

Changes in gateway/idp/oidc/oidc.go:

- tokenBoundToClient now also reads the client_id claim and compares it exactly and case-sensitively against the explicitly configured static client IDs, alongside the existing aud and azp checks. Empty configured IDs still never match.
- The audience-mismatch error now includes client_id in the diagnostic for easier debugging.
- Doc comment updated to describe the third binding path.

This remains limited to explicitly configured static client IDs, matching the existing security posture of the azp fallback: since the IdP supplies no resource audience, the configured public PKCE client must be dedicated to MCP access for that gateway.

### DIFF
--- a/gateway/idp/oidc/oidc.go
+++ b/gateway/idp/oidc/oidc.go
@@ -397,8 +397,8 @@ func (p *Provider) VerifyAccessTokenForResource(accessToken, expectedResource st
 	}
 
 	if !audienceContains(claims["aud"], expectedResource) && !tokenBoundToClient(claims, staticClientIDs) {
-		return uinfo, nil, fmt.Errorf("audience mismatch, got aud=%v azp=%v, want %q or a static client of %v",
-			claims["aud"], claims["azp"], expectedResource, staticClientIDs)
+		return uinfo, nil, fmt.Errorf("audience mismatch, got aud=%v azp=%v client_id=%v, want %q or a static client of %v",
+			claims["aud"], claims["azp"], claims["client_id"], expectedResource, staticClientIDs)
 	}
 
 	uinfo = p.parseUserInfo(claims)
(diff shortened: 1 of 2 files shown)


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1736.0.0-gceebd32

PR #1752, opened 2026-08-27
docs(controlplane): bootstrap folder and component rules

p3rotto:
## 📝 Description

Bootstrap task for the Control Plane initiative: creates `controlplane/` with the rules and the wire contract, so the four MVP component workstreams can start in parallel instead of each inventing its own shape.

Docs only. No Go, no `go.mod`, no `go.work` entry, deliberately: the control plane backend base is landing separately and this should not collide with it.

## 📣 User-facing impact

None. Internal groundwork.

## 🔗 Related Issue

Control Plane initiative, Foundation project.

## 🚀 Type of Change

- [x] 📚 Documentation update

## 📋 Changes Made

Two rule files, not seven. This started as one per component and collapsed, because that was a lot of prescription for a product with no code yet. Split again when there is code to split.

- `controlplane/CLAUDE.md` governs the backend the way `hoopinspect/CLAUDE.md` governs its own tree. Five non-negotiables, the decided/open tables from the 26 Aug architecture session, the sidecar wire contract, and a scope plus done-criteria section per component (desired state, inventory, admin auth, sidecar auth). **The wire contract is marked a proposal needing sign-off**, since every component crosses it.
- `controlplane/frontend/CLAUDE.md` governs the admin UI. Short on purpose: nothing is built and the stack is not chosen, so it records what is already settled (admins only, desired vs actual shown as two different things) and what is still open.
- `controlplane/README.md`: entry point and how the pieces fit.

## 🧪 Testing

No code, so nothing to run. Verified by inspection:

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: macOS

### Tests performed:
- [x] Every repo path referenced in the docs exists (`gateway/idp/`, `gateway/cmd/spiffe-mint/`, `common/dsnkeys/`, `common/license`, `hoopinspect/config/yaml`, `store/sqlite`, `pii/alcatraz`, `scripts/local-spiffe-*.sh`)
- [x] Claims about `sidecar.Config` checked against `hoopinspect/sidecar/config.go` (JSON-native, exhaustive validation)
- [x] House style check: no em dashes

## 📄 Additional Notes

Three things reviewers should push back on if they disagree, because they are written as rules and will be followed:

1. **The wire contract is a proposal.** Full config documents rather than deltas, mandatory NACK, `hello` gates everything. Sign it off or change this file first.
2. **Rule 3, never fall back to the bootstrap file after first sync.** The meeting notes say "se não conectado: sidecar segue config local (YAML)". Written as-is that is a security regression: reverting to the file on disconnect silently discards every rule the control plane pushed, so cutting the network loosens enforcement. This file says cache the last known control plane config instead. Worth an explicit yes/no.
3. **Rule 4, WebSocket must not break the sidecar's zero-dependency invariant.** Go has no stdlib WebSocket, and `hoopinspect`'s root module is stdlib-only by enforced test. The client belongs in a nested module under `hoopinspect/`, matching the existing pattern. Fallback if that gets awkward: SSE down plus POST up, both pure `net/http`.

Also recorded as open, not decided: whether `controlplane/` becomes its own Go module, the license reissue plan for existing customers, and who starts sidecars (currently documented as adopt-only).


sandromello:
✅ Build Completed with Success, Version=1752.0.0-g9c567dd

p3rotto:
Closing. Rogerio owns EVL-230 and will land this with the control plane backend base. Branch worktree-controlplane-bootstrap is kept, not deleted.

PR #1679, opened 2026-08-06
feat(webapp_v2): migrate AI Session Analyzer page to React (EVL-148)

luanlorenzo:
## 📝 Description

Migrates `/features/ai-session-analyzer` from ClojureScript to React (`webapp_v2`). Roadmap row **B3.2**.

Legacy URLs are kept, so sidebar links and activation-journey deep links keep working.

## 📣 User-facing impact

AI Session Analyzer settings now use the new interface. The resource filter on the rules list, which did nothing before, works.

## 🔗 Related Issue

EVL-148

## 🚀 Type of Change

- [x] ♻️ Code refactor
- [x] 🎨 Style/UI update
- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

## 📋 Changes Made

- New page `webapp_v2/src/pages/Features/AiSessionAnalyzer/` — Rules tab, Configure tab, create/edit form, store, ported template catalog
- Three routes in `Router.jsx` above the CLJS catch-all, `adminOnly` + `licenseFeature="ai-session-analyzer"`
- New `components/ConnectionNamesMultiSelect`; `SelectionCard` gains optional `image` and `badge` props; `usePaginatedConnections` also returns `items`
- Extended `services/aiSessionAnalyzer.js`; every interpolated rule name is `encodeURIComponent`d
- Docs: `MIGRATION_ROADMAP.md`, `CONTEXT_MIGRATION.md`, `COMPONENTS.md`

### Three things a reviewer should know

1. **Name is read-only when editing.** The gateway never writes the `name` column (`gateway/models/ai.go:238`), so the CLJS rename silently 404'd. Renaming needs a gateway change and its own ticket.
2. **The resource filter now filters.** `rule_list.cljs:28` rendered the unfiltered list.
3. **A 404 from `GET /providers` means "never configured"**, so the store settles as `success` with `provider: null`.

### CLJS side untouched

The CLJS routes stay in place, shadowed by the React routes that match first — same as #1650 and #1666. Deleting them would break `bidi/path-for` lookups in the sidebar, activation journey and promotion, which fail silently. `CONTEXT_MIGRATION.md` lists what Track A may delete and what must stay.

### Test Configuration
- **Browser(s)**: Chrome 131
- **OS**: macOS 15
- **Setup**: `cd webapp_v2 && npm run dev:full` against a local gateway

### How to test

Admin account, license with `ai-session-analyzer`. Go to `/features/ai-session-analyzer`.

[...]

PR #1754, opened 2026-08-27
feat(controlplane): the control plane frontend

rogefm:
## 📝 Description

The frontend for the **Control Plane** initiative, as a new app under
`controlplane/frontend`. It is the Navigation Routing project's deliverable: the
information architecture, settled ahead of the gateway conversion so that Reviews,
Session Analyzer, Guardrails and Data Masking each have a defined place to land instead
of inventing their own surface.

Nothing ships from this PR. There is no Makefile target, no build integration and no
backend — the dev server proxies `/api` to the existing gateway on `:8009`.
`webapp_v2` is untouched.

## 📣 User-facing impact

None. Not built, not served, not reachable by any customer.

## 🔗 Related Issue

Fixes EVL-230

## 🚀 Type of Change

## 📋 Changes Made

- **`controlplane/frontend/`** — 179 source files, the dependency closure of the pages
  the control plane keeps, extracted from `webapp_v2` (371 files).
- **One route map, no catch-all.** Sidecars and Reviews are placeholders that name the
  Linear project that owes them; approval rules and Slack move under `/reviews/*`;
  Session Analyzer, Guardrails and Data Masking are functional; unmatched paths 404.
- **One theme.** `src/theme.js` keeps hoop's ramps, radius scale, headings and the 40px
  control scale, and drops the layer that existed to match the ClojureScript app's
  Radix look.
- **`scripts/check-assets.mjs`** — fails the build when a referenced image is not in
  `public/`.

### Test Configuration
- **Browser(s)**: Chrome 151 (headless, via CDP)
- **OS**: macOS 15

### Tests performed
- [ ] Unit tests pass — the app has no test suite yet, same as `webapp_v2`

```bash
make run-dev                            # gateway on :8009
cd controlplane/frontend && npm install && npm run dev
```

Verified:

- `npm run build` and `npm run lint` green. Lint reports 13 problems, all in code copied
  verbatim plus `vite.config.js`; none in a file written for this PR. They differ from
  `webapp_v2`'s count only because a fresh install resolves
  `eslint-plugin-react-hooks` 7.1.1 against 7.0.1 in its lockfile, and 7.1.1 added the
  setState-in-effect rule.
- `/dashboard`, `/agents`, `/rulepacks`, `/settings/*`, `/client`, `/sessions`,
  `/resources` and the old `/features/access-request` all render the 404 page.
  `/login` renders; protected routes redirect to it without a session.
- Every theme variable resolves, including `--brand-navy` and `--hoop-surface-raised`
  (an undefined one silently invalidates the declaration — `AuthPageLoader` would be
  white text on white).
- Zero files unreachable from `main.jsx`; zero code references to ClojureScript.
- `git diff main -- webapp_v2/` is empty.

**Not verified: the signed-in pages.** Login is Auth0 and cannot be driven headlessly.
A reviewer should open `/reviews/rules`, `/guardrails` and `/features/ai-session-analyzer`
and confirm they work against the gateway — those three are what make V1 more than a
skeleton.

## 📄 Additional Notes

Worth a reviewer's attention:

- **The duplication is deliberate.** `webapp_v2` is frozen and diverges from here on. A
  fix to a shared component has to be made twice while both live. That is only
  acceptable while `webapp_v2` is in end-of-life; if the gateway is not retired, the
  answer is a shared package, not letting the copies drift. An ADR recording this comes
  with a later PR in the initiative.
- **A contradiction to settle in Linear, not here.** Navigation Routing lists "No
  dedicated Sessions section" under *Confirmed*, but the Reviews project says "sessions
  narrowed to Reviews" and treats it as product direction. This PR follows the latter:
  one surface, containing only review sessions, with the review detail inside the
  session. The *Confirmed* line needs editing before someone builds on it.
- **~2.7k lines of rule templates have no UI to reach them.** They were driven by the
  gateway's activation-journey deep link, which does not exist here. Kept, and labelled
  as such in the files, because Feature Configuration Across the Fleet asks for exactly
  this data — deleting curated security rules a named backlog project needs would be
  destroying work.
- **`public/` is 4.9 MB**, almost all of it four promotion PNGs (`guardrails` alone is
  2.2 MB). Same files `webapp_v2` serves today, so not a regression, but cheap to fix in
  a new app.


luanlorenzo (approved):

sandromello:
✅ Build Completed with Success, Version=1754.0.0-g6e6f383

rogefm:
## Qodo review — triage

Seven of the ten are fixed in `16951b5`. Splitting them by cause, because it changes who
should act:

### Caused by the extraction — my bugs, all fixed

Removing the gateway's routes left code still navigating to them. These are the ones
that mattered:

| # | Finding | Fix |
|---|---|---|
| 1 | Auth lands on 404 | Login (×2), Register (×2), Signup and Callback all went to `/client`. Now `/`, which routes an admin to Sidecars. |
| 2 | Onboarding routes absent | The zero-connection gate in `ProtectedRoute` is **removed**, not repointed — it existed to push an admin into a wizard this app does not have. A fresh org now lands on `/sidecars`, whose empty state already says the first step is connecting a sidecar. First-admin setup goes to `/` too. |
| 5 | Resource configuration opens 404 | The per-resource **Configure** button in Data Masking is gone; `RuleListItem` renders it only when given a destination. |
| 7 | License recovery opens 404 | Expired and unverifiable licenses now route to support, the only remediation that exists here. Dropped the `update` action and the banner's `/settings/license` suppression. |

**Qodo caught a class my own check could not see.** I had a script asserting every
sidebar and command-palette destination resolves — it never looked at `navigate()` in
pages, and its route matcher included the `*` route, so `^.*$` claimed every path and it
reported clean while eight targets were dead. Both are fixed, and
`scripts/check-routes.mjs` now runs in `npm run build`. Its limit is documented: it only
sees literal paths, so #2 and #7 — held in `setRedirectTo(...)` and a `LICENSE_PAGE`
constant — would still have needed a human.

### Pre-existing in `webapp_v2` — fixed here, still open there

| # | Finding | Note |
|---|---|---|
| 3 | Predictable reused passwords | Real. 512 possibilities from `Math.random`, and `useState` in a modal that never unmounts meant every user invited before a reload got the *same* one. Now four words from a 44-word list plus four digits, from `crypto.getRandomValues`, with the modal remounted per open. Weighs more here than in the gateway: inviting admins is the control plane's own flow. |
| 6 | Logout retains previous identity | Real. `useUserStore.clear()` existed with **zero callers**, so `logout()` left the user populated and `ProtectedRoute` skipped `/userinfo` on the next sign-in. Wired up. |
| 9 | Docs link misspelled | Real — `managing-accesss`, three s's. |
| 8 | Data Masking docs route | Accepted. `/features/data-masking` resolves to a real page; the old path serves the unresolved shell. |

`webapp_v2` still has #3, #6, #8 and #9. It is frozen for features, but these two
security ones are worth their own PR there — flagging rather than silently diverging.

### Not taking

| # | Finding | Why |
|---|---|---|
| 4, 10 | Promotion localStorage keys unscoped | Real, and pre-existing. Qodo notes the identical finding was **rejected in #1679**, the PR that shipped the analyzer page. Fixing it here would diverge from `webapp_v2` on a decision the team already made once. If the rule should hold, it should change in both, in its own PR. |

### One thing I checked and then distrusted

Verifying #8 and #9 I swept all 50 docs URLs the app links to, and 14 came back titled
"Overview". I nearly changed all of them. They are not comparable: the typo URL and a
legitimately-named overview page return **byte-identical** responses, because the docs
site is a Mintlify SPA and curl does not run its JS. Only the two above have evidence
behind them. Someone with the documentation repo should check the other twelve —
`/setup/deployment` in particular returns a real 404 with a canonical pointing at
`/setup/deployment/overview`.

sandromello:
✅ Build Completed with Success, Version=1754.0.0-g16951b5

PR #1759, opened 2026-08-27
EVL-230: bootstrap controlplane/backend

p3rotto:
## 📝 Description

Bootstraps `controlplane/backend` as an API scaffold: a Go module that serves `GET /healthz` and nothing else.

EVL-230 was closed when the frontend landed (#1754). The backend half was never merged, so no Go code existed. This is that half.

An earlier revision of this PR carried GORM, Postgres, migrations and four component packages answering `501`. Review said those encoded decisions the team has not made yet, so they are gone. What is left is the module, the build target and the docs. Every open question is listed as TBD rather than answered in code.

## 📣 User-facing impact

None. The binary serves a health check, is not wired into a release, and no existing behaviour changes.

## 🔗 Related Issue

EVL-230

## 🚀 Type of Change

- [x] 🔧 Build configuration change
- [x] 📚 Documentation update
- [x] 🧹 Chore

## 📋 Changes Made

- `controlplane/backend/`: new Go module `github.com/hoophq/hoop/controlplane/backend`, added to `go.work`. Seven Go files.
- One direct dependency, Gin 1.12.0. No ORM, no database driver, no migration tool. Everything else is stdlib.
- `internal/api`: Gin engine, `http.Server` with all four timeouts, graceful drain on SIGTERM, one structured log line per request, and `GET /healthz`.
- `internal/config`: four environment variables, all with defaults. `internal/logging`: an `*slog.Logger` constructor.
- `controlplane/README.md` and `controlplane/backend/CLAUDE.md`: written from the published product docs at hoop.dev, with a TBD list for everything those docs do not settle.
- `Makefile`: `build-controlplane` writes to `dist/release-binaries/`, with its own `LDFLAGS` because this binary does not import `common/version`.
- `go.work` and `Dockerfile.localbuild`: Go 1.27.0. See Additional Notes.
- `sidecar/CLAUDE.md`: corrects a claim that CI does not reach the sidecar tests. It does. Main's own Makefile comment already says so, this only makes the two agree.

### Test Configuration:
- **Browser(s)**: n/a, no UI in this PR
- **OS**: macOS 15 (Darwin 25.5.0), Go 1.27.0

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass

Integration tests are unchecked because there is nothing to integrate with: no database, no external service, no second endpoint.

`cd controlplane/backend && go vet ./... && go test ./...` is green and `gofmt -l .` is empty. CI reaches these through `make test-oss`, since `go test github.com/hoophq/hoop/...` matches the module now that it is in `go.work`.

Six unit tests. `internal/api`: `/healthz` returns 200 with the injected version, an unmatched path returns 404, and a third test fails if any route other than `GET /healthz` is registered, so the surface cannot grow untested. `internal/config`: defaults, overrides, and a table asserting that `soon`, `0s` and `-1s` are all rejected.

Manually, with the binary running:

- `curl -s localhost:8020/healthz` returns `{"status":"ok","version":"dev"}`.
- An unmatched path returns 404.
- The listen line, both request lines and the shutdown line are structured JSON on stderr. The request line carries the matched route, never the raw path, the query string or the body.
- `make build-controlplane` from the repository root produces `dist/release-binaries/controlplane-linux-amd64`, and `-X main.version` still resolves to the `version` variable in `main.go`.

n/a.

## ✅ Checklist

## 📄 Additional Notes

**Two things worth a reviewer's attention.**

1. **The published docs describe the control plane as the existing gateway.** `/docs/control-plane/deployment/kubernetes` and `/docs/control-plane/deployment/docker-compose` deploy image `hoophq/hoop`, service `hoopgateway`, HTTP on 8009, reading `API_URL` and `POSTGRES_DB_URI` with state in the `private` schema. That is not a separate Go binary. This PR does not resolve the contradiction. It documents both and makes "whether this module is what the published deployments run" the first TBD in each file. It probably needs deciding before the next line of code lands here. The scaffold binds 8020 only so it does not collide with a gateway running locally.

2. **This bumps the whole repository to Go 1.27.0.** `go.work` goes from 1.26.5 to 1.27.0 because the new module declares `go 1.27.0`, and every CI job reads `go-version-file: go.work`. So gateway, agent, client, common, sidecar and tunnel all build and test on Go 1.27 from this merge onward. Their own `go.mod` files still declare 1.26.5 and are unchanged. `Dockerfile.localbuild` moves to `golang:1.27-bookworm` for the same reason: it copies `go.work` into the build stage, and a 1.26 image would download a toolchain on every build or fail where that is disabled. The envoy-stack `hoop-inspect` image builds from the `sidecar/` context, which has no `go.work`, so its `golang:1.26-alpine` stage is untouched. If a repo-wide bump is unwelcome in this PR, the new module can declare 1.26.5 instead and nothing else here changes. Say the word.

**What the docs already decide, and the code therefore must not contradict.** The sidecar dials out over ordinary HTTP and the control plane never dials in. Configuration is pull-based, picked up on the sidecar's ping. A per-sidecar token is the whole credential. A sidecar keeps serving with its loaded config when this service is down, so this service is not on the data path. Those are recorded in `controlplane/backend/CLAUDE.md` as constraints. Everything else, including route paths, the ping interval, the config document shape, token issuance and rotation, the datastore, packaging and HA, is TBD.

**Conventions in the code, all reversible.** An injected `*slog.Logger` with no wrapper and no package-level logger, because `common/log` sits in a separate module with 101 requires including go-git and k8s.io/api. Config as a value passed from `main`, no global. `gin.New()` rather than `gin.Default()`. `SetTrustedProxies(nil)`, since Gin's default trusts any `X-Forwarded-For`. No `NoRoute` handler, because this binary serves no UI. Liveness carries no dependency check, so a dependency blip cannot restart every replica at once.

**No analytics events. No env var reaches an existing helm chart, and there is no chart for this service yet.** When one lands, the root `CLAUDE.md` rule applies and the chart changes in the same PR as the variable.

`skip-release`: nothing ships and no existing behaviour changes.


controlplane/backend/internal/config/config.go
```diff
@@ -0,0 +1,207 @@
...
+	ListenAddr string
+	// PostgresURI is the connection URI for the state store.
+	PostgresURI string
+	// MaxOpenConns caps the connection pool. Zero means unlimited.
+	MaxOpenConns int
+	// MigrationPathFiles reads migrations from disk instead of the embedded
+	// copy. Empty means embedded.
+	MigrationPathFiles string
+	// Deployment marks how seriously the process should treat itself.
+	Deployment DeploymentType
```
matheusfrancisco: Why not be the whole ConfigMap bounded by the deployment type?

config[deploymentType]

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g546e750

controlplane/README.md
```diff
@@ -0,0 +1,176 @@
...
+        ┌──────────┴───────────┐
+        │  control plane       │
+        │  (backend)           │
+        │                      │
+        │  desiredstate  what should run
+        │  inventory     what IS running
+        │  adminauth     who may change it
+        │  sidecarauth   who a sidecar is
+        └──────────┬───────────┘
+                   │ WebSocket, always dialled by the sidecar
```
matheusfrancisco: Will it be WebSocket?

controlplane/README.md
```diff
@@ -0,0 +1,176 @@
...
+      ┌────────────┼────────────┐
+      │            │            │
+ ┌────┴────┐  ┌────┴────┐  ┌────┴────┐
+ │ sidecar │  │ sidecar │  │ sidecar │   sidecar/, on customer infra
+ └────┬────┘  └────┬────┘  └────┬────┘
+      │            │            │
+   database     database     database    traffic never passes through us
```
matheusfrancisco:
```

 ┌─────────┐                                      ┌────────────────────────┐     ╭───────────────────────────╮
 │         │                                      │                        │     │                           │
 │         │                                      │                        │     │                           │
 │ sidecar ├◄┄┄┄┄┄┄┄┄┄┄GETTor11eonfigs┄┄┄┄┄┄┄┄┄┄┄►┤     control plane      ├────►│          Postgres         │
 │         │        If-NonevMatch0sgen-42         │ (N stateless replicas) │     │ desired state + inventory │
 │         │                                      │                        │     │                           │
 │         │                                      │                        │     │                           │
 └─────────┘                                      └────────────────────────┘     ╰───────────────────────────╯

 Recommended: versioned HTTP pull + status POST. Two endpoints, no framing, no affinity, no fan-out bus.```






 - GET /v1/desiredstate with If-None-Match: <generation> → 304 or a config document. This is daemon.LoadConfigBytes (config.go:331) verbatim on the
 sidecar side; it is already exported for exactly this ("config from somewhere other than a file — a ConfigMap, a secret manager",
 config.go:326-330).
 - POST /v1/status every 30 s carrying lane list, version, /stats counters → that is inventory, and it stays true across replicas because it is a
 row with a timestamp, not a socket.
 - Cost arithmetic: 10,000 sidecars at 30 s = 333 req/s of mostly-304s. A single replica handles that; N replicas need no coordination at all.
 Rollout latency is bounded by the poll interval, which is already faster than the restart the sidecar needs to apply anything.
 - Jitter the interval (interval * (0.8 + 0.4*rand)) and you have thundering-herd protection the existing fleet channel lacks.

File: controlplane/backend/internal/api/routes.go [hunk 4]
```diff
@@ -0,0 +1,114 @@
...
+	// the router. Test the handler directly until then. The alternative, a
+	// middleware that calls c.Next() so development is convenient, is a
+	// middleware that ships and leaves the config store unauthenticated.
+	protected := api.Group("")
+	protected.Use(admin.RequireAdmin)
+
+	protected.POST("/logout", admin.Logout)
+	protected.GET("/userinfo", admin.UserInfo)
+
+	// Desired state: what SHOULD run. Owned by EVL-231.
```
matheusfrancisco:
```
   POST   /api/auth/login|register|logout      GET /api/auth/me
   GET|POST   /api/sidecars                    GET|DELETE /api/sidecars/:id
   GET|PUT    /api/sidecars/:id/config    desired   EVL-231
   GET        /api/sidecars/:id/status    actual    EVL-232
   POST       /api/credentials            enroll    EVL-234
   POST       /api/credentials/rotate     rotate    EVL-234
   DELETE     /api/sidecars/:id/credentials  revoke EVL-234
 ```
maybe new api names without `-`

Comment (matheusfrancisco) [reply controlplane/backend/internal/api/routes.go]: Three naming systems in fourteen routes: bare verbs (login, userinfo), hyphenated resources (sidecar-configs), an abstract collective noun (fleet).

 - userinfo is the OIDC endpoint's name and there's no OIDC here → /api/auth/me
 - fleet is not a resource; /api/fleet/:name reads as an element of a fleet, which is a sidecar
 - sidecar-configs and fleet are the same object  same :name, same cardinality, same 404s. routes.go:102-105 says a shared prefix "would invite a
 handler that merges them", but a subresource isn't a merge, it's how spec and status stay apart. Kubernetes doesn't serve /podspecs and /podstatuses.
 - enroll/rotate/credentials: three routes, one resource, resource named in only the DELETE

File: controlplane/backend/internal/api/routes.go [hunk 5]
```diff
@@ -0,0 +1,114 @@
...
+// workstreams owns a prefix and no two collide, but nothing outside this
+// repository depends on them yet: controlplane/frontend still calls the
+// gateway on :8009 for everything. Renaming them is cheap right now and stops
+// being cheap the moment the frontend is repointed.
+func (s *Server) routes(engine *gin.Engine) {
+	health := newHealth(s.db, s.version)
+	admin := adminauth.New()
+	configs := desiredstate.New()
+	fleet := inventory.New()
+	sidecars := sidecarauth.New()
```
matheusfrancisco: I would prefer dependency injection so we can test more 

 `httpapi` constructs its own collaborators here. That's construction, not injection — it only looks harmless because all four `New()` are nullary
 today.

   EVL-231 needs a store, EVL-233 a session issuer, EVL-234 an `Anchor`. At that point this function either becomes `desiredstate.New(s.db, s.logger)`
 `httpapi` deciding how every other package is built — or every `New` signature change ripples into this file.

   Move the graph to `cmd/controlplane` and have `httpapi.New` take what's already built, as a struct rather than a fifth positional arg:

   ```go
   api := httpapi.New(httpapi.Deps{
      Config: cfg, Logger: logger, Version: version,
      Readiness: db,                     // interface{ Ping(ctx) error }, not *gorm.DB
      RequireAdmin: admin.RequireAdmin,  // gin.HandlerFunc
      DesiredState: desiredstate.New(db), Inventory: inventory.New(reg),
   })
   ```

   Two things fall out. `httpapi` stops depending on GORM  `s.db` is read once, at line 53, only to ping. And EVL-231/232 can finally test through
 the real middleware chain by passing a stub `RequireAdmin`, instead of the "test the handler directly until then" the comment on line 83 has to
 concede. `routes()` stays the single top-to-bottom answer to "what's exposed and what guards it"; it just stops also answering "how is it built".

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g49de62b

sandromello:
✅ Build Completed with Success, Version=1759.0.0-gd069ae1

controlplane/README.md
mtmr0x: good catch! No, it shouldn't, it's supposed to be a simple http pooling, no stateful connections for these ping/pong auth and state checks. Sidecar does POST/GET to Control Plane for auth and periodically for grabbing state when state is allocated in Control Plane

controlplane/backend/internal/api/routes.go
mtmr0x: isn't doing dependency injection with Golang the same of doing dependency injection with rocks? I remember that dependency injection in Go would make navigating to the modules literally impossible from inside another module. Say, you are in a controller like part of the app, wants to go to the logging thing under your cursor, Go-To-Definition just can't find it because it's a function parameter passed in memory from another module and the language server can't infer that.

controlplane/backend/internal/api/adminauth/adminauth.go
```diff
@@ -0,0 +1,114 @@
...
+// here. Read it, do not copy it.
+package adminauth
+
+import (
+	"github.com/gin-gonic/gin"
+
+	"github.com/hoophq/hoop/controlplane/backend/internal/api/apierr"
+)
+
+const ticket = "EVL-233"
```
rogefm: What's the purpose of this constant or the ticket parameter?

controlplane/backend/internal/api/routes.go
mtmr0x: or did they finally arrived at 2010 and fixed that?

controlplane/backend/internal/api/adminauth/adminauth.go
p3rotto: `Four component packages exist as stubs. Each answers 501 Not Implemented naming the ticket that owns it, so an unimplemented feature cannot be mistaken for a working one returning an empty list.`

Every endpoint on this bootstrap returns 501 not implemented, the ticket is just to give AI agents context later

EmanuelJr: Just slop generated, we are still cleaning up 🫠

controlplane/backend/internal/api/routes.go
matheusfrancisco: Partly right, but the diagnosis is aimed at the wrong thing. It's not DI that breaks navigation  it's interface-typed parameters, and those are separable.
1. Concrete injection navigates fine.

```golang
func NewController(log *slog.Logger, db *sql.DB) *Controller
```
2. Interface injection is what costs you the jump.

```golang
type Logger interface{ Info(string, ...any) }
func NewController(log Logger) *Controller

```
Now log.Info resolves to the interface method decl, because that's the only thing statically true. That's not a gopls limitation, it's the type system being honest at that call site there genuinely is no single implementation.


>Use go-to-implementation, not go-to-definition.
textDocument/implementation is the correct verb for this question. gopls indexes implicit satisfaction across the workspace, so it works  but it degrades exactly where you'd expect: many implementations (real + 3 mocks + a test fake) turn one jump into a picker, and implementations living outside your workspace modules may not show up. Go's lack of an implements keyword makes this structurally harder than in Java/C#, where the edge is written down.

controlplane/backend/internal/config/config.go
EmanuelJr: No, it's just used to set the Gin release mode.

p3rotto: Building on Emanuel's answer: `gin.SetMode` is the only live consumer today, plus the EVL-234 refusal. None of the nine fields vary by deployment, so I would rather not key the struct by it.

The real question under this is "what stops a production deploy running development-shaped settings?" I think that is a `Config.Validate()` refusing named dangerous combinations when the deployment is production. Not in this PR: there is nothing dangerous to refuse until EVL-234 adds the dev anchor, and an empty guard is worse than none.

controlplane/README.md
p3rotto: Confirmed, no socket. The docs were wrong rather than ahead of the code, so this was a paragraph fix.

controlplane/README.md
p3rotto: Accepted in full and implemented. `GET /v1/desiredstate` with `If-None-Match`, `POST /v1/status` every 30s, jittered.

The `LoadConfigBytes` citation settled it: the function is already exported for config from a ConfigMap or a secret manager, so the sidecar needs no new parsing path.

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g9385cdb

controlplane/backend/internal/api/routes.go
p3rotto: Accepted, full rename. `/api` for admins, `/v1` for sidecars, config and status as subresources of a sidecar.

Three deviations from your sketch:

- Enroll and rotate under `/v1`, not `/api/credentials`. A sidecar calls them, and enroll has its own guard because it presents a single-use bootstrap credential.
- No `POST /api/sidecars`. "Who starts sidecars" is still open; the standing answer is that we adopt sidecars and never create them.

controlplane/backend/internal/api/routes.go
p3rotto: Accepted, built as sketched. `cmd/controlplane` owns the graph and `api.New(api.Deps{...})` takes it. gorm is out of `internal/api`: `Readiness` is one method, `Ping`, satisfied by `database.Pinger`.

The concrete win is a test that did not exist. `newTestServer` passed `nil` for the db, so `/readyz` had no unit test at all. There is one for the 503 branch now.

controlplane/backend/internal/api/adminauth/adminauth.go
p3rotto: Removed. The four constants are gone and `apierr.NotImplemented` lost the parameter, so a 501 now reads `{"message": "reading a sidecar config is not implemented yet"}`. Ticket IDs stay in the package comments, which keeps the context for a later reader without putting a tracker ID in a public response body.

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g7e7064f

sandromello:
✅ Build Completed with Success, Version=1759.0.0-gb6f1502

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g26763a8

sandromello:
✅ Build Completed with Success, Version=1759.0.0-gdca6aa1

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g4ebae83

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g4830b1c

controlplane/backend/go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/hoophq/hoop/controlplane/backend
+
+go 1.27.0
```
sandromello, line 3: Could you make this aligned with the whole codebase GO version?
Use a separate PR to introduce bumping GO from all packages and CI as well.

go.work
```diff
@@ -1,11 +1,12 @@
-go 1.26.5
+go 1.27.0
 
-toolchain go1.26.5
+toolchain go1.27.0
```
sandromello, line 3: Don't do this now. Do this in a separate PR and change it for all the codebase (all modules, CI).

controlplane/backend/CLAUDE.md
```diff
@@ -0,0 +1,116 @@
...
+
+## What the docs already decide
+
+Constraints on anything built here.
+
+- The sidecar dials out over ordinary HTTP and the control plane never dials
+  in. No tunnel, no bidirectional connection held open, nothing pushed.
+- Configuration is pull-based: a sidecar picks it up on its ping.
+- A per-sidecar token is the whole credential. No second credential and no
+  certificate exchange. One token per sidecar, never shared between two.
```
sandromello, line 20: It doesn't make much sense when running in sidecar mode in any platform. Pods scale horizontally and you can even define different environment variables per replica and neither doesn't make sense because it's a replica of the same instance.

controlplane/backend/CLAUDE.md
```diff
@@ -0,0 +1,116 @@
...
+The module is listed in `go.work`, so `make test-oss` runs these tests too.
+
+## Configuration
+
+| Variable | Default | What |
+|---|---|---|
+| `CONTROLPLANE_LISTEN_ADDR` | `0.0.0.0:8020` | Bind address. The published deployments serve 8009, but the gateway owns 8009 on a development machine and both run there. Reconciling the two is TBD. |
+| `CONTROLPLANE_SHUTDOWN_GRACE` | `15s` | Drain time for in-flight requests on SIGTERM. |
+| `LOG_LEVEL` | `info` | `trace`, `debug`, `info`, `warn` or `error`. |
+| `LOG_FORMAT` | `json` | `json` or `text`. |
```
sandromello, line 75: An opportunity here:
I saw that are some envs that has HOOP_ prefix and some like this one does not.
We could take advantage of this bootstrap and add HOOP_ prefix to all envs landing on the control plane.

Don't have strong opinions about that, keep this as suggestion

sandromello (changes requested):

sandromello:
✅ Build Completed with Success, Version=1759.0.0-g3f5fe81

controlplane/backend/go.mod
p3rotto: LGTM, will do

go.work
p3rotto: LGTM, will do

controlplane/backend/CLAUDE.md
p3rotto: Derived from the docs: https://hoop.dev/docs/control-plane/connect-sidecar

I'll remove it from here, we can discuss it later and then change the docs as well

controlplane/backend/CLAUDE.md
p3rotto: LGTM, will do

controlplane/backend/CLAUDE.md
EmanuelJr: I think @mtmr0x may have some idea for this.

PR #1767, opened 2026-08-31
 DEP-178: revamp sidecar config into guardrails, opa and mask

matheusfrancisco:
> [!IMPORTANT]
> Label: **`major`**. Nothing is removed and every deployed config still loads, but three
> defaults change behaviour on upgrade with no config edit (see "Breaking behaviour" below).
> Downgrade to `minor` if the team reads a default flip as non-breaking.

## 📝 Description

Implements [ADR-0011](docs/adr/0011-sidecar-config-schema.md) and
[ADR-0012](docs/adr/0012-sidecar-enforcement-defaults.md).

One word, `policy`, named two products in the sidecar's `config.yaml`: Hoop's own rule set
and a client for someone else's Rego. This splits it into `guardrails` and `opa`, drops
three fields that restate what a list already says, and fixes two defaults that pointed the
wrong way for a component whose job is refusing statements.

Every pre-ADR-0011 config still loads. A single `normalize()` step folds the old spellings
onto the new ones between the strict decode and validation, so nothing downstream carries
compat logic, and each deprecated field prints a warning naming its replacement. Writing
one setting in both spellings is refused rather than resolved by precedence.

**Breaking behaviour on upgrade, no config edit required:**

1. A config that omitted `policy.enforce` ran observe-only and denied nothing. It now
   enforces. A rule set nobody validated because it never fired will start firing.
2. A config that omitted `audit.fail_closed` allowed a statement whose audit record could
   not be written. It now refuses. Audit-sink availability becomes traffic availability.
3. A config that omitted `pii` had detection off, so `mask.enabled: true` refused to start.
   It now starts and masks.

Configs that wrote any of those fields down keep their exact behaviour, including
`audit.fail_closed`, which inverts in value as well as name.

## 📣 User-facing impact

The sidecar config is smaller and says what it means: `guardrails` for Hoop's rules, `opa`
for the OPA integration, one entity list instead of two spellings, and no separate on/off
switches for masking and enforcement. Guardrails now enforce by default, and `mode: observe`
became a real dry run that records which rule *would* have refused each statement, so you
can size a rule set against live traffic for a week before turning it on. Omitting the `pii`
section enables detection for all 54 entity types instead of disabling it. Existing config
files keep working and print a warning naming each field to rename.

## 🔗 Related Issue

Fixes #

## 🚀 Type of Change

- [x] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

**Schema**

- `policy.rules` → `guardrails.rules`; `policy.opa` → a sibling `opa` block, top level and
  per listener.
- `policy.enforce` (`*bool`) → `guardrails.mode: enforce|observe`, defaulting to `enforce`.
  The empty string already means "inherit", so the pointer that carried three states is gone.
- `mask.rules[].entity` → `entities`, a list. One rule fans out to one rewrite per entity.
- `mask.enabled` removed: a non-empty rule list is the switch. The flag used to skip the
  whole validation block, so a lane could carry rules that could never fire and load clean.
- `listeners[].connection` removed. `name` fills the audit key and `input.context.connection`.
- `audit.fail_closed` → `audit.fail_open`, matching `opa.fail_open` and `analyzer.fail_open`.
  All three now read the same direction; one of them used to be negated.
- Omitting `pii` builds a detector over all 54 entity types instead of disabling detection.
  The section narrows rather than creates.
- Each inheritable section gained an opt-out: `guardrails: {rules: []}`, `opa: {}`,
  `mask: {rules: []}`. Two of the three had no way to say it before.

**Behaviour**

- `mode: observe` evaluates every rule, denies nothing, and records the rule that would have
  denied as `guardrails.would_deny` on the audit line. It previously evaluated nothing, which
  `docs/adr/0010` already flagged as "not a dry run". It now costs what enforcing costs.
- `action: defer` on a lane with no `opa.url` denies instead of refusing the config at
  startup. `defer` names a decision-maker; with none, the safe reading is refusal. One config
  file now serves a deployment with OPA and one without.
- `analyzer.fail_open` keeps its `true` default, deliberately. It guards a third-party
  network call on the request path, and an Anthropic incident should not be a database outage.

**Tooling**

- `--strict` / `-strict` on both entry points, exiting non-zero when a deprecated field is in
  use, so a pipeline can fail before the release that fails for it.
- Deprecation warnings go to stderr; `--validate` keeps stdout parseable.
- `GET /config` gains `observing`, `guardrails`, `opa_url` and `notes` beside the existing
  `enforcing`/`rules`/`opa`/`masking`, plus a `deprecated_fields` array in the payload so a
  control plane can warn its own developers.

**Rollout**

- Six shipped configs and two dev-script heredocs migrated.
  `envoy-stack/mssql2019/config-mssql2019.yaml` is deliberately left on the old spelling as
  the regression fixture for the deprecation path.
- `metabase-stack` had `name: appdb` with `connection: metabase`; the listener is renamed to
  `metabase` so its audit key keeps its value.
- `sidecar/README.md`, `sidecar-binary.md` and two stack READMEs rewritten, with a
  `### Deprecated fields` migration table placed before the first example.
- ADR-0006 marked superseded; ADR-0009 and ADR-0010 amended in place where this reverses them.

### Test Configuration:
- **Browser(s)**: N/A
- **OS**: macOS 15 (darwin 25.5.0, arm64), go1.26.5

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass

`make test-sidecar` at the repo root: **0 failures** across the root module and all five
nested ones. `go vet` clean everywhere. `go build ./client/...` clean.

`gofmt -l` reports nine files dirty; all nine were already dirty at `HEAD`
(`git show HEAD:<f> | gofmt -l /dev/stdin`), three of which this PR never opens. No new
drift, so they are left as the repo has them.

**Integration tests were not run.** The compose stacks
(`envoy-stack/`, `metabase-stack/demo.sh`) need Docker and were not exercised. Their configs
validate clean against a real binary, but the end-to-end masking assertion in
`metabase-stack/demo.sh` is worth running before merge.

**New tests, each failing without its change:**

| Area | Pins |
|---|---|
| `daemon/config_test.go` | enforce is the default; observe wraps rather than skips; observe over an empty chain stays nil; `defer` denies with no OPA and reports with one; the three opt-out spellings, before and after a real decode; `audit.failOnAuditError` is the inverse of `failOpen`; both-spellings-at-one-scope refused; every conflict reported in one run; the deprecated-window suite (policy block folds, `enforce: false` → observe, `connection` → name, `mask.enabled: false` keeps masking off, `fail_closed` inverts in four cases) |
| `gate/gate_test.go` | the dry run at the **audit row**: `allowed: true`, `kind: statement`, the rule name, `metadata["guardrails.would_deny"]`, and no violation record |
| `policy/` | `DenyDeferred` on both defer branches; `Observe` keeps `Rule`/`Message`/`Err`, forwards `EvalContext` to a contextual inner, and never writes into the inner annotation map; `ScopedScanner` receives the rule's entity list |
| `pii/alcatraz/` | empty `Entities` means every entity; `ignored` subtracts; `entities` fans out; `entity`+`entities` and `entities`+`columns` refused; `ScanTextFor` restricts the scan; `PluginFromConfig(nil)` returns a 54-entity plugin. `config.go` had no test file and now has one |

**Manual, against a real binary:**

```
$ hoop start sidecar --config <each of the 6 shipped configs> --validate
appdb      postgres  enforcing 2 rule(s) + masking
httpbin    http      enforcing 4 rule(s) + masking
metabase   postgres  enforcing 3 rule(s) + masking
mssqldb    mssql     enforcing 2 rule(s) + masking

$ hoop start sidecar --config <old spelling> --validate      # stderr, exit 0
warn: deprecated config: mssql2019: policy.rules is deprecated; rename it to guardrails.rules
warn: deprecated config: mssql2019: policy.enforce: true is deprecated; write guardrails.mode: enforce
warn: deprecated config: mssql2019: listeners[].connection is deprecated and duplicates name; drop it
warn: deprecated config: mssql2019: mask.enabled: true is deprecated; ...

$ hoop start sidecar --config <old spelling> --validate --strict     → exit 1

$ hoop-inspect -validate -config <both spellings of one field>
invalid config:
  - policy.enforce is deprecated and guardrails.mode is also set at this scope; keep guardrails.mode

$ hoop-inspect -validate -config observe.yaml
reporting  postgres  observing 1 rule(s), denying none
             note: observe mode: every rule is evaluated and nothing is denied.
                   Matches are recorded on the audit line as guardrails.would_deny

$ hoop-inspect -validate -config deferred.yaml       # action: defer, no opa block
appdb      postgres  enforcing 1 rule(s)
             note: rule(s) defer to a decision this lane has no opa.url for,
                   so a match denies instead of reporting a finding

$ hoop-inspect -validate -config optouts.yaml
inherits   postgres  enforcing 1 rule(s) + opa + masking
bare       postgres  no rules to enforce
```

N/A.

## ✅ Checklist

## 📄 Additional Notes

**Worth a reviewer's attention:**

1. **The `major` label.** Three defaults change behaviour for configs that omitted the
   corresponding field. Nothing is removed, so a case for `minor` exists. Your call.
2. **`guardrails: {rules: []}` means "opt out", not "add nothing".** For a field that
   concatenates, "adds nothing" is the other defensible reading. ADR-0012 promised the
   opt-out and the alternative leaves a lane unable to express it at all, so presence beats
   length in the merge. One `switch` in `resolve` if you disagree.
3. **`audit.fail_closed` → `fail_open` inverts polarity as well as name.** `normalize` reads
   the old field rather than pattern-matching the new one, so a config that wrote it down
   keeps its behaviour. Four test cases cover the matrix. This is the single riskiest line in
   the PR.
4. **`Validate()` on a hand-built `Config` does not understand deprecated fields.** They fold
   in `normalize`, which only `LoadConfigBytes` runs. Three tests moved because of it. Worth
   confirming you want that boundary.

**Corrections made while implementing:** ADR-0011 claimed a lane with mask rules on MSSQL
"loads clean today and will refuse to start" and called it the dangerous migration
direction. False, and verified so against a real build: `gate.MaskSupported` asks the codec
for a `Reframer` rather than listing protocol names, and all three shipped codecs have one.
The ADR was corrected and the false release-note hazard removed. Its conflict count was also
wrong (four fields have two writable spellings, not seven).

**Deprecation timeline is unset.** ADR-0011 proposes two minor releases warning, one
refusing with instructions, then removal. The releases need naming before merge, or the
deprecated fields follow `hoop start inspect` into permanent residency.

Both ADRs ship as `Status: Proposed` and per `docs/adr/README.md` want their own review
window before this merges.

### DIFF
--- a/client/cmd/startsidecar.go
+++ b/client/cmd/startsidecar.go
@@ -28,6 +28,7 @@ const deprecatedSidecarAlias = "inspect"
 var (
 	sidecarConfigFlag   string
 	sidecarValidateFlag bool
+	sidecarStrictFlag   bool
 )
 
 var startSidecarCmd = &cobra.Command{
@@ -50,9 +51,14 @@ does not require a different binary. The file may be YAML or JSON; the
 extension picks the parser.
 
 This command was named "inspect". That name still works as a deprecated
-alias.`,
+alias.
+
+The config schema changed in ADR-0011: "policy" split into "guardrails" and
+"opa", and three fields were dropped. Both spellings load, and the old one
+prints a warning naming its replacement. Use --strict to fail on one.`,
 	Example: `  hoop start sidecar --config /etc/hoop-inspect/config.yaml
-  hoop start sidecar --config config.yaml --validate`,
+  hoop start sidecar --config config.yaml --validate
+  hoop start sidecar --config config.yaml --validate --strict`,
 	// A bad config is not a usage error, and dumping the flag list under one
 	// buries the message that says which field is wrong.
 	SilenceUsage: true,
@@ -70,15 +76,20 @@ alias.`,
 			return err
 		}
 
+		// Same channel as the rename notice above, and for the same reason:
+		// --validate writes a parseable report to stdout.
+		daemon.ReportDeprecations(os.Stderr, cfg.Deprecations)
+		if sidecarStrictFlag && len(cfg.Deprecations) > 0 {
+			return fmt.Errorf("%d deprecated config field(s) in use and --strict is set",
+				len(cfg.Deprecations))
+		}
+
 		if sidecarValidateFlag {
 			lanes, err := daemon.Validate(cfg, det)
 			if err != nil {
 				return err
 			}
-			fmt.Println("config OK:", len(lanes), "listener(s)")
-			for _, ln := range lanes {
-				fmt.Printf("  %-16s %-9s %s\n", ln.Name, ln.Protocol, ln.Summary())
-			}
+			daemon.PrintLanes(os.Stdout, lanes)
 			return nil
 		}
 
@@ -117,8 +128,12 @@ func sidecarConfigFromEnv() string {
 // buildSidecarPlugin constructs the PII detector from the config's "pii"
 // section.
 //
-// A nil alcatraz.Plugin converts to a nil daemon.Plugin, so an absent section
-// stays nil rather than becoming a non-nil interface holding a nil pointer,
+// An absent section no longer disables detection: the plugin builds a
+// detector over every entity type it knows, and the section narrows it. A nil
+// Plugin means only that a build linked no detector, which this one does not.
+//
+// The conversion still matters. A nil alcatraz.Plugin converts to a nil
+// daemon.Plugin rather than to a non-nil interface holding a nil pointer,
 // which the sidecar would call through.
 func buildSidecarPlugin(raw json.RawMessage) (daemon.Plugin, error) {
 	return alcatraz.PluginFromConfig(raw)
@@ -134,6 +149,9 @@ func init() {
 		"Path to the inspection config file (YAML or JSON)")
 	startSidecarCmd.Flags().BoolVar(&sidecarValidateFlag, "validate", false,
 		"Validate the config, report what each listener resolved to, and exit")
+	startSidecarCmd.Flags().BoolVar(&sidecarStrictFlag, "strict", false,
+		"Fail when the config uses a deprecated field, so a pipeline can catch it "+
+			"before the release that removes it")
 
 	startCmd.AddCommand(startSidecarCmd)
 }
--- a/deploy/docker-compose/envoy-stack/README.md
+++ b/deploy/docker-compose/envoy-stack/README.md
@@ -91,10 +91,11 @@ and reuses it afterwards. After a library change:
 The image builds `hoop-inspect` with
 [alcatraz](https://github.com/hoophq/alcatraz) PII detection linked in, so the
 demo can mask Brazilian CPFs and IBANs and deny a query that embeds one. The
-`pii` section of `sidecar/config.yaml` names which of the 45 entity types
-are active, five here. The section is required rather than defaulted: turning
-on every recognizer rewrites ordinary numeric columns, because `US_SSN` fires
-on roughly a third of random nine-digit business ids.
+`pii` section of `sidecar/config.yaml` names which of the 54 entity types are
+active, five here. Omitting the section activates all 54, and this file
+narrows on purpose: a mask rule naming `US_SSN` rewrites ordinary numeric
+columns, because that recognizer fires on roughly a third of random
+nine-digit business ids.
 
 Sidecar knobs live in `sidecar/config.yaml`; validate a change without
 starting anything:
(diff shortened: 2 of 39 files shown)


sandromello:
✅ Build Completed with Success, Version=1767.0.0-gfc51bd8

sandromello:
✅ Build Completed with Success, Version=1767.0.0-g0f24b9e

sandromello:
✅ Build Completed with Success, Version=1767.0.0-gccb73e2

sandromello:
✅ Build Completed with Success, Version=1767.0.0-g00f15ff

sandromello (approved):

sandromello:
✅ Build Completed with Success, Version=1767.0.0-gae686dc

Issue #1764, opened 2026-08-30
[Bug]: MCP endpoint returns 415 for legal Content-Type spellings (charset parameter, letter case) — go-sdk pinned at v1.5.0


DudiPeretz-Orca:
### Bug Description

The MCP endpoint (`POST /api/mcp`) rejects `Content-Type` values that are legal spellings of `application/json`, returning **HTTP 415 `Content-Type must be 'application/json'`**.

Two forms are refused that should be accepted per [RFC 9110 §8.3](https://www.rfc-editor.org/rfc/rfc9110#section-8.3):

- `application/json; charset=utf-8` — parameters are explicitly permitted
- `Application/JSON` — media types are case-insensitive

This is not hoop's own code: the check comes from `modelcontextprotocol/go-sdk`, which `gateway/go.mod` pins at **v1.5.0**. Upstream fixed it in **v1.6.0** ([PR #878](https://github.com/modelcontextprotocol/go-sdk/pull/878), "mcp: allow Content-Type parameters in streamable transport").

**Why it matters in practice:** Apache HttpClient (Java) appends `charset=utf-8` to JSON bodies by default, so *any* Java-based MCP client is refused. We hit it with AWS Bedrock AgentCore, but AWS is incidental — it just happens to use that library. AgentCore sends a capability probe first (bare `application/json`, accepted), then falls back to a standard `initialize` carrying the charset parameter — and that fallback is refused, so MCP server registration never completes.

### Steps to Reproduce

Any valid hoop API key against a hoop gateway (`initialize` is a read-only handshake):

```bash
export HOOP_API_KEY='<key>'
URL=https://<your-hoop-host>/api/mcp
BODY='{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"probe","version":"0"}}}'

for CT in 'application/json' 'application/json; charset=utf-8' 'Application/JSON'; do
  printf '%-34s -> ' "$CT"
  curl -s -o /tmp/p.txt -w '%{http_code}  ' -X POST \
    -H "Authorization: Bearer $HOOP_API_KEY" \
    -H "Content-Type: $CT" \
    -H 'Accept: application/json, text/event-stream' \
    --data "$BODY" "$URL"
  head -c 80 /tmp/p.txt; echo
done
```

### Expected Behavior

All three succeed. The media type is `application/json` in every case; a `charset` parameter and letter case are not semantic differences.

[...]

PR #1768, opened 2026-09-01
fix: update dependencies and improve MCP server handler

EmanuelJr:
## What & Why

`POST /api/mcp` returned **415 Unsupported Media Type** for legal Content-Type spellings like `application/json;
charset=utf-8` and `Application/JSON` (#1764). `go-sdk v1.5.0` did a literal string compare on the header; Java MCP clients (Apache HttpClient — e.g. AWS Bedrock AgentCore) always send the charset parameter, so registration against the gateway failed outright.

- Bump `github.com/modelcontextprotocol/go-sdk` v1.5.0 → **v1.7.0** (`gateway/go.mod`, indirect pin in `client/go.mod`).   Upstream fixed the check in v1.6.0 via `mime.ParseMediaType`, so parameters and case variants are now accepted.
- Preserve cross-origin (CSRF) protection in `gateway/api/mcpserver/mcpserver.go`: v1.6.0 flipped the default so a nil `StreamableHTTPOptions.CrossOriginProtection` no longer enables Origin verification. The MCP handler is now wrapped with `http.NewCrossOriginProtection().Handler(...)` — the SDK-recommended replacement for the deprecated options field (removed in v1.8.0) — keeping the same semantics as v1.5.0: cross-origin non-safe requests are rejected with 403, requests without an Origin/`Sec-Fetch-Site` header (curl, server-to-server clients) pass.
- Add `gateway/api/mcpserver/transport_test.go`: table test posting the issue's raw `initialize` body with all three Content-Type spellings (expect 200), plus a `Sec-Fetch-Site: cross-site` → 403 guard so silently dropping the wrapper fails CI.

## Notes for Reviewers

- **Stateful server unchanged**: v1.7.0's new `2026-07-28` stateless protocol only activates with `Stateless: true`, which we leave unset — existing clients keep negotiating the legacy protocol (`2025-06-18` from the issue repro included). No protocol migration here.
- **New SDK default accepted**: `MaxRequestBodyBytes` now defaults to 4 MiB (previously unlimited). Left unset intentionally;  MCP payloads near 4 MiB aren't a supported use case. If one ever surfaces, set it explicitly in `StreamableHTTPOptions`.
- `MCPServer.handler` type widened from `*mcp.StreamableHTTPHandler` to `http.Handler` to hold the wrapped handler; `GinHandler` is unchanged.
- Everything else hoop uses from the SDK survives with identical signatures/semantics (`SessionTimeout` idle-only behavior, `AddTool`, progress notifications, tool annotations). Existing session-lifecycle tests  (`TestKeepAlive_ReproducesSessionTermination`, `TestProductionConfig_SurvivesIdleGap`) pass unmodified on v1.7.0.
- Wire-only serialization change: `ToolAnnotations` now emits `readOnlyHint`/`idempotentHint: false` instead of omitting them — no code impact.
- This wrapper is CSRF protection (server-side 403 on cross-site requests), distinct from and complementary to the permissive CORS headers set by `CORSMiddleware()` in `gateway/api/middleware.go`.

### DIFF
--- a/client/go.mod
+++ b/client/go.mod
@@ -135,7 +135,7 @@ require (
 	github.com/gomlx/go-xla v0.4.1 // indirect
 	github.com/gomlx/gomlx v0.28.2 // indirect
 	github.com/gomlx/onnx-gomlx v0.5.2 // indirect
-	github.com/google/jsonschema-go v0.4.2 // indirect
+	github.com/google/jsonschema-go v0.4.3 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.3.16 // indirect
 	github.com/googleapis/gax-go/v2 v2.22.0 // indirect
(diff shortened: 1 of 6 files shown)


p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1768.0.0-g59f14df

PR #1770, opened 2026-09-01
EVL-240: port the control plane API surface and narrow the frontend to it

rogefm:
## 📝 Description

#1773 landed the `APP_MODE=control-plane` skeleton: the config, the boot split in `Run()` that never starts gRPC or the proxies, a gRPC-free liveness probe, helm and `.env.sample`. Its control-plane surface is deliberately **one route**, `GET /api/healthz`.

This fills it in. `buildControlPlaneRoutes` now registers the **66 routes the shipped `controlplane/frontend` actually calls** — measured from its service files, then verified endpoint by endpoint against a running gateway. It also narrows the frontend to that surface, which is not optional: three pages break without it.

Registrations are copied from `buildRoutes` with their exact role, audit and analytics middleware, so no route gains or loses a guard by being ported.

> **Rebased.** This PR previously carried its own `APP_MODE` implementation with a route-allowlist mechanism, developed in parallel before #1773 merged. That half is gone — config, `appmode_test.go`, `serverinfo`, openapi and helm all duplicated what #1773 shipped, and its ADR (#1772) is withdrawn in favour of #1774. What survives is the part #1773 does not have.

## 📣 User-facing impact

None yet. The control plane is not pointed at anything in production; this makes its UI functional against `APP_MODE=control-plane`.

## 🔗 Related Issue

EVL-240 (this PR) · EVL-237 (#1773, the skeleton) · #1774 (ADR-0013, the design)

## 🚀 Type of Change

- [ ] 📚 Documentation update
- [ ] 🎨 Style/UI update
- [ ] ♻️ Code refactor
- [ ] ✅ Test update
- [ ] 🧹 Chore

## 📋 Changes Made

**Backend (3 files)**

- **`buildControlPlaneRoutes`** — 66 routes: auth and session bootstrap, users, feature flags, guardrails, data masking, AI session analyzer, reviews and review rules, sessions (read only), Slack via `/plugins`, connections, and a read-only `GET /users/groups`.
- **`controlplane_routes_test.go`** — the `want` list grows from 2 entries to 99 (66 routes plus the 33 implicit `HEAD`s). Their test is the review gate and stays that way.
- **`gateway/api/plugins/plugin.go`** — `/plugins` is one route for every plugin and the name arrives in the body, so a route list cannot express "Slack only". Filtered in the handler behind `appconfig.Get().IsControlPlane()`, including the list response.

**Frontend (20 files)** — the surface excludes `/attributes` and all group *writes*, and three pages break on that:

- **Administrators was fatal.** `Promise.all([users.list(), users.listGroups(), publicServerInfo()])` in one `try` — a 404 on groups rendered *"Failed to load users."*, killing the whole page. The group selector is gone; the form still sends each user's existing groups, so editing does not clear what a gateway operator set.
- **Attribute pickers, filters and store slices removed** from Guardrails, Data Masking and Review Rules. Review Rules Create was hard-erroring on `attributesStatus === 'error'`. Same preservation rule: each form carries the record's existing `attributes` through untouched.
- **Five dead service methods deleted** — all targeting blocked routes, none with a call site: `userGroups.create/remove`, `connections.testConnection`, the MCP OAuth pair, `mcpCatalog`.
- **`layout/ModeBanner`** — dev-only banner when `/serverinfo` reports `application_mode: "gateway"`, because developing against the full surface hides the gap until production.

### Test Configuration:
- **Browser(s)**: not exercised — see gaps
- **OS**: macOS 15.5, Go 1.26.5, Node 25.9.0

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass

`make test-oss` green: 99 packages with tests, 0 failures. Frontend `npm run build` clean; `npm run lint` reports 13 problems, **identical to `main`** — all pre-existing, none added.

Verified against a running gateway in control-plane mode (before the rebase, on the equivalent surface):

- 15 allowed routes → 200; 22 blocked → 404.
- `GET /users/groups` 200; `POST` and `DELETE /users/groups/:name` 404.
- `/plugins` lists `["slack"]` only, in an org that has others in the database; `dlp`, `review`, `audit`, `access_control`, `webhooks`, `runbooks` all 404 on read, config-write and create.
- All 39 endpoints the frontend can call are registered.

**Debugging note:** a blocked route and a handler answering "no such record" can return the same 404 body — `gateway/api/ai` uses `{"message":"resource not found"}` for a missing provider. Send the request **without a token** to tell them apart: a registered route answers 401 from the auth middleware, an unregistered one 404.

### Gaps — not verified

- **Browser.** Administrators rendering its list (the regression check for the group removal), the absence of attribute pickers, and ModeBanner appearing were not visually confirmed.
- **The `/plugins` narrowing has no test.** `appconfig` is a singleton with no test seam, so a test in `apiplugins` cannot flip the mode. Verified by hand only. This is an instance of the general gap below.
- **Re-verification after the rebase.** The route set is unchanged and the tests pass, but the manual curl pass above was run against the pre-rebase build.

## ✅ Checklist

## 📄 Additional Notes

**This ports 66 routes at once, and #1774 says routes should be ported one at a time.** That tension is deliberate and worth a decision in review. The argument for the batch: this is the measured set the shipped UI already calls, and any subset leaves the control plane's UI partly dead. The argument against is #1774's, and it is a fair one — no route here has been reviewed against a deployment with no agents and no sessions. **If you want them staged, the 66 are the checklist; say so and I will split it.** Two entries deserve scrutiny either way: `/reviews` and `/sessions` are gateway-session-shaped, and ADR-0009 says the sidecar's enforcement path is a different thing.

**The gap in #1774 to watch.** Its option-3 rejection does not consider middleware duplication. Every route ported here copies its role, audit and analytics chain out of `buildRoutes`. Adding `AuditMiddleware` to a route on one side and not the other is silent, and it is security drift rather than cosmetic. #1774 already anticipates converging on the allowlist, but its trigger is *"if the control plane ends up serving most of the gateway's routes"* — that threshold arrives after the drift is possible. Worth amending to "the first route whose middleware chain exists in two places", which this PR reaches immediately.

**Behaviour divergence has no enforcement, unlike route divergence.** The route list is pinned by a test; `/plugins` was found by reading code, not by a failing test, and there will be others. That asymmetry is the weakest part of the design and the strongest argument for finishing the separate module later.

**Two follow-ups, out of scope:**
1. **Approval deep links break.** Seven sites build `<apiURL>/sessions/<sid>` (Slack, Jira, webhooks, audit, access requests) and the control-plane router does not claim `/sessions/:id`, so every approval link lands on a 404.
2. **The served OpenAPI spec still lists all 264 routes** in control-plane mode — already recorded in #1774.


sandromello:
✅ Build Completed with Success, Version=1770.0.0-gfe59df4

sandromello:
✅ Build Completed with Success, Version=1770.0.0-g2c7bda0

sandromello:
✅ Build Completed with Success, Version=1770.0.0-gfe59df4

rogefm:
Holding this: #1773 merged the `APP_MODE` skeleton in parallel and #1774 records the design as ADR-0013. Rebase tracked in EVL-240 — the route surface, `Router` hardening, `/plugins` narrowing and frontend work all still apply; the config/helm/openapi half is now duplicated and comes out.

p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1770.0.0-g80903a0

rogefm:
Went through the Qodo review. Two of three are real and are fixed in d67791d; the third I'm declining with evidence.

### 2. Review approval calls nil callback — **confirmed, fixed**

Correct, and worse than it reads. `reviewapi.ReviewByIdOrSid` calls `h.TransportReleaseConnection` unconditionally once the status is Approved or Rejected (`review.go:173` — the only call site), and `runControlPlane` builds `&api.Api{TLSConfig: tlsConfig}` with no `ReleaseConnectionFn`. The panic lands *after* `DoReview` commits, so a recovered 500 hides a finalized review from the client.

`PUT /reviews/:id` is now off the surface. Not nil-guarded: there is nothing to release in this mode — a sidecar pulls its configuration over HTTP and holds no stream — and a sidecar review is a different entity from a gateway session review (ADR-0009), so approving belongs with that entity rather than with a guard on this one. Nothing calls it today; the Reviews pages are `NotImplemented`.

`GET /reviews` and `GET /reviews/:id` stay: read-only, and neither touches the callback. The handler is now constructed with a stub that logs a named error, so if someone re-adds the write route it fails diagnosably instead of nil-dereferencing.

### 3. Slack integration never starts — **confirmed, documented, not removed**

Also correct. `plugintypes.RegisteredPlugins` is populated at `main.go:219`, inside `runGateway`, so in this mode `processOnUpdatePluginPhase` iterates an empty slice and no Slack runtime starts.

I kept the routes and wrote the limitation into `buildControlPlaneRoutes` and `controlplane/frontend/CLAUDE.md`: they persist configuration and nothing else, the config is read when a runtime does start, and with `PUT /reviews/:id` gone there is no review decision to notify about anyway.

**This one is worth a second opinion.** Removing `/plugins` is also defensible and would drop a page from the sidebar. Say the word and I'll take it out.

### 1. ModeBanner bypasses the Text wrapper — **declining**

The rule is real, but the fix would make things less consistent. There is no `Text` wrapper in `src/components/` (39 wrappers, none for `Text`), and `layout/LicenseBanner` — the sibling component `ModeBanner` sits directly above, already shipped — imports `Text` from `@mantine/core` at line 2. Creating a wrapper here would make `ModeBanner` its only consumer while every other call site keeps the direct import.

Worth doing as its own change that introduces the wrapper and migrates all call sites. Not worth doing for one component in this PR.

EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1770.0.0-gd67791d

luanlorenzo (approved):

PR #1711, opened 2026-08-12
fix(rdp): apply connection data masking rules

racerxdl:
RDP live masking now uses the Data Masking rules assigned to the selected connection instead of gateway-wide entity and confidence environment settings.

- Resolves direct and attribute-based rules, sends the union of `supported_entity_types[].entity_types` plus the lowest applicable `score_threshold` to the RDP agent, and restricts Presidio to that allowlist.
- Stops using gateway-wide threshold and denylist settings for live masking without changing offline analysis; masked sessions require an allowlist-capable agent, legacy gateways remain compatible with upgraded agents, and unsupported custom entities fail closed.
- Verification: the focused Go packages pass, the Rust PII-gate suite passes 230 tests, and dedicated checks cover `DATE_TIME`/`PERSON` rule resolution, zero thresholds, legacy denylist compatibility, custom-rule rejection, and the Presidio request payload.

Automated by MisterMal

PR #1773, opened 2026-09-01
EVL-237 add control plane mode to the gateway

p3rotto:
## 📝 Description

Adds a run mode to the gateway binary. `APP_MODE=control-plane` starts the HTTP API only; `APP_MODE=gateway` (the default, and the value when the variable is unset) changes nothing.

The control plane administers a fleet of sidecars, so it carries no agent or client traffic. This PR is the boot path only: the API surface is ported route by route in follow-ups, so today the control plane answers `/api/healthz` and nothing else.

|  | gateway (default) | control-plane |
|---|---|---|
| Routes registered | 264 | 2 (`GET`/`HEAD /api/healthz`) |
| gRPC `:8010` | open | closed |
| Protocol proxies, transport plugins, agent controller | started | not started |
| Static UI + SPA fallback, `/.well-known/*`, `/ssm`, `/rdpproxy` | served | absent |
| Bootstrap (migrations, default org, auth) | runs | runs |

Builds on `bf40f968` from `feat/modes`, which introduced `APP_MODE` and the `application_mode` field of `/serverinfo`.

## 📣 User-facing impact

None. The new mode is opt-in and serves no product surface yet.

## 🔗 Related Issue

EVL-237

## 🚀 Type of Change

## 📋 Changes Made

- `gateway/appconfig`: `AppMode` type with `AppModeGateway` / `AppModeControlPlane`, `parseAppMode` (case-insensitive, empty means gateway), `AppMode()` and `IsControlPlane()` accessors.
- `gateway/main.go`: `Run()` keeps the shared bootstrap, then dispatches to `runControlPlane()` or `runGateway()`. Everything that carries agent and client traffic moved into `runGateway`, so a future addition there is off in control-plane mode by construction.
- `gateway/api/server.go`: `newEngine()` and `newAPIRouter()` extracted so both modes share one middleware chain; `buildControlPlaneEngine()` and `buildControlPlaneRoutes()` added.
- `gateway/api/healthz`: `ControlPlaneLivenessHandler()`.
- `/serverinfo` reports `application_mode`; OpenAPI regenerated.
- Gateway helm chart: `APP_MODE` pass-through and a documented default.

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: macOS 15 (darwin/arm64)

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

`make test-oss` green, zero failures.

New tests:
- `gateway/appconfig/appmode_test.go` — mode parsing, including that a zero-value config reads as the gateway.
- `gateway/api/controlplane_routes_test.go` — asserts the exact route list, so a route added to `buildControlPlaneRoutes` without being listed fails the test; and asserts `/api/healthz` answers 200 with no gRPC server running.

Manual, both modes booted against embedded PGlite:

```
# control plane
APP_MODE=control-plane POSTGRES_DB_URI=pglite:///tmp/cp hoop start gateway

GET /api/healthz     -> 200 {"liveness":"OK"}
GET /api/serverinfo  -> 404
GET /                -> 404
GET /.well-known/oauth-protected-resource -> 404
tcp 127.0.0.1:8010   -> closed
gin registered 2 routes

# gateway (regression)
POSTGRES_DB_URI=pglite:///tmp/gw hoop start gateway

GET /api/healthz          -> 200
GET /api/publicserverinfo -> 200
GET /api/serverinfo       -> 401 (unauthenticated)
tcp 127.0.0.1:8010        -> open
gin registered 264 routes
```

## ✅ Checklist

## 📄 Additional Notes

**Why a separate liveness handler.** The gateway's `LivenessHandler` dials `127.0.0.1:8010` and returns 400 when the port is closed. The control plane never opens it, so reusing that handler would fail every health check — and `/api/healthz` is what the helm service, the AWS load balancer template and the docker-compose healthcheck all probe. `ControlPlaneLivenessHandler` carries no swagger annotation on purpose: a second `@Router /healthz [get]` made swag drop the documented 400 response from the spec.

**Sentry moved up.** `monitoring.StartSentry` is now part of the shared bootstrap, so in gateway mode it initialises before transport-plugin startup instead of after. Both modes get it.

**`appconfig.go` diff noise.** Adding an accessor to the aligned single-line block made gofmt realign the whole group. The behaviour change there is the six lines around `parseAppMode`, the struct field and the two accessors.

**Helm guard on the bundled default agent.** `defaultAgent.enabled=true` with `config.APP_MODE=control-plane` renders an agent pointed at `127.0.0.1:8010`, which this mode never opens. The chart now fails rendering for that combination, unless `defaultAgent.grpcHost` names an external gateway.

**Reviewers:** the decision worth pushing back on is the split between shared bootstrap and `runGateway`. Migrations, the default org, default runbooks and rulepack seeding all still run in control-plane mode; `ProvisionOrgAgentKey` and the default connection tags are agent- and connection-shaped and arguably belong on the gateway side of the split.

No ADR yet. This picks between real alternatives and sets how every future route is gated per mode, so it may deserve one before the route ports start.

### DIFF
--- a/deploy/helm-chart/chart/gateway/templates/secret-configs.yaml
+++ b/deploy/helm-chart/chart/gateway/templates/secret-configs.yaml
@@ -4,6 +4,7 @@ metadata:
   name: hoop-config
 type: Opaque
 stringData:
+  APP_MODE: '{{ .Values.config.APP_MODE | default "gateway" }}'
   ORG_MULTI_TENANT: '{{ .Values.config.ORG_MULTI_TENANT | default "false" }}'
   MIGRATION_PATH_FILES: '{{ .Values.config.MIGRATION_PATH_FILES | default "" }}'
   STATIC_UI_PATH: '{{ .Values.config.STATIC_UI_PATH | default "/app/ui/public" }}'
--- a/deploy/helm-chart/chart/gateway/values.yaml
+++ b/deploy/helm-chart/chart/gateway/values.yaml
@@ -132,7 +132,7 @@ config:
   # Which component this deployment runs as: 'gateway' (default) serves
   # agents and clients; 'control-plane' administers a fleet of sidecars and
   # starts no gRPC transport and no protocol proxies.
-  # APP_MODE: 'gateway'
+  APP_MODE: 'gateway'
   # postgres://user:pass@host:5432/dbname — external PostgreSQL (recommended).
   # pglite:///path/to/data-dir — embedded PostgreSQL (single-node/standalone
   # only; requires a persistent volume mounted at the data directory).
--- a/deploy/helm-chart/chart/gateway/values.yaml
+++ b/deploy/helm-chart/chart/gateway/values.yaml
@@ -129,6 +129,10 @@ securityContext: {}
 
 # hoop gateway configuration. Please refer to https://hoop.dev/docs/self-hosting
 config:
+  # Which component this deployment runs as: 'gateway' (default) serves
+  # agents and clients; 'control-plane' administers a fleet of sidecars and
+  # starts no gRPC transport and no protocol proxies.
+  # APP_MODE: 'gateway'
   # postgres://user:pass@host:5432/dbname — external PostgreSQL (recommended).
   # pglite:///path/to/data-dir — embedded PostgreSQL (single-node/standalone
   # only; requires a persistent volume mounted at the data directory).
--- a/gateway/appconfig/appconfig.go
+++ b/gateway/appconfig/appconfig.go
@@ -20,6 +20,31 @@ import (
 
 // TODO: it should include all runtime configuration
 
+// AppMode selects which component the gateway binary runs as.
+type AppMode string
+
+const (
+	// AppModeGateway is the default: the full gateway, serving agents over
+	// gRPC, protocol proxies and the complete HTTP API.
+	AppModeGateway AppMode = "gateway"
+	// AppModeControlPlane runs the HTTP API only, to administer a fleet of
+	// sidecars. It serves no agents and no clients.
+	AppModeControlPlane AppMode = "control-plane"
+)
+
+// parseAppMode resolves the APP_MODE value. An empty value keeps the gateway
+// behaviour, so an existing deployment that never heard of APP_MODE is
+// unaffected.
+func parseAppMode(v string) (AppMode, error) {
+	switch mode := AppMode(strings.ToLower(strings.TrimSpace(v))); mode {
+	case "":
+		return AppModeGateway, nil
+	case AppModeGateway, AppModeControlPlane:
+		return mode, nil
+	default:
+		return "", fmt.Errorf("invalid APP_MODE %q (want %s|%s)", v, AppModeGateway, AppModeControlPlane)
+	}
+}
 
 type pgCredentials struct {
 	connectionString string
@@ -31,6 +56,7 @@ type pgCredentials struct {
 	pgliteDataDir string
 }
 type Config struct {
+	appMode                         AppMode
 	apiKey                          string
 	askAICredentials                *url.URL
 	authMethod                      idptypes.ProviderType
@@ -91,6 +117,10 @@ var runtimeConfig Config
 func Load() error {
 	if runtimeConfig.isLoaded {
 		return nil
+	}
+	appMode, err := parseAppMode(os.Getenv("APP_MODE"))
+	if err != nil {
+		return err
 	}
 	apiURL := os.Getenv("API_URL")
 	if apiURL == "" {
@@ -262,6 +292,7 @@ func Load() error {
 	}
 
 	runtimeConfig = Config{
+		appMode:                         appMode,
 		apiKey:                          os.Getenv("API_KEY"),
 		apiURL:                          fmt.Sprintf("%s://%s", apiRawURL.Scheme, apiRawURL.Host),
 		grpcURL:                         grpcURL,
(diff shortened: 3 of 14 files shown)


sandromello:
✅ Build Completed with Success, Version=1773.0.0-g067c317

EmanuelJr (approved):

rogefm (approved):

luanlorenzo (approved):

p3rotto:
## Qodo Fix Summary — Round 1

Reviewed all 5 findings. 3 fixed, 2 deferred. `make test-oss` green, zero failures.

### ✅ Fixed (3)

| Issue | Location | Change |
|---|---|---|
| **APP_MODE** lacks chart documentation | `deploy/helm-chart/chart/gateway/values.yaml` | Uncommented `APP_MODE: 'gateway'` so the chart carries an active default. |
| **.env.sample** omits **APP_MODE** | `.env.sample` | Added `APP_MODE=gateway` with the supported values documented above it. |
| Default agent has no transport | `deploy/helm-chart/chart/gateway/templates/deployment.yaml` | Helm now fails rendering when `defaultAgent.enabled=true` and `config.APP_MODE=control-plane` while `defaultAgent.grpcHost` is still the local `127.0.0.1:8010`. Pointing the bundled agent at an external gateway stays legal, per the "if connecting the agent to an external gateway is intended" caveat in the prompt. |

The helm guard was verified against a dockerized helm 3.16.2 in four combinations:

| `APP_MODE` | `defaultAgent.enabled` | `grpcHost` | Result |
|---|---|---|---|
| gateway | true | local | renders |
| control-plane | false | — | renders |
| control-plane | true | local | **fails** with the incompatibility message |
| control-plane | true | external | renders |

### ⏭️ Deferred (2)

**`runGateway` terminates the process** — `gateway/main.go:208`

This commit added **0** and removed **0** `log.Fatal` calls in `main.go` (31 insertions / 1 deletion); the split only re-scoped existing lines into a new function. `Run()` itself holds ~10 more `log.Fatal` calls that were not flagged only because they fall outside the diff. Converting `runGateway` alone leaves `main.go` half-converted with no behavior change, since the boundary still calls `log.Fatal`; converting `Run()` too is a startup-path refactor beyond EVL-237. Worth its own ticket.

**`ControlPlaneLivenessHandler` lacks Swagger annotations** — `gateway/api/healthz/healthz.go:34`

The omission is deliberate and documented in the handler comment. A second `@Router /healthz [get]` block was tried during implementation: swag treated it as a duplicate operation and dropped the documented `400` response from both `autogen/docs.go` and `openapiv3.json`. The prompt's own caveat, "ensure generation does not create conflicting duplicate operations for the shared route", is the constraint that makes the fix impossible as stated — one route cannot carry two `@Router` blocks. Qodo rates this `● Weak` and cites its own rejected precedent in #1711.

---
*Generated by Qodo PR Resolver skill*

deploy/helm-chart/chart/gateway/values.yaml
```diff
@@ -129,6 +129,10 @@ securityContext: {}
 
 # hoop gateway configuration. Please refer to https://hoop.dev/docs/self-hosting
 config:
+  # Which component this deployment runs as: 'gateway' (default) serves
+  # agents and clients; 'control-plane' administers a fleet of sidecars and
+  # starts no gRPC transport and no protocol proxies.
+  # APP_MODE: 'gateway'
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** — uncommented `APP_MODE: 'gateway'` so the chart carries an active default. Skipped the chart README: it is a bare `# Hoop` / `- TODO` placeholder documenting none of the other config keys.

gateway/appconfig/appconfig.go
```diff
@@ -92,6 +118,10 @@ func Load() error {
 	if runtimeConfig.isLoaded {
 		return nil
 	}
+	appMode, err := parseAppMode(os.Getenv("APP_MODE"))
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** — added `APP_MODE=gateway` to `.env.sample` with the supported values documented above it.

gateway/main.go
```diff
@@ -175,6 +176,36 @@ func Run() {
...
+
+	bootstrap.Ready(map[string]string{"Control Plane": appconfig.Get().ApiURL()})
+
+	a.StartAPI()
+}
+
+// runGateway serves the gateway: everything that carries agent and client
+// traffic — the transport plugins, the protocol proxies, the gRPC server —
+// plus the full HTTP API.
+func runGateway(tlsConfig *tls.Config, apiURL, defaultOrgID string, isOrgMultiTenant bool) {
```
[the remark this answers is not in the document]
p3rotto: ⏭️ **Deferred** — this commit added 0 and removed 0 `log.Fatal` calls in `main.go` (31 insertions / 1 deletion); the split only re-scoped existing lines. `Run()` holds ~10 more, unflagged only because they fall outside the diff. Converting `runGateway` alone leaves the file half-converted with no behavior change; converting `Run()` too is a startup-path refactor beyond EVL-237. Tracked separately.

deploy/helm-chart/chart/gateway/templates/secret-configs.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: hoop-config
 type: Opaque
 stringData:
+  APP_MODE: '{{ .Values.config.APP_MODE | default "gateway" }}'
```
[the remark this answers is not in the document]
p3rotto: ✅ **Fixed** — helm now fails rendering when `defaultAgent.enabled=true` and `config.APP_MODE=control-plane` while `grpcHost` is still the local `127.0.0.1:8010`. Pointing the bundled agent at an external gateway stays legal, per the caveat in the prompt. Verified in all four combinations with helm 3.16.2.

matheusfrancisco, line 7: What do you think about, instead of being an env var today, when we run the gateway, I think it is hoop start gateway 

https://github.com/hoophq/hoop/blob/main/client/cmd/start.go#L68

Why not have Hoop start the control plane, and have it start the gateway in control plane mode?

matheusfrancisco: hey qodo what are the tradeoffs in starting with env var or start with the command? 

tradeoffs base on the docker images/builds and kubernates templates?

deploy/helm-chart/chart/gateway/values.yaml
```diff
@@ -129,6 +129,10 @@ securityContext: {}
 
 # hoop gateway configuration. Please refer to https://hoop.dev/docs/self-hosting
 config:
+  # Which component this deployment runs as: 'gateway' (default) serves
+  # agents and clients; 'control-plane' administers a fleet of sidecars and
+  # starts no gRPC transport and no protocol proxies.
+  APP_MODE: 'gateway'
```
matheusfrancisco, line 135: qodo if this config is omitted in the config the default is gateway?
If I already have a Kubernetes config deployed and updated this will be the default?

gateway/main.go
```diff
@@ -175,6 +176,36 @@ func Run() {
...
+
+	bootstrap.Phase("Starting API")
+	apiStep := bootstrap.Step("HTTP API")
+	apiStep.OK(fmt.Sprintf("%s mode=%s", appconfig.Get().ApiURL(), appconfig.AppModeControlPlane))
+
+	bootstrap.Ready(map[string]string{"Control Plane": appconfig.Get().ApiURL()})
+
+	a.StartAPI()
+}
+
```
matheusfrancisco, line 204: Qodo, can you think of any implications of not running the gRPC alongside the api?  is there any middl in the api that needs the grpc be up? or something?

deploy/helm-chart/chart/gateway/values.yaml
matheusfrancisco: done

deploy/helm-chart/chart/gateway/templates/secret-configs.yaml
matheusfrancisco: okay :)

p3rotto: I don't think they are mutually exclusive; we can have both

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1773.0.0-g19fcc10

PR #1776, opened 2026-09-01
EVL-241 let admins dismiss the setup checklist

feraio:
## What

The sidebar setup checklist only goes away when all 9 sub-items are done. An admin who does not want it is stuck with it taking space on the UI (acting as an offender for old customers).

This adds an `X` to the card header. The dismiss is per user, in `localStorage`:

- **1 sub-item left** at dismiss time: permanent. The admin read the last item and said no. Nearly every complaint is an org stuck on `Set Protection Level`.
- **More than 1 left**: 7 days, then the checklist returns.

The two knobs are `PERMANENT_AT_OR_BELOW` and `SNOOZE_MS` in `webapp_v2/src/stores/useUIStore.js`.

## Notes

- The expired snooze is pruned once at module load, not in the selector. `Date.now()` inside a Zustand selector is an unstable snapshot, and the DEP-136 gate must stay pure and hook-free. A snooze therefore ends on the next page load, not mid-session.
- Corrupt or hand-edited JSON reads as "not dismissed" — it fails toward showing the checklist.
- **Known gap**: after a permanent dismiss there is no in-app way to bring the checklist back, short of clearing `localStorage`. `LicenseBanner` sets the same precedent, and every checklist destination stays reachable from the sidebar nav.

## How to test

As an **admin of an org with `onboarding_completed_at` NULL** — the checklist renders for nobody else.

1. `cd webapp_v2 && npm run dev:full`, open **:5173**.
2. Checklist visible in the expanded sidebar. Click `X` — it disappears.
3. Hard-reload (Cmd+Shift+R) — still gone.
4. DevTools: `localStorage.getItem('config-status-dismissed')`. With more than 1 task left, `until` is an epoch ~7 days out. With 1 left, `until` is `null`.
5. Force the expiry: `localStorage.setItem('config-status-dismissed', JSON.stringify({userId: '<your-id>', until: 1}))`, reload — the checklist is back.
6. Log in as a **different admin** — the checklist is visible for them. The dismiss is per user.
7. Keyboard: Tab reaches the `X` separately from the expand toggle; Enter dismisses.
8. Regression (DEP-136): while dismissed, no `focus` or `hoop:session-executed` listener is registered.

Both branches were checked against a local stack: with 8 tasks pending the stored `until` was 7 days out; with only `Set Protection Level` pending it was `null`.

`npm run lint` and `npm run build` are green — lint reports the same 15 errors + 1 warning as `main`.

### DIFF
--- a/webapp_v2/src/layout/Sidebar/ConfigStatus/ConfigStatus.module.css
+++ b/webapp_v2/src/layout/Sidebar/ConfigStatus/ConfigStatus.module.css
@@ -9,11 +9,24 @@
 
 /* ── Widget header (always visible) ────────────────────────── */
 
+.header {
+  display: flex;
+  align-items: center;
+  gap: var(--mantine-spacing-xsAlt);
+  width: 100%;
+}
+
 .headerBtn {
   display: flex;
   align-items: center;
   gap: var(--mantine-spacing-sm);
-  width: 100%;
+  flex: 1;
+  min-width: 0;
+}
+
+.closeBtn {
+  flex-shrink: 0;
+  color: var(--mantine-color-dimmed);
 }
 
 .headerLabel {
--- a/webapp_v2/src/layout/Sidebar/ConfigStatus/index.jsx
+++ b/webapp_v2/src/layout/Sidebar/ConfigStatus/index.jsx
@@ -25,11 +25,10 @@ export function ConfigStatus() {
   const completed = useConfigStatusStore((s) => s.completed && s.forUserId === userId)
   // Read through a selector, not useState(() => localStorage...): the gate must
   // stay effect-free, and a store read matches the three selectors above. The
-  // store already dropped an expired dismiss, so this is a plain comparison.
-  // userId is null before /userinfo resolves and the stored value is null when
-  // never dismissed — compare only once we actually have a user, or null === null
-  // hides the checklist from everyone.
-  const dismissed = useUIStore((s) => userId !== null && s.configStatusDismiss?.userId === userId)
+  // store already dropped expired entries, so presence of the key is the answer.
+  // The userId guard is for the window before /userinfo resolves: no key of the
+  // map may stand in for "no user yet".
+  const dismissed = useUIStore((s) => userId !== null && userId in s.configStatusDismiss)
 
   if (!isAdmin || !showSetupChecklist || completed || dismissed) return null
   return <ConfigStatusWidget />
--- a/webapp_v2/src/layout/Sidebar/ConfigStatus/index.jsx
+++ b/webapp_v2/src/layout/Sidebar/ConfigStatus/index.jsx
@@ -2,7 +2,8 @@ import { useEffect, useState } from 'react'
 import { useLocation, useNavigate } from 'react-router-dom'
 import { Box, Collapse, Text, UnstyledButton } from '@mantine/core'
 import { useClickOutside } from '@mantine/hooks'
-import { ChevronDown, ChevronUp } from 'lucide-react'
+import { ChevronDown, ChevronUp, X } from 'lucide-react'
+import ActionIcon from '@/components/ActionIcon'
 import RingProgress from '@/components/RingProgress'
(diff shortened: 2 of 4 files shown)


sandromello:
✅ Build Completed with Success, Version=1776.0.0-ged9165a

feraio:
Reviewed all six Qodo findings. Two are fixed in 410a8a12, four are declined with reasons below.

### Fixed

**1. Later users overwrite dismissals** — correct. `localStorage` now holds a map keyed by user id, so a second admin dismissing in the same browser no longer erases the first one's entry. The gate reads `userId in map`, which stays a stable selector value.

**5. Storage failure breaks dismiss** — correct. The write is wrapped; a blocked or full store degrades to an in-session dismiss and logs a warning instead of throwing out of the click handler.

### Declined

**2. Snooze outlives seven days** — intended, and documented in the code. Pruning happens at module load because `Date.now()` inside a Zustand selector is an unstable `getSnapshot`, and the DEP-136 gate must stay pure. The suggested fix, a timer for the remaining duration, means a 7-day `setTimeout` that browsers park or drop in a backgrounded tab, so it would not deliver the guarantee it appears to add. The real behaviour is: the checklist returns on the first page load after the deadline. For a 7-day snooze that is not a user-visible difference.

**3. Dismissal parser lacks tests** — `webapp_v2` has no test runner and no testing dependency; the gate is `npm run lint` plus `npm run build`. Adding Vitest to cover one parser is a separate decision about the package, not part of this change.

**4. Stale count chooses duration** — real but not worth the cost. The window is: complete the second-to-last task, return within the 15s TTL, and dismiss without expanding the card (expanding forces a refresh). The wrong outcome is a 7-day snooze where a permanent one was due — self-correcting, and the user can dismiss again. The proposed fix makes an instant button await a network round-trip and adds a failure path where a refresh error could produce an accidental permanent dismiss.

**6. `Box` bypasses local wrapper** — no `Box` wrapper exists in `src/components`, 57 files outside it already import Mantine primitives directly, and this file already used `Box` for the card root before this PR. Qodo scores its own relevance `Weak` and cites #1727, where the same finding was rejected. Introducing a wrapper is a repo-wide decision, not one to make inside this diff.

sandromello:
✅ Build Completed with Success, Version=1776.0.0-g410a8a1

luanlorenzo (commented):
@feraio the dismiss is browser-only right now since the flag stays in the `localstorage` It comes back on a new browser, a new device, or cleared site data.

Is this meant to be per user or per org? If org-wide, DEP-136 (#1709) already gives us the place to persist it: write a key into `orgs.onboarding_steps` and check it in `show_setup_checklist` (`gateway/api/user/user.go:478`). That makes the dismiss permanent and device-independent, and the frontend just reads one boolean.

feraio:
> @feraio the dismiss is browser-only right now since the flag stays in the `localstorage` It comes back on a new browser, a new device, or cleared site data.
> 
> Is this meant to be per user or per org? If org-wide, DEP-136 (#1709) already gives us the place to persist it: write a key into `orgs.onboarding_steps` and check it in `show_setup_checklist` (`gateway/api/user/user.go:478`). That makes the dismiss permanent and device-independent, and the frontend just reads one boolean.

@luanlorenzo per user, not per org. Two admins on the same browser each keep their own dismissal, this is why the idea was to do it as a browser-only fix.

Kept it client-side on purpose for this PR: the checklist is a UI nudge, and localStorage matches the blast radius. Agreed it does not survive a new device.

If we persist it, `orgs.onboarding_steps` is the wrong seam twice: it is org-scoped, and `LatchOrgOnboardingSteps` merges with the stored key winning, so a second dismiss cannot extend the 7-day snooze. The right precedent is `users.signup_origin` / `ShouldShowOriginSurvey` : per-user column, server-evaluated window, one boolean out.

Gateway work is on hold for Control Plane, so we could do that as a follow-up ticket.

luanlorenzo (approved):

PR #1781, opened 2026-09-02
feat: sidecars

EmanuelJr:
## What & Why

Adds a first-class `sidecar` entity to the gateway. An operator registers a sidecar, receives a token once, and the running sidecar fetches its
`daemon.Config` from the gateway over HTTP instead of loading a local YAML file.

Six endpoints under `/api`: create, list, get, delete (JWT), plus `handshake` and `configuration` (sidecar token). The config is built from the database
on every request and never stored — connections carrying the new `connections.sidecar_id` become the listeners, and each connection's guardrail and
data-masking rules become that listener's `guardrails` and `mask` sections. The persisted row holds identity only; the reported version and last-seen
time live in process memory, so a restart forgets them and the next handshake refills them.

Gateway side only. The sidecar daemon and its CLI are unchanged and still load YAML from disk.

## Notes for Reviewers

- Migration `000114` adds `private.sidecars` and `connections.sidecar_id` (`ON DELETE SET NULL`, so delete unassigns). Delete is hard: the token stops
working at once. No masked-key or status column.
- `SidecarID` is a **pointer** on `openapi.Connection` and `ConnectionPatch`. This is load-bearing: `PUT /connections/:name` binds the whole struct and
rebuilds the model, so a non-pointer would let the existing web app blank the assignment on every save. Absent means "leave as is", `""` unassigns.
- Auth is a separate `hoop-sidecar-token` header with an `hsc_` prefix, which keeps it out of the `hpk_` branch in `AuthMiddleware`. Those two routes
take no role middleware and no `AuthMiddleware`; the middleware installs an org-scoped context with no user.
- `handshake` and `configuration` are refused as sidecar names — they would shadow the static routes beside `/sidecars/:nameOrID`.
- Guardrail **output** rules and **custom** masking entity types are refused with 422, not silently dropped. The sidecar cannot express either, and
dropping them would be weaker than the operator configured.
- The emitted config is round-tripped through `daemon.LoadConfigBytes` before it is returned, so the gateway never hands out a config the sidecar would
refuse. `hoop start sidecar --validate` accepts the output.
- Routes are registered on the control plane as well as the gateway; `TestControlPlaneRoutes` is extended to match.
- Two new analytics constants, `EventCreateSidecar` and `EventDeleteSidecar`. None removed or renamed.
- Assignment is validated when it is made, not when the sidecar fetches: unknown sidecar and unservable connection type both return 422, and a request
that only changes the connection type is re-checked against any assignment it does not mention.
- Listener `listen` mirrors the upstream port, so the sidecar takes over the address the application already dials. A topology where listen and upstream
differ (the envoy stack) needs a per-connection listen port — follow-up.
- Runtime state is process-local. With more than one gateway replica a sidecar's version shows only on the replica it last reached.
- No wire-contract change and no new env vars, so no helm chart change.


sandromello:
✅ Build Completed with Success, Version=1781.0.0-ge88f109

sandromello:
✅ Build Completed with Success, Version=1781.0.0-g53b7bf5

sandromello:
✅ Build Completed with Success, Version=1781.0.0-gb59c0ce

sandromello:
✅ Build Completed with Success, Version=1781.0.0-ge1366d3

sandromello:
✅ Build Completed with Success, Version=1781.0.0-g19c0b86

gateway/services/sidecar.go
```diff
@@ -0,0 +1,398 @@
...
+	switch pb.ToConnectionType(connType, subType) {
+	case pb.ConnectionTypePostgres:
+		return string(inspect.Postgres), nil
+	case pb.ConnectionTypeMSSQL:
+		return string(inspect.MSSQL), nil
+	case pb.ConnectionTypeHttpProxy:
+		return string(inspect.HTTP), nil
+	}
+	return "", fmt.Errorf("connection type %q cannot be served by a sidecar; supported: postgres, mssql, httpproxy",
+		strings.TrimSpace(connType+"/"+subType))
```
matheusfrancisco, line 73: The sidecar protocols came from  https://github.com/hoophq/libhoop/blob/main/v2/codec/types/types.go#L32

matheusfrancisco: idk if we need to use the pb.* 🤔

gateway/services/sidecar.go
```diff
@@ -0,0 +1,398 @@
...
+var ErrSidecarNoConnections = errors.New("no connections are assigned to this sidecar")
+
+// ErrSidecarLookup wraps a database failure. The caller answers 422 for an
+// operator mistake and 500 for this: a sidecar that retries a 422 never
+// recovers, and the driver text must not reach it.
+var ErrSidecarLookup = errors.New("failed loading sidecar configuration")
+
+// SidecarProtocol maps a hoop connection type/subtype onto a sidecar codec.
+// It returns an error naming the connection type when no codec speaks it.
+func SidecarProtocol(connType, subType string) (string, error) {
```
matheusfrancisco, line 61: what is connType and subType in this case for sidecar?

gateway/services/sidecar.go
```diff
@@ -0,0 +1,398 @@
...
+		Admin:    daemon.AdminConfig{Listen: sidecarAdminListen},
+		LogLevel: sidecarLogLevel,
+	}
+
+	var piiEntities []string
+	seenEntity := map[string]bool{}
+	var piiThreshold *float64
+
+	for _, conn := range conns {
+		protocol, err := SidecarProtocol(conn.connType, conn.subType)
```
matheusfrancisco, line 150: sadly i kind understood :(

sandromello:
✅ Build Completed with Success, Version=1781.0.0-g9de5ad8

sandromello:
✅ Build Completed with Success, Version=1781.0.0-ga38a4c6

sandromello:
✅ Build Completed with Success, Version=1781.0.0-g50a61da

matheusfrancisco (approved):
it looks 👍🏽

p3rotto (approved):

p3rotto (commented):
it looks good, great job! 👍

PR #1791, opened 2026-09-04
feat: OPA

EmanuelJr:
## What & Why

Adds OPA (Open Policy Agent) configuration management to the gateway. An admin
registers an OPA decision endpoint and points a sidecar-fronted connection at
it; the sidecar consults it as its per-listener `opa` block, so wire statements
can be gated by an external policy engine.

- New `opa_configs` table plus a nullable `connections.opa_config_id` FK
  (`ON DELETE RESTRICT`), migration `000115`. Model in `gateway/models/opa_configs.go`.
- New API package `gateway/api/opaconfigs/` with CRUD under `/opa-configs`:
  mutations are `AdminOnly`, reads `AdminAndAuditor`. Requests are shape-validated
  only (no reachability probe — the gateway usually can't reach the OPA the
  sidecar talks to).
- Connection `Post`/`Put`/`Patch` accept `opa_config_id` (name or id) and
  resolve/validate the assignment; `absent = leave as is`, empty string
  unassigns. Column threaded through connection reads and `ToOpenApi`.
- `BuildSidecarConfig` emits a per-listener `OPA` block via one batched lookup
  for all assigned connections; a nil pointer leaves the block absent (no
  top-level default, so a listener block replaces rather than merges).
- Analytics events (`hoop-{create,update,delete}-opa-config`), new OpenAPI
  types (`OPAConfigRequest`/`OPAConfigResponse`, connection `opa_config_id`),
  regenerated autogen docs.

## Notes for Reviewers

- `gate: true` is refused for now: the gateway emits no `ai_analysis` rules, so
  a gate-before-analyzer config would make the sidecar reject the whole config.
- Delete is refused while any connection still points at the config (the
  `Connections` field names them) — losing enforcement silently is worse than a 409.
- `fail_open` defaults false: an OPA outage stops traffic rather than silently
  disabling enforcement. `timeout_sec: 0` falls back to the sidecar's 2s default.
- Config is rebuilt on every sidecar fetch (nothing cached), so an edited OPA
  config reaches the sidecar on its next fetch.
- No frontend changes here — API/backend only.

### DIFF
--- a/gateway/analytics/events.go
+++ b/gateway/analytics/events.go
@@ -49,6 +49,11 @@ const (
 	EventCreateSidecar = "hoop-create-sidecar"
 	EventDeleteSidecar = "hoop-delete-sidecar"
 
+	// opa
+	EventCreateOPAConfig = "hoop-create-opa-config"
+	EventUpdateOPAConfig = "hoop-update-opa-config"
+	EventDeleteOPAConfig = "hoop-delete-opa-config"
+
 	// plugins
 	EventCreatePlugin          = "hoop-create-plugin"
 	EventUpdatePlugin          = "hoop-update-plugin"
--- a/gateway/api/connections/connections.go
+++ b/gateway/api/connections/connections.go
@@ -187,9 +187,7 @@ func Put(c *gin.Context) {
 		c.JSON(http.StatusUnprocessableEntity, gin.H{"message": err.Error()})
 		return
 	}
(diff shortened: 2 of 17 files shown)


sandromello:
✅ Build Completed with Success, Version=1791.0.0-g8eed700

p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1791.0.0-g4664721

PR #1801, opened 2026-09-09
feat: sidecar config

EmanuelJr:
## What & Why

Sidecars had no stored configuration: the gateway answered `handshake` and `configuration` with three hardcoded fields and no listeners, so every connected sidecar refused to start.

- Migration `000116` adds `sidecars.configuration JSONB NOT NULL DEFAULT '{}'`.
- The column is typed as `sidecar/daemon.Config` above the database edge (Valuer/Scanner in `gateway/models/sidecars.go`).
- Set on create, replaced by the new `PUT /api/sidecars/:nameOrID`, returned on read, and served to the sidecar verbatim.
- `services.BuildSidecarConfig` and its three constants are deleted. The gateway injects no defaults; writes are decoded strictly (`DisallowUnknownFields`), so a misspelled key is a 422 instead of a silently disabled control.
- New event `hoop-update-sidecar`; OpenAPI artifacts regenerated.

The branch also carries the earlier removal of `opa_configs` and the per-connection `opa_config_id` / `sidecar_id` assignments, which is what left the served config empty.

## Notes for Reviewers

- Two column lists exist in `gateway/models/sidecars.go`: `sidecarColumns` and the inline SELECT in `GetSidecarByKeyHash`. The second one is the row the sidecar receives.
- `daemon.Validate` is NOT run on write: it resolves TLS keypairs and license paths on the sidecar host, not on the gateway. An empty document stays storable, and the sidecar refuses it at startup for having no listeners.
- `PUT` replaces the whole document. There is no per-key patch, so a caller that changes one key sends the full config.
- Migration `000116` is amended, not renumbered. If it lands on `main` first, move the two statements to `000117`.
- The daemon heartbeat already re-fetches and hot-reloads, so an update reaches a running sidecar without a restart.


matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1801.0.0-gae4adbb

sandromello:
✅ Build Completed with Success, Version=1801.0.0-g45e7c5b

p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1801.0.0-gdd98ce5

PR #1810, opened 2026-09-11
feat(sidecar): Analyzer first class feature

matheusfrancisco:
## 📝 Description

The AI analyzer is now a first-class, per-listener component instead of a guardrail rule (`type: ai_analysis`). Each listener takes its own `analyzer:` block — trigger, `high`/`medium`/`low` actions, `prompt`, `message` — beside `guardrails`, `opa` and `mask`. The top-level `analyzer:` section keeps what is genuinely process-wide (provider, model, credential) and becomes the defaults every lane inherits and overrides field by field (`send`, `fail_open`, `timeout_sec`, `max_input_bytes`, `max_calls`, `cache`).

The old rule spelling **keeps working**: it loads, builds the same evaluator through the same code path (`specFromRule` → `buildAnalyzerEvaluator`), and prints a deprecation naming the block. `-strict` turns the warning into a non-zero exit. Both spellings can serve one lane during migration, each as its own evaluator.

`hoop-inspect -migrate -config config.yaml -migrate-out config-new.yaml` (also `hoop start sidecar --migrate`) rewrites any old config onto the current schema: it folds every deprecated field through the same strict decoder the daemon uses, moves `ai_analysis` rules onto listener blocks only where the move is faithful (one rule per lane; a top-level rule only when every lane is free), reports what it left for you on stderr, and emits a document — in the input's syntax, keys in the documented order — that passes `-validate -strict`.

Hardening that landed with the review of this branch:

- **`send: redacted` now actually redacts.** New `Plugin.RedactText` (alcatraz implements it from engine offsets) replaces each detected value with `<ENTITY_CLASS>` before the statement leaves the process. The previous behavior appended a note while transmitting the values.
- **`send: refuse` runs on cache hits.** The redact/refuse callback moved ahead of the verdict cache, so a sensitive literal can no longer ride a cached SQL shape past the local denial.
- **Effective `redacted`/`refuse` refused without a detector** — a lane-level override can no longer produce a nil redactor that transmits raw text.
- **Call budgets namespaced by form** (`rule:<name>` / `lane:<name>`): a deprecated rule named after a lane no longer spends that lane's `max_calls`.
- **Cache overrides merge field-wise** — `cache: {ttl_sec: 60}` keeps the inherited size instead of disabling caching.
[...]

PR #1813, opened 2026-09-11
docs(adr): ADR-0016, the control plane licenses the sidecar fleet

rogefm:
## 📝 Description

ADR-0016 records how a sidecar fleet gets its license: the control plane owns
it, and a sidecar connected to one takes it from there and nowhere else.

Its own PR on purpose — `docs/adr/README.md` asks for the decision to be
approvable, or objectable, without the diff that implements it. The
implementation is #1812, in draft.

Numbered 0016 because #1810 took 0015 while this was being written.

**The decision, in short**

- The organization's license is served to every sidecar inside the configuration
  document the handshake already answers with, in the `license` key the schema
  already declares. Filled at serve time; never stored per sidecar.
- A configured control plane is the only license source. A plane whose
  organization holds no license runs the free tier, and the local sources are
  ignored out loud. Standalone sidecars are untouched.
- A license replaced or removed in the control plane reaches a running sidecar
  on the next heartbeat, with no restart, and always as a signed document the
  sidecar verifies itself.

**The alternatives, and why they lost**

- A new response envelope, `{"configuration": …, "license": …}`, reads cleaner —
  a license is not config. It changes the response format, and the sidecar
  decodes that body with `DisallowUnknownFields`, so it needs a version gate on
  a response that has no versioning.
- A dedicated `GET /api/sidecars/license`, polled beside the handshake, needs no
  format change, but doubles the requests per sidecar per minute and splits one
  fact across two responses that can disagree.

**What it commits us to, openly**

A license that grants less than the running config needs is not applied hot. A
control plane that becomes unreachable cannot revoke from a partitioned sidecar.
Both are written down as consequences rather than left to be discovered.

## 📣 User-facing impact

None. Documentation.

## 🔗 Related Issue

No GitHub issue. Tracked in the Linear project "Connect License with Sidecar",
following DEP-191.

Fixes #

## 🚀 Type of Change

- [x] 📚 Documentation update

## 📋 Changes Made

- `docs/adr/0016-control-plane-license-for-sidecars.md`, `Status: Proposed`.

## 🧪 Testing

Not applicable — no code. The claims about current behaviour it cites
(`DisallowUnknownFields` on three sides, `Config.UseLicense` taking a reference
and never a verdict, the heartbeat as the poll) are checkable against the files
named in the ADR.

## ✅ Checklist

## 📄 Additional Notes

Deciders listed from the Slack thread where the call was made. Correct the
frontmatter if I attributed anyone wrongly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SwvguWWtrqnnimaZLjdES9)_


sandromello:
✅ Build Completed with Success, Version=1813.0.0-ge3f9c90

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1813.0.0-g82b21e3

PR #1814, opened 2026-09-14
EVL-259: reconcile stale reviews in control-plane mode too

p3rotto:
## 📝 Description

`ReconcileStaleReviews` was written inline inside `runGateway`, so only a gateway ever ran it. It is now a named function that both boot paths call.

**Worth knowing before you read the ticket:** the ticket says "a control-plane review left in PROCESSING or UNKNOWN never settles". Taken literally that case is not reachable. Nothing in control-plane mode writes `PROCESSING`: the only two writers are `transport/plugins/review/review.go:71` and `transport/interceptors/accessrequest/accessrequest.go:73`, both gRPC transport code, and `controlPlanePlugins` is Slack only with no gRPC server started. `UNKNOWN` has no writer at all; the model calls it legacy.

What makes the change worth having is the topology ADR-0013 describes: a control plane may share its database with a gateway. The rows this settles there were written by the gateway, and the `UPDATE` is idempotent, so both processes settling the same row costs nothing. The comment on the function says exactly this rather than implying the control plane produces these rows itself.

## 📣 User-facing impact

On a control plane sharing a gateway's database, reviews left `PROCESSING` by an execution whose session finished during a restart now settle as `EXECUTED` instead of sitting in the list forever.

## 🔗 Related Issue

EVL-259.

## 🚀 Type of Change

## 📋 Changes Made

- `gateway/main.go`: `reconcileStaleReviews(db *gorm.DB)` extracted from `runGateway`'s inline goroutine; both `runGateway` and `runControlPlane` now call `go reconcileStaleReviews(models.DB)`.
- `gateway/reconcile_test.go`: new, two database-backed tests.

The function takes the `*gorm.DB` rather than reading the global, which is what makes it testable at all. `runControlPlane` blocks on `StartAPI()`, so the boot path itself cannot be called from a test.

### Deliberately not in scope

`credentialsweeper.Run` is the other background job that needs nothing but a database, and it is still gateway-only. Every other one is transport-bound (`connectionstatus` reads gRPC state, `InitProxyMemoryCleanup` walks gRPC proxy streams). Worth its own ticket; widening this one would have gone past what EVL-259 asked for.

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: macOS 15.6, PostgreSQL 17 (live run), embedded pglite (tests)

### Tests performed:
- [x] Unit tests pass

`make test-oss` green: exit 0, 5117 passes, 0 failures.

Both new tests run against the real schema on embedded pglite, because the reconciliation is one `UPDATE` joining two tables on a status enum and an in-memory fake would not exercise what breaks. They are mutation-tested, and each mutation fails the intended test:

| Mutation | Test that fails |
|---|---|
| the job body never runs | settles-a-finished-execution |
| drop `s.status = 'done'` | leaves-a-review-nobody-ran-alone |
| drop `r.status IN (PROCESSING, UNKNOWN)` | leaves-a-review-nobody-ran-alone |

The second test is the one that matters: settling a `PENDING` review as `EXECUTED` would close it with no approval and let the statement past the gate.

**Live verification**, which is the part the unit tests cannot reach, since they call the function directly and never touch `runControlPlane`. Against a real Postgres: boot a control plane once to migrate, seed a finished session with a `PROCESSING` review plus a `PENDING` control row, then boot again.

```
before:  aaaa…|PROCESSING   bbbb…|PENDING
log:     gateway/main.go:237  reconciled 1 stale review(s) to executed status
after:   aaaa…|EXECUTED     bbbb…|PENDING
```

## 📄 Additional Notes

The `runControlPlane` doc comment said "the HTTP API, and Slack"; it now also names the database-only background work, so the list stays true.

https://claude.ai/code/session_017sx9jEmnE7kqHjUSQxJ91L


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1814.0.0-g40e884f

sandromello:
✅ Build Completed with Success, Version=1814.0.0-g13c096c

PR #1815, opened 2026-09-14
feat: source of truth

EmanuelJr:
## What & Why

The sidecar now honors the control plane's `load_from_disk` answer. Before, the flag existed on the gateway side but the sidecar ignored it.

- **Disk mode**: when the plane answers `{"load_from_disk": true, "license": ...}`, the sidecar serves its own config file's document (listeners, guardrails, masking, PII) and keeps the plane connection for heartbeat, license, and last-seen.
- **License**: the plane's license goes through the existing `UseLicense` seam, above every local source. A license drift (renewal, shortened term) hot-applies through a shared holder read by the expiry watcher, the admin endpoint, and the reloader — renewal never costs a restart.
- **Ownership flips are hot**: flipping `load_from_disk` on the plane swaps the incoming owner's document (the plane's full config on takeover, the re-read file on handover) through the same rule-only-vs-topology reload machinery. Rules/masking/PII swap live in both directions; only a listener/upstream change stays restart-bound, same as plane-owned reload today (ADR-0014).
- **Guard rails**: `load_from_disk` in a local config file is refused at startup and on handover — only the plane says it. A disk-mode answer with no file, no listeners, or a garbage license fails loudly.

## Notes for Reviewers

- Disk-mode license precedence is plane > flag > `HOOP_LICENSE` > file, matching the documented `UseLicense`/`ResolveLicense` contract. Plane-owned mode still resolves its injected license at config-key precedence; aligning it is a possible follow-up.
- License **removal** (plane drops to free tier) is restart-bound on purpose: caps narrow and running lanes may exceed them — same posture as expiry (`watchLicense` holds the rationale).
- On handover the file is **re-read from disk**, not served from a startup snapshot; `SetupWith` retains the path and loader on `controlPlane` for that.
- `watchLicense` and `serveAdmin` now take `*atomic.Pointer[license.Status]` instead of a value; `reloader.apply` split into the flip switch + `applyOwned(raw, from)` with the owner named in log lines.
- Verified end-to-end against a local gateway: one relay process, zero restarts — disk mode serves the file; flip to plane-owned applied a `deny_words_list` rule live (statement denied); flip back removed it (statement passed). Unit tests cover both flips, license drift/refusal/removal, watchdog follow-through, and the startup refusals.


sandromello:
✅ Build Completed with Success, Version=1815.0.0-ga115e6f

gateway/services/sidecar.go
```diff
@@ -49,5 +49,58 @@ func ParseSidecarConfiguration(raw json.RawMessage) (daemon.Config, error) {
 	if err := dec.Decode(&cfg); err != nil {
 		return cfg, fmt.Errorf("invalid sidecar configuration: %w", err)
 	}
+	// load_from_disk false is the same fact as absent: the control plane owns
+	// the document. Drop it so the stored document never carries a key an
+	// older sidecar rejects via DisallowUnknownFields on its next handshake.
+	if cfg.LoadFromDisk != nil && !*cfg.LoadFromDisk {
+		cfg.LoadFromDisk = nil
+	}
```
matheusfrancisco, line 57: hey qodo this mean a running sidecar when restarting, and is there a control plane? Will this happen?

sandromello:
✅ Build Completed with Success, Version=1815.0.0-gc0392f6

matheusfrancisco (approved):

sandromello:
✅ Build Completed with Success, Version=1815.0.0-g0f7b22c

p3rotto (approved):

sandromello:
✅ Build Completed with Success, Version=1815.0.0-g8149df1

PR #1817, opened 2026-09-15
sidecar: log the pgwire principal once on the session line

NaurisSadovskis:
[claude]

## Problem

On a postgres lane the `session opened` and `session closed` log lines carry two `principal` keys:

```json
{"msg":"session opened","principal":"anonymous","principal":"app_user","upstream":"db.example.internal:5432"}
```

The logger is built with `principal=anonymous` before the pgwire StartupMessage is read. When the user is learned, `log.With("principal", user)` appends a second key. slog never replaces an attribute. The deferred teardown logs through the same logger, so `session closed` shows it too.

## Fix

Rebuild the logger from the root with the final principal. One line in `sidecar/proxy/proxy.go`.

## How to test

`make test-sidecar`, or the one test:

```sh
cd sidecar && CGO_ENABLED=0 go test ./proxy/ -run TestSessionLinesLogThePrincipalOnce -count=1
```

The test starts a postgres lane with a JSON logger, sends a StartupMessage for user `alice`, and decodes the `session opened` and `session closed` lines with a token decoder that counts top-level `principal` keys. It fails on the old code with `principal keys = ["anonymous" "alice"]`.

Manual: open `psql` through a postgres lane and check both lines have one `principal` key.

## Not in scope

The `session_start` audit event also reports `anonymous`, because the gate starts before the startup message is read. The store folds later events, so the session row ends up correct. Reordering that is a separate change.


NaurisSadovskis:

PR #1819, opened 2026-09-15
control-plane: create, edit and delete sidecar listeners

rogefm:
> [!IMPORTANT]
> Labelled `minor`: new functionality, no breaking change. **Frontend only** — nothing outside `webapp_v2/`.

## 📝 Description

The Control Plane can hold a sidecar's configuration but had no way to author the listeners inside it. This adds the listener flow — list, expand, create, edit, delete — on top of the existing sidecar CRUD. No new endpoint and no backend change: a listener is an element of the configuration document, not a row, so the save is a read-modify-write of the whole document through the `PUT /sidecars/:nameOrID` that already exists.

The form owns twelve keys and preserves the rest. `guardrails`, `mask`, `opa`, `analyzer` and the two deprecated spellings are never rendered and must survive a round trip byte for byte: the document is decoded with `DisallowUnknownFields` at three layers, so a dropped key is silent data loss rather than an error anyone sees.

Squashed onto current `main`, which landed #1807 in the same domain. The service returns the raw Axios promise again, `updateSidecar` unwraps it and moves both generation counters, and the listener form reads `selected` from the store instead of hand-rolling its own loading and error state.

## 📣 User-facing impact

Sidecar listeners can now be managed from the Control Plane instead of by editing the sidecar's own config file by hand. The listener list shows what each one enforces — guardrails, masking, AI analyzer, OPA — and where each rule came from.

## 🔗 Related Issue

Tracked in Linear. No GitHub issue.

## 🚀 Type of Change

- [x] 🎨 Style/UI update

## 📋 Changes Made

**The flow**
- Listener list with an expandable row: the rules a listener actually runs, each marked `Listener` or `Inherited`, because the merge is not a union — guardrails concatenate, masking and OPA replace, and each has a spelling that means "none" rather than "inherit".
- One page per listener with a pinned footer, not a modal.
- `resolve.js` ports `Config.resolve` from the daemon so the UI and the sidecar agree on what a listener runs.

**Validation, client side**
- `validateListener` mirrors what the daemon refuses, so the form says it next to the field instead of letting a save become a sidecar that will not boot: duplicate names and bind addresses, half a downstream keypair, capture without descriptors, the four forbidden headers, and the capture an analyzer, PII rules or mask rules need.
- Some of these are checks **no server runs**. See Additional Notes.

**Fixes found while building it**
- `identity_header` was dropped from every non-http listener. gRPC and Spanner listeners read it, so an unrelated edit silently downgraded them to the peer certificate.
- The AI Analyzer chip appeared on every listener: the code read the process-wide `analyzer` section instead of the listener's own block.
- A listener with `mask: {rules: []}` — masking switched off — showed the Data Masking chip.
- `downstream_tls` was rebuilt from the two rendered fields, dropping anything else on the block.

**UI**
- `Badge` gains `tag` (a phrase, not a status), `icon` and `chip`, with the geometry in `Badge/theme.js` and a module. Address and feature chips were misaligned and the icon-only ones had five pixels of dead space; both are now centred, one height across the row.
- `variant="warning"` mapped to `yellow`, a palette this theme does not define, and painted white on it at 1.57:1. Now amber at 5.23:1.
- `SectionRow` and `FormFooter` graduated to `src/components/` from their page-local copies.
- Help text and validation messages say **listener**, not *lane*. The page, the table, the button and the field all said listener while the help text said lane — one screen naming the same object twice.

**Follows the docs, deliberately**
- The protocol picker lists the six protocols the published reference names. The daemon accepts a seventh, `spanner`, which the docs still carry as unreleased. Offering it would put a choice in front of an operator that the documentation denies.
- Nothing narrows what the form *writes*: a listener already on `spanner` keeps its blocks, round trips byte for byte, and keeps its own protocol as an option so the field never renders blank. When the docs gain it, one list changes.

### Test Configuration:
- **Browser(s)**: Chromium 1.56 (Playwright), 1440×880 and 400×900
- **OS**: Linux

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

**How to test**

1. `make run-dev-control-plane` (port 8019), then `API_URL=http://localhost:8019 npm run dev` in `webapp_v2/`.
2. Create a sidecar, then `PUT /api/sidecars/:id` a configuration with guardrails and mask rules at both the top level and on one listener.
3. Open the sidecar. Expand a row: rules show with the correct `Listener` / `Inherited` badge.
4. Add a listener on an address another one already binds — refused next to the field.
5. On a listener carrying `analyzer` or a `pii` guardrail, untick the capture switch — refused, because the daemon would not start from it.
6. Edit a listener that has `guardrails`, `mask` and `opa`, change only the name, and diff the stored document: everything else is unchanged.

`npm run lint` and `npm run build` green. Full CI green on the previous head.

## 📸 Screenshots

listener-flow (2)

Generated by Claude-Code

## ✅ Checklist

https://claude.ai/code/session_01WaDjmdUdiRoMXdZBcxwn2r


sandromello:
✅ Build Completed with Success, Version=1819.0.0-g8eb8260

rogefm:
Worked the review. Six findings were real and are fixed in `09a97078`; four I am pushing back on; three need a decision that is not mine.

## Fixed

**#1, #2, #4 — the store.** All three are one defect and the review is right about every part of it. `updateSidecar` treated the Axios response as the sidecar, so `updated.id` was `undefined` and no fleet row was ever replaced; it left `selected` untouched; and it bumped `requestId`, which exists nowhere in this store, so it stored `NaN`.

The reachable consequence is #2, and it is worse than a stale render. The details page deletes **without navigating**, and builds the next whole-document PUT from `selected.configuration`. I reproduced it end to end against the pre-fix code — deleting two listeners in a row:

```
before      : ["payments-db","billing-api","reporting-ro"]
after 1st   : ["payments-db","billing-api","reporting-ro"]   <- UI never updated
after 2nd   : ["payments-db","billing-api","reporting-ro"]
server holds: ["billing-api","reporting-ro"]                 <- the 2nd delete did nothing
```

After the fix the same script gives one removal per click and the server agrees. Worth saying plainly: this survived my own testing because the *save* path navigates back, which unmounts the page, clears `selected` and refetches — masking the bug. Delete does not navigate. That is the gap.

**#3 — masking is configured by raw bytes.** Correct, and a good catch. `MaskConfig.hasRules()` is `len(m.Rules) > 0 && !isEmptyJSONList(m.Rules)` over a `json.RawMessage`: the four bytes of `null` read as configured, and only `[]` reads as off. My resolver flattened `null` to an empty array — right for "what rules run", wrong for "does the daemon think masking is on", and the descriptor check asks the second question. Split into `maskConfigured()` for that question, with the resolved list unchanged for display.

**#6 — negative limits.** Correct. A `NumberInput`'s `min` stops the spinner, not a paste. Both limits are now refused in validation and the builders will not serialize a negative even if validation is bypassed.

**#7 — unsupported protocol.** Correct, and it is a consequence of keeping a listener's current protocol selectable when the picker does not offer it. Readable is not savable: validation now checks the daemon-supported set, not the offered set. `spanner` — supported, just undocumented — still saves.

**#5 — label collision.** Correct. An unnamed listener's generated `listener[i]` can equal another listener's real name. Names are unique, so a real name now wins; the generated label only resolves when no listener carries it.

Each one has a test that fails without the fix.

## Pushing back

**#10 — direct Mantine imports.** These are layout primitives (`Stack`, `Text`, `Title`, `Group`, `Grid`) that have no wrapper. `webapp_v2/CLAUDE.md` says a `sections/` file "may still import Mantine directly if no wrapper exists yet". The review's own relevance rating is Weak and it cites two precedents that rejected this.

**#11 — `@/` alias for intra-page imports.** `CLAUDE.md` says "Group: external libraries first, then `@/` imports, then **relative imports**" — relative imports inside a page are the documented convention, not a violation. Weak, and the cited precedent rejected it.

**#13 — `SectionRow` leaves ten copies.** Consolidating ten unrelated pages is not this PR's scope. The component was extracted so the eleventh copy did not get written; migrating the rest is its own change.

**#8 — docs conflict.** Right, and already known: it is one of eight divergences between the listener docs and the shipped product that I found while building this, and it is recorded in the PR body. It needs its own PR on `hoophq/documentation`.

## Needs a decision

**#9 — no unit tests.** Fair, and I cannot close it here: `webapp_v2` has no test runner at all — no vitest, no jest, no script. Adding one is a project decision, not something to slip into a feature PR. What exists today is 52 assertions I run outside the repo covering exactly what #9 asks for: round-trip preservation, protocol gating, inheritance, duplicate detection, and now one per finding above. They can be landed as `*.test.js` the moment someone says which runner.

**#12 — `ListenerForm` placement.** The rule points at `sections/` (one consumer); the precedent the review cites points the other way. Happy to move it if a human reviewer prefers — one rename and one import.

---
_Generated by [Claude Code](https://claude.ai/code)_

sandromello:
✅ Build Completed with Success, Version=1819.0.0-g875b8d1

sandromello:
✅ Build Completed with Success, Version=1819.0.0-g36642d7

rogefm:
Finding 7 ("Some listener links edit the wrong row") is closed in `85c36c9b`.

`listenerIndexByLabel` already preferred the row that actually carries the name, so Edit on the named listener opens the named listener. That left the mirror case: the unnamed row at position 0 shares the URL and becomes unreachable. So `validateListener` now refuses a name matching `^listener\[\d+\]$` — that is the shape `listenerLabel` generates for an unnamed row, and reserving it keeps the two label sets disjoint. Names are already unique and generated labels are unique by position, so no document this form can write has two rows under one label.

I did not take the recommendation to put the position in the edit URL. The label is what an operator can read and share, and the editor works on the position underneath, so a rename stays one edit; reserving nine characters of name-space costs less than that.

Three checks, each failing without the change: the reserved shape is refused; `listener[abc]`, `my-listener[0]`, `listener0`, `listener[]` and `listener[0]x` are not; and every row of a form-written document resolves back to itself.

Same commit drops a duplicated copy of the negative-limit checks that `09a97078` left in `validateListener` — no behaviour change, the assignments were identical.

---
_Generated by [Claude Code](https://claude.ai/code)_

matheusfrancisco (commented):
hey qodo , can you answer this for me today this is the current state of 

 Grounded in sidecar/daemon/controlplane.go, reload.go, and README "Connecting it to a Control Plane":

 Steady state: no sync of the file. The plane owns the config.
 - Once the plane holds a configuration, the local file is never written, never re-read, never merged. Listeners still in the file are ignored with a warning (the config file's
   listeners are ignored: the control plane owns the running config, daemon.go:747).
 - The direction is plane → sidecar only. The heartbeat (controlPlane.heartbeat, every minute) re-runs POST /api/sidecars/handshake; the response is the running config. Rule-only
   drift is hot-applied (ADR-0014, reload.go); topology drift logs restart to apply it. A failed heartbeat changes nothing.

 The one exception: first connect to an empty plane.
 - If the handshake answers 412 (or 200 with no listeners), the sidecar does a one-shot PUT /api/sidecars/configuration with the file's document, minus control_plane_url and license
   (importLocalConfig, controlplane.go:345). Then it serves what the plane sends back. Only happens when the plane holds nothing; afterwards the plane is authoritative.

 File-side facts that stay local regardless: control_plane_url and license. The plane sends no license today



what would happen in this cases:
Sidecar running with 2 listeners and some rules; the control plane is added with the license. What will happen with the listeners running? 

sidecar running with 2 listeners and some rules; the control plane is added with the license and the same listener and rules what will happen? 

Sidecar running with 2 listeners and some rules; the control plane is added with the license and 2 different listeners and rules what will happen? 


can you think more cases?

matheusfrancisco:
hey qodo , can you answer this for me today this is the current state of 


what would happen in this cases:
Sidecar running with 2 listeners and some rules; the control plane is added with the license. What will happen with the listeners running? 

sidecar running with 2 listeners and some rules; the control plane is added with the license and the same listener and rules what will happen? 

Sidecar running with 2 listeners and some rules; the control plane is added with the license and 2 different listeners and rules what will happen? 


can you think more cases?

rogefm:
@matheusfrancisco — the daemon is in this repo (`sidecar/daemon/`), so here is a traced answer rather than an inferred one. Two corrections to the reply above first, because both change what an operator does:

**1. `reloadRestart` does not restart anything.** It logs `"the control plane changed the configuration beyond the rules; restart to apply it"` (`sidecar/daemon/reload.go:271`) and the process keeps serving the OLD topology. Nothing consumes that outcome outside tests, and the heartbeat discards it (`controlplane.go:582`). Someone — an operator or an orchestrator — has to restart the sidecar.

**2. The plane does send a license today.** `Handshake` returns it in the answer and sets `hoop-sidecar-license-managed` (`gateway/api/sidecar/sidecar.go`); the daemon moves it onto the connection and stops treating the file's key as a source, warning that it is ignored (`daemon.go:889`). That changed in #1812.

One framing note before the cases: "the control plane is added" always means a restart. `control_plane_url` and the token are read once, by `resolveConfigSource` (`controlplane.go:220`), and the file is never re-read afterwards.

**Case 1 — sidecar runs 2 listeners + rules, plane holds nothing.**
The handshake answers 412 (`controlplane.go:278`; an older gateway answers 200 with no listeners — same branch). `importLocalConfig` (`controlplane.go:441`) PUTs the file's whole document minus `control_plane_url` and `license`, then re-fetches and serves what comes back. The same 2 listeners and rules keep running, now owned by the plane. If someone authored a config between those two requests the push gets 409, **their** document wins, and the file's listeners are ignored with a warning (`daemon.go:884`).

**Case 2 — plane holds the same 2 listeners and rules.**
No import: the handshake answers 200 with listeners, and that document is what loads. The file's listeners are still ignored, never merged — same warning. On each heartbeat `handle` sees the identical bytes and returns `reloadUnchanged` (`reload.go:158`): nothing swapped, nothing logged.

**Case 3 — plane holds 2 different listeners and rules.**
At startup the plane's document wins outright; the file's two never run. While running, the heartbeat notices within a minute (`heartbeatEvery`, `controlplane.go:72`) and `nonRuleDoc` (`reload.go:435`) decides:
- **rules only** — guardrails, mask, opa, pii, and a lane's `analyzer` block — hot-swapped, new generation, no restart.
- **anything else** — listen, upstream, protocol, tls, the http/grpc blocks, audit, log_level, the top-level analyzer section — restart-bound, which per correction 1 means *it logs and keeps running the old lanes*.

`nonRuleDoc` renders the whole Config with the rule sections nulled, so a new field is restart-guarded by default. Forgetting to list it fails safe.

**More cases worth writing down:**
- **Document this build cannot load** → `reloadRefused`, running rules kept, warning. The heartbeat runs the same `LoadConfigBytes` and caps as startup, so it is never a side door past validation.
- **Plane unreachable, or the token rejected, mid-life** → warn and serve the last good config forever. It never kills the data path.
- **Token rejected at startup** → hard stop, no self-heal: a lost token means registering a new sidecar.
- **`load_from_disk` flipped on while running** (the toggle from #1815) → handled hot: `adoptFile` (`reload.go:207`) **re-reads the file**, stamps the plane's license into it, and runs it through the same swap-or-restart decision. No config file, a file that will not load, a file carrying `load_from_disk` itself, or a file with no listeners → each logs its own reason and stays restart-bound.
- **Flipped back off** → the plane's document goes through `applyOwned` and the mode flips only if it applied (`reload.go:186`).
- **License lapses while running** → the watchdog stops a process whose config the free tier would refuse. Re-applying the caps live would mean deleting guardrail and mask rules from a running proxy.
- **Two admins editing at once** → this PR's saves are whole-document PUTs with no ETag, so the second write wins silently. Worth an ETag or a `PATCH` on `listeners` before this gets busy; `PATCH /sidecars/:nameOrID` from #1815 already merges by key.

---
_Generated by [Claude Code](https://claude.ai/code)_

rogefm:
**CLAAssistant** is the one red check on `a8c3fcf2`. Everything else that has finished is green; the rest of the suite is still running.

It fails because two commits here are authored by `Claude <noreply@anthropic.com>`, which has no GitHub account and so can never sign the CLA. `2ebc5df2` passes — it carries @rogefm as author.

I am not fixing this myself: re-authoring a commit under someone's name is their call, not mine, and the CLA is a legal act attributable to them. @rogefm — `git rebase --exec 'git commit --amend --no-edit --reset-author' 3090c37b` then a force-push does it, and the `Co-Authored-By` trailers stay.

---
_Generated by [Claude Code](https://claude.ai/code)_

matheusfrancisco:
"Token rejected mid-life → serve forever" is right, but it also means a deleted sidecar keeps proxying indefinitely with its last config. Nothing in the daemon distinguishes
 "plane down" from "plane says you no longer exist" after boot.


If the user has a control plane and it deletes the sidecar from the control plane , what are our expectations?

rogefm:
@matheusfrancisco — you are right, and the code is worse than "indistinguishable". `fetchControlPlaneConfig` does build a distinct message for 401 (`controlplane.go:363`), but it is a plain `fmt.Errorf`, not a sentinel like `errPlaneHasNoConfig`. The heartbeat then throws the classification away entirely:

```go
raw, _, err := fetchControlPlaneConfig(cp.url, cp.token, Version)
if err != nil {
    log.Warn("control plane handshake failed; serving the last good config", ...)
    continue
}
```

So "the plane says you no longer exist" and "the plane is behind a dead load balancer" take the same branch, and there is no revocation path anywhere in the daemon — I grepped.

**What actually happens when an admin deletes a sidecar.** The row goes, and with it the hashed token, so every later handshake is 401. `forgetRuntime(deletedID)` also drops the in-memory runtime, so the fleet view stops showing it. The process keeps proxying its last good config, and the only two things that ever end it:

1. **A restart.** Boot is where revocation is enforced: `resolveConfigSource` surfaces the 401 and the process refuses to start ("check it against the one shown when this sidecar was created"). Correct and final.
2. **The license term ending**, and only if `dependsOnLicense()` is true — `watchLicense` stops the relay. A config inside the free tier is never stopped, so that bound can be a year, or never.

**The asymmetry is the part I would call a bug.** Removing the org's license stops a running sidecar inside a minute: the next successful handshake sets `overCap` and the watchdog does a controlled stop. Deleting the sidecar outright stops nothing — precisely because it *prevents* the handshake that would carry that verdict. The weaker action has the stronger effect.

**My expectation, for what it is worth.** Stopping the relay on a single 401 is not it: a gateway deploy, a DB blip or a token rotation race would take down the data path, which is the outage "serve the last good config" exists to prevent. But a *sustained* 401 is not ambiguous. I would give it a grace window — N consecutive 401s, say five minutes — and then the same controlled stop `overCap` already gets, with its own log line. That reuses a mechanism that exists rather than inventing one, and it fails closed only once the evidence is unambiguous. A transport error would keep serving forever, unchanged.

Until that exists, deletion is boot-time revocation, and the UI should say so where an admin deletes a sidecar — that a running process keeps proxying until it is restarted.

Both of those are daemon and fleet-page work, outside this PR, which is frontend-only and touches neither the delete flow nor `sidecar/`. Happy to open a ticket with this trace if you want it captured.

---
_Generated by [Claude Code](https://claude.ai/code)_

sandromello:
✅ Build Completed with Success, Version=1819.0.0-ga8c3fcf

matheusfrancisco:
Being more direct, I would like to ask

If the user deletes a whole sidecar from the control plane, should the sidecar with the old config file stop? As I understand it, if so, the sidecar will keep the old config file, right? 

Also, I'm not saying it's in this pr I just saw the screen edits and realized that.

rogefm:
@matheusfrancisco — direct answers.

**"It keeps the old config file" — no, and this is the part worth correcting.** A plane-owned sidecar is not running its file. It is running the last document the *plane* sent, held in memory. The file contributed `control_plane_url` and nothing else; its listeners were ignored from boot, out loud (`daemon.go:884`). After the delete, the process keeps serving that in-memory document. It never falls back to the file: `resolveConfigSource` reads the file once at boot, and the reloader re-reads it only inside `adoptFile` (`reload.go:213`), which runs only when the plane answers `load_from_disk: true`. That answer stops arriving the moment the sidecar is deleted.

Three shapes:
- **Plane-owned** — keeps serving the plane's last document, forever, from memory.
- **Started with no file at all** (URL + token via env) — same, and there is nothing it *could* fall back to.
- **`load_from_disk`** — it was already running the file, and keeps running it. The license stops rotating, since that came from the plane.

In every shape, a restart is fatal: the boot handshake 401s and the process refuses to start.

**Should it stop? Yes — and specifically it should stop *without* falling back to the file.** A fallback would be the worst of the options: the operator deleted the sidecar, and the process would resume serving a document that predates the plane entirely, possibly with rules the org tightened since. That is a silent policy downgrade, and the daemon already refuses to merge the two sources for that exact reason ("no operator can predict which half of a merged config wins", `controlplane.go:205`).

That is why I would not stop on the first 401 but would stop on a sustained one. A 500, a dial error or a gateway deploy are ambiguous; a 401 that repeats for five minutes is not. And "stop" lands well operationally: the process exits, the orchestrator restarts it, boot 401s, and it crash-loops with a message naming the cause. Loud and carrying no traffic beats a quiet proxy nobody is watching.

**The gap today is that deleting is weaker than revoking the license**, which does stop a running sidecar within a minute via `overCap`. Deletion should be at least as strong.

Agreed it is not this PR — and yes, the sidecar delete modal (which is on `main`, not in this diff) should say that a running process keeps proxying until it is restarted, for as long as the daemon behaves this way. Say the word and I will open a ticket with this trace.

---
_Generated by [Claude Code](https://claude.ai/code)_

rogefm:
> Being more direct, I would like to ask
> 
> If the user deletes a whole sidecar from the control plane, should the sidecar with the old config file stop? As I understand it, if so, the sidecar will keep the old config file, right?
> 
> Also, I'm not saying it's in this pr I just saw the screen edits and realized that.

Yes, the sidecar should keep the old file working normally. The sidecar shouldn't inherit configs from the control plane.

sandromello:
✅ Build Completed with Success, Version=1819.0.0-ge610a6b

p3rotto (approved):

luanlorenzo (approved):

mtmr0x:
> > Being more direct, I would like to ask
> > 
> > If the user deletes a whole sidecar from the control plane, should the sidecar with the old config file stop? As I understand it, if so, the sidecar will keep the old config file, right?
> > 
> > Also, I'm not saying it's in this pr I just saw the screen edits and realized that.
> > 
> 
> Yes, the sidecar should keep the old file working normally. The sidecar shouldn't inherit configs from the control plane

@rogefm @matheusfrancisco Rogerio is right in there, but here are the catches: 
- if the Control Plane deletes it, the Sidecar loses its authentication, meaning two things: enterprise features won't work; and Control Plane URL will fail connection. 
- if there's a Control Plane URL, and it refuses, the right path is to keep the sidecar working, maybe throwing warning retries logs that Control Plane URL failed to connect if there's a Control Plane URL set. **Why not fail entirely:** people can set up Control Plane after the sidecar, but knowing the Control Plane URL, and control plane could be temporarily offline, so if we avoid exiting in case Control Plane is off because of an inconsistency or just timing for the user setup, we open the possibility of a lot of journeys to just work rather than one depending on the other and creating weird cases
- If the config file has configurations that attend the free plan, it should just work
- If that sidecar has a license that is valid in its deployment, it should keep working along with Enterprise features. This avoids messy states and is simpler, we can think of a complete decomission of a Sidecar journey from the Control Plane, that it "bricks" the Sidecar, but let's not worry about this now

rogefm:
> > > Being more direct, I would like to ask
> > > If the user deletes a whole sidecar from the control plane, should the sidecar with the old config file stop? As I understand it, if so, the sidecar will keep the old config file, right?
> > > Also, I'm not saying it's in this pr I just saw the screen edits and realized that.
> > 
> > 
> > Yes, the sidecar should keep the old file working normally. The sidecar shouldn't inherit configs from the control plane
> 
> @rogefm @matheusfrancisco Rogerio is right in there, but here are the catches:
> 
> * if the Control Plane deletes it, the Sidecar loses its authentication, meaning two things: enterprise features won't work; and Control Plane URL will fail connection.
> * if there's a Control Plane URL, and it refuses, the right path is to keep the sidecar working, maybe throwing warning retries logs that Control Plane URL failed to connect if there's a Control Plane URL set. **Why not fail entirely:** people can set up Control Plane after the sidecar, but knowing the Control Plane URL, and control plane could be temporarily offline, so if we avoid exiting in case Control Plane is off because of an inconsistency or just timing for the user setup, we open the possibility of a lot of journeys to just work rather than one depending on the other and creating weird cases
> * If the config file has configurations that attend the free plan, it should just work
> * If that sidecar has a license that is valid in its deployment, it should keep working along with Enterprise features. This avoids messy states and is simpler, we can think of a complete decomission of a Sidecar journey from the Control Plane, that it "bricks" the Sidecar, but let's not worry about this now

Great! I understand that this is not related to this PR, so I will merge this one :)

PR #1825, opened 2026-09-16
EVL-288: return the existing sidecar review and consume an approved one once

p3rotto:
## 📝 Description

A sidecar that holds a statement retries it. Today every retry files a new review and posts a new Slack message, and nothing ever tells the sidecar a statement was approved, so an approval is unreachable.

`POST /sidecars/reviews` now hashes the exact statement bytes and looks for a live review on org, sidecar, listener, rule and hash. `PENDING`, `REJECTED` and `REVOKED` come back as they stand. `APPROVED` is claimed by one conditional `UPDATE`, and only the request that wins the claim gets `forward: true`.

The exact bytes, not the analyzer's cache key: that key strips literals, so approving `DELETE FROM users WHERE id = 1` would also release `id = 999`.

## 📣 User-facing impact

None. The sidecar half is EVL-289, and `require_review` is still refused at sidecar startup, so nothing reaches this path yet.

## 🔗 Related Issue

EVL-288. Follows EVL-287, blocks EVL-289.

## 🚀 Type of Change

## 📋 Changes Made

- Migration `000118`: `statement_hash CHAR(64)` on `private.reviews`, plus a partial unique index on org, sidecar, listener, rule and hash over every status except `EXECUTED`. Racing first requests cannot both file; a rejection keeps denying; a consumed approval stops blocking.
- `models.HashStatement`, `GetLiveSidecarReview`, `ClaimApprovedSidecarReview` and `CreateSidecarReview`.
- `UpsertSession` and `CreateReview` gain transaction-scoped variants, no behavior change, so the session and the review share one rollback boundary.
- Response is `openapi.SidecarReviewResponse` with an explicit `forward` boolean. The status cannot carry it: a claim loser also reads `EXECUTED`.
- `EventConsumeSidecarReview`, emitted only on a won claim.

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: macOS 15.6, Go 1.26.5, PostgreSQL 17.6 and embedded PGlite

### Tests performed:
- [x] Unit tests pass
- [ ] Integration tests pass

```
go test ./gateway/models/... ./gateway/api/sidecar/...
```

Covered by unit tests: `HashStatement` (exact bytes, so `id = 1` and `id = 999` stay distinct; stable and not folded for case or whitespace) and `newSidecarReview` carrying the hash onto the row.

Migration rollback, run by hand against a throwaway PostgreSQL:

```
migrate -path gateway/migrations -database "$URI" up      # 118
migrate -path gateway/migrations -database "$URI" down 1  # column and index gone
migrate -path gateway/migrations -database "$URI" up      # re-applies, not dirty
```

## ✅ Checklist

### Known gap: the database behavior has no automated coverage

This PR deliberately ships unit tests only. Three guarantees are enforced by SQL and cannot be reached from a unit test, so nothing in CI defends them:

1. The partial unique index, which stops two racing first requests from both filing.
2. The conditional `UPDATE` in `ClaimApprovedSidecarReview`, which is what makes an approval consumable exactly once.
3. The `gorm.ErrDuplicatedKey` fallback in `PostReview`.

They were verified during development with a control-plane integration suite (9 tests, PostgreSQL and PGlite) that was removed before merge. Each guarantee was confirmed by breaking it deliberately and watching a test fail:

| Mutation | Result |
|---|---|
| Claim not gated on the conditional `UPDATE` | 8 consumptions instead of 1 |
| Hash strips digit literals | `id = 1` and `id = 999` merged |
| Hash unique per request | dedup, rejection and filing-race all broke |

The suite is recoverable from commit `41c0bae8` if it is ever wanted back. Reviewers should know that a regression in any of the three above will not be caught by CI.

### Worth a reviewer's attention

The real approval path is `models.UpdateReview`, which hands a whole struct to GORM `Updates`. `GetReviewByIdOrSid` does not select `statement_hash`, so an approved review carries a zero value for it, and the hash survives only because GORM skips zero-valued fields. If that changes, the approved review drops out of the index and the approval becomes unreachable, which is this ticket's bug reintroduced through the approval itself. This was covered by the removed suite and is now unguarded.

### Left alone deliberately, as this project is control-plane only

- The `jsonb_agg` review-groups subquery is now written three times in `gateway/models/reviews.go`. Extracting it would touch `GetReviewByIdOrSid` and `ListReviews`, both gateway review paths.
- `UpdateReview` sets a session to `done` without `ended_at`, unlike every other close in the codebase. Pre-existing and gateway-side; the new claim sets both.


sandromello:
✅ Build Completed with Success, Version=1825.0.0-g41c0bae

sandromello:
✅ Build Completed with Success, Version=1825.0.0-gad9bda8

sandromello:
✅ Build Completed with Success, Version=1825.0.0-g9b0a81c

sandromello:
✅ Build Completed with Success, Version=1825.0.0-g3e9ebb7

matheusfrancisco (approved):

PR #1826, opened 2026-09-16
sidecar: gs:// descriptor sets for grpc lanes; raw HTTP path and query to the analyzer

matheusfrancisco:
## 📝 Description

Three sidecar changes. All live under `sidecar/`; none touches the gateway ↔ agent contract.

**gRPC descriptor sets from GCS.** A `listeners[].grpc.descriptors` entry may be `gs://BUCKET/OBJECT[?generation=N]`, mixed with file paths. A descriptor set for a large API does not fit a ConfigMap (1 MiB), and several teams publishing sets on their own cadence should not need a redeploy of the sidecar's volume; a bucket is where CI already puts the artifact. The sidecar fetches at startup with Application Default Credentials, merges URLs and files into one schema, and hands it to libhoop already loaded.

```yaml
grpc:
  descriptors:
    - gs://acme-schemas/billing/v42.pb
    - gs://acme-schemas/ledger.pb?generation=1726480000123456
    - /etc/hoop/local-override.pb
```

**Per-database dialect on spanner lanes.** A Spanner database is created as GoogleSQL or as the PostgreSQL interface, one instance holds both, and the data plane never says which: `ExecuteSql` carries a session and a string. Read PostgreSQL with the GoogleSQL lexer and `"songs"` is a string literal, so a table rule fencing `songs` never fires. The dialect is now configuration, keyed on the database resource name every SQL-bearing request carries, never inferred from text a client controls.

```yaml
spanner:
  dialect: googlesql                 # lane default; absent block = googlesql
  databases:
    projects/p/instances/i/databases/ledger-pg: postgresql
```

**The literal request target to the analyzer.** The analyzer on an HTTP lane sent the model `METHOD /users/*/orders` and the body. It now sends the request-target as the client sent it (libhoop `HTTPDetail.Target`, hoophq/libhoop#120), then the normalized resource on its own line. The model judges intent; `?export=all`, `?limit=100000` and the literal id are part of it. The resource form exists to remove them for policy, not for classification.

## 📣 User-facing impact

- `grpc.descriptors` accepts `gs://` URLs. Same key, same single-or-list spelling; existing configs do not move. A scheme the binary does not link (`s3://`) is refused at config validation naming what is linked. A changed `descriptors` list, or a new object version behind an unchanged URL, is applied by a restart, not a reload — the same as a replaced file.
- New `spanner:` block on spanner lanes. Absent means GoogleSQL everywhere, today's behavior. `dialect: per_database` is the fail-closed shape: SQL against an unlisted database is `unknown`, so a rule naming `unknown` refuses it. Each extracted statement records `spanner.dialect` and `spanner.database`.
- HTTP lanes with an analyzer send the model the literal path and query. Path identifiers reach the model under the existing `send: redacted` pass. The cache keys on resource, query string and body, so a different `?export=` value is a new verdict. No config change; body-less requests stay skipped and `capture_body: true` stays required.
- Credentials for `gs://`: a service account key inline in `GOOGLE_APPLICATION_CREDENTIALS_JSON` (the gateway's own variable, so one Secret serves both processes) is consulted first; otherwise Application Default Credentials (Workload Identity, attached service account, `GOOGLE_APPLICATION_CREDENTIALS`). Set but malformed is an error naming the variable, never a fallthrough to another identity, and the key material is never echoed. `.env.sample` documents both; the sidecar chart README shows the three deployment shapes.

## 🔗 Related Issue

Depends on hoophq/libhoop#119 and hoophq/libhoop#120 (both merged, pinned here at `c49b31f`).

## 🚀 Type of Change

## 📋 Changes Made

gRPC descriptors:

- libhoop pin bumped in `sidecar`, `sidecar/cmd`, `client`, `agent`, `gateway`: `SchemaSource`, `LoadSchemaSources`, `NewServerWithSchema`, `HTTPDetail.Target`.
- `sidecar/descriptors/` (root, stdlib only): fetcher registry by URL scheme. `Scheme` requires `://`, so `C:\x` and `dir:with:colons` stay files.
- `sidecar/descriptors/gcs/` (nested module, `golang.org/x/oauth2` only, no `cloud.google.com/go/storage`): one GET on the JSON API, `devstorage.read_only`, object name path-escaped, `?generation=N` passed through via `url.ParseQuery` (a malformed or unknown parameter is refused, never dropped, so a typo cannot silently unpin), 512 MiB cap. Credentials: `GOOGLE_APPLICATION_CREDENTIALS_JSON` via `google.CredentialsFromJSON` first, ADC second. The access token is cached, not the token source: each fetch resolves credentials under its own deadline, and the metadata-server mint is bounded by a select so a backend that never answers is a failed fetch, not a hung startup. On the success path the body is closed explicitly and a close error is returned instead of the bytes. Registers `gs` on import; linked in `sidecar/cmd/main.go` and `client/cmd/startsidecar.go`.
- `sidecar/daemon/grpc.go`: reads files and fetches URLs itself, one merge so a conflict between a file and a bucket artifact is caught, `NewServerWithSchema`. `GRPCCodecConfig.validate` refuses an unlinked scheme. `-validate` performs the fetch.
- `sidecar/daemon/reload.go`: `grpc.descriptors` is restart-bound drift, like TLS material.
- `go.work`, `sidecar/cmd/go.mod`, `client/go.mod`: the new module.

Spanner dialect:

- `sidecar/inspect/sqlmeta.go`: `AnalyzeSQLIn(sql, lexer.Dialect)`; `AnalyzeSQL` keeps the protocol switch and delegates.
- `sidecar/daemon/spanner.go`: `SpannerConfig{Dialect, Databases}` with validation (dialect names, `projects/P/instances/I/databases/D` keys, `per_database` needs a map). `spannerSQLStatements` now yields the database (`session` or `database` field) and, for `CreateDatabase`, the request's own `database_dialect`, which wins for that request's statements. `dialectFor` resolves declared → mapped → lane default; `per_database` refuses to guess.
- `sidecar/daemon/grpc.go`: dialect decided once per message; an unlisted database under `per_database` takes the same fail-closed path as an unreadable rendering.
- `sidecar/daemon/config.go`: `ListenerConfig.Spanner`, refused on a non-spanner lane, the rule `http:`/`grpc:` follow.

Analyzer:

- `sidecar/analyzer/content.go`: `HTTPBuilder.Build` renders `METHOD <target>`, a `Resource:` line where it differs, content type, body; `httpTarget` reads `HTTPDetail.Target` and falls back to path plus canonical query for a hand-built detail. `httpCacheKey` keys on method, resource, encoded query and body.

Docs and demos:

- `sidecar/README.md` (Protocols: gRPC descriptors, spanner dialects), `sidecar/CLAUDE.md` module tables, `deploy/helm-chart/chart/sidecar/README.md` (gs:// credentials: Workload Identity, inline JSON via `extraSecret`, key file via `extraVolumes`), `.env.sample`.
- `envoy-stack/README.md`, `sidecar-binary.md`, `sidecar/config.yaml`: what the model receives on an HTTP lane, rendered.
- `envoy-stack/grpc`, `envoy-stack/spanner`, `gcloud-stack`: the bucket shape noted beside `descriptors:`; the spanner demo gains a `spanner:` block mapping `demodb-pg` and a DIALECTS beat; `gcloud-stack` carries the block commented.
- `envoy-stack/spanner/demo-spanner.sh`: the pre-existing DELETE beat now sends `WHERE true`; the current emulator image refuses a DELETE with no WHERE.

### Test Configuration:
- **Browser(s)**: n/a
- **OS**: macOS 15.5, Go 1.26.5, Docker 29.4 (Cloud Spanner emulator, Envoy, OPA)

### Tests performed:
- [x] Unit tests pass
- [x] Integration tests pass

```
cd sidecar && GOPRIVATE=github.com/hoophq/libhoop CGO_ENABLED=0 go test ./analyzer/ ./daemon/ ./gate/ ./inspect/ ./descriptors/...
cd sidecar/descriptors/gcs && CGO_ENABLED=0 go test ./...
make test-sidecar
```

gRPC descriptors, unit:

- `TestGRPCLaneLoadsDescriptorsFromARegisteredFetcher`: a URL entry and a file entry merge into one schema, `-validate` reports both methods, a missing object fails naming the URL.
- `TestGRPCConfigRefusesADescriptorSchemeTheBinaryDoesNotLink`: `Config.Validate` refuses an unlinked scheme and names what is linked.
- `TestAGRPCDescriptorsEditKeepsTheRestartPath`: a changed list is restart-bound.
- `descriptors/gcs`: request path escaping, `generation`, malformed and unknown query parameters refused, bearer header, 404/403/401/5xx messages, size cap, token reuse while valid, mint bounded by the fetch deadline, inline JSON outranks ADC and a malformed value errors without echoing the key.
- libhoop: `TestLoadSchemaSourcesMergesAndNamesSourcesInConflicts`, `TestNewServerWithSchemaServesItAndRefusesASecondSource`.

gRPC descriptors, by hand with `hoop-inspect -validate`:

| Config | Result |
|---|---|
| `descriptors: s3://acme-schemas/billing.pb` | refused at config validation: `uses scheme "s3", which this binary does not link (linked: gs)` |
| `gs://` on a nonexistent bucket, laptop ADC | token minted, `404 Not Found` naming the URL |
| `GOOGLE_APPLICATION_CREDENTIALS=/nonexistent.json` | `no application default credentials found (...)` |
| `GOOGLE_APPLICATION_CREDENTIALS_JSON='{"type":"service_account"'` (truncated) | `GOOGLE_APPLICATION_CREDENTIALS_JSON is set but is not a service account key (want the key file's JSON, not a path; ...)` |
| valid SA JSON inline with a bogus private key, ADC path pointing at a missing file | fails at the mint (`private key should be a PEM...`), so the inline key was used, not ADC |
| `gs://cloud-samples-data/speech/brooklyn_bridge.raw` (public, not a set) | bytes fetched, libhoop refuses: `is not a FileDescriptorSet` |

Spanner dialect, unit:

- `TestSpannerSQLStatements`: database extracted from `session`/`database`, `database_dialect` from `CreateDatabase` in both JSON spellings, absent dialect is GoogleSQL.
- `TestSpannerDialectResolution`: absent block → googlesql; mapped database wins over lane default; `CreateDatabase` declaration wins over the map; `per_database` refuses an unlisted database and a request naming none. The observable consequence: `DELETE FROM "Songs" WHERE id = $1` is `delete` on `songs` under postgresql and names no table under googlesql.
- `TestSpannerConfigValidation`: dialect names, key shape, value set, block on a grpc lane.

Spanner dialect, integration (`deploy/docker-compose/envoy-stack`, `./run.sh --rebuild`, spanner overlay, `./spanner/demo-spanner.sh`, exit 0). The DIALECTS beat creates `demodb-pg` with `databaseDialect: POSTGRESQL` and sends `SELECT id, title FROM "songs" WHERE title = '<mark>'` to both databases; `/api/events` records:

```
projects/demo/instances/demo-instance/databases/demodb-pg postgresql songs
projects/demo/instances/demo-instance/databases/demodb    googlesql -
```

The emulator confirms the semantics on the GoogleSQL side (`Syntax error: Unexpected string literal "songs"`).

Analyzer:

- `TestHTTPWithBodyIsClassified` asserts the request line, `Resource:` line and body reach the provider.
- `TestHTTPCacheKeyFoldsIdsNotQueryValues` pins the cache: ids fold, query values do not.

## ✅ Checklist

### Decisions

- GCS: process-wide credential (inline JSON or ADC), no anonymous read, no per-lane credential knob. A public bucket reads with any valid token; a lane decoding payloads against a schema nobody authenticated for is not built.
- GCS: the fetch runs during `-validate`, on purpose: a wrong object name or a missing IAM binding must fail there, not at the first RPC. Fetch timeout is 2 minutes per entry; the artifact this exists for is the one too large for a ConfigMap.
- GCS: descriptor sets are restart-bound. The schema is bound into the endpoint when it is built, so a reload does not refetch; `?generation=N` makes a restart deterministic.
- Spanner: no dynamic detection. The data plane carries no dialect flag, `DatabaseAdmin.GetDatabase` would put an authenticated outbound call in the data path, and sniffing `$1` vs `@p` is a heuristic a client controls in the one component where a misread is an outage. `CreateDatabase.database_dialect` is the one request-level fact and is honored.
- Spanner: the PostgreSQL side gets the `lexer/conformance` oracle for free; GoogleSQL still has none.

### Not covered by an automated test

- A fetch against real GCS. The `gcs` tests run against an `httptest` stand-in for the JSON API and assert the request shape; the real read was performed by hand (table above).
- The spanner dialect switch end to end runs only in the compose demo (needs Docker), not in `make test-sidecar`. The daemon unit tests use a minimal test descriptor set whose `ExecuteSqlRequest` has no `session` field, so per-database resolution is pinned at the parser and resolver level.

### Worth a reviewer's attention

`Validate` wraps the lane name twice (`billing: billing: grpc descriptors: ...`). Pre-existing, from `buildGRPCServer` and `Validate` both prefixing; left alone.


EmanuelJr (approved):

sandromello:
✅ Build Completed with Success, Version=1826.0.0-gb0f973b

sandromello:
✅ Build Completed with Success, Version=1826.0.0-gc388baa

sandromello:
✅ Build Completed with Success, Version=1826.0.0-gaa80509

sandromello:
✅ Build Completed with Success, Version=1826.0.0-g49155a7

sandromello:
✅ Build Completed with Success, Version=1826.0.0-gb9338b4

sandromello:
✅ Build Completed with Success, Version=1826.0.0-g1be0b84

sandromello (approved):

PR #1833, opened 2026-09-18
EVL-297: the control plane review page

luanlorenzo:
## 📝 Description

An approver settled a sidecar review from Slack only: `/reviews` rendered a placeholder, and the Slack button pointed at the home page because there was no page to point at.

`/reviews` now lists the statements sidecars are holding and settles them. A row opens a modal with the statement, the approval rule, who decided and the verdict.

## 📣 User-facing impact

Control plane admins and approvers can now read and settle sidecar reviews in the app: the held statement, who approved or rejected it and when, with Approve and Reject from the same screen.

## 🔗 Related Issue

EVL-297

## 🚀 Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## 📋 Changes Made

- `pages/Reviews/`: list with an All / Waiting / Settled filter, detail modal, reject-with-reason modal, page store.
- `/reviews/:sessionId` opens the list with that review already open, so a link resolves to one review.
- `gateway/api/sidecar/reviews.go`: the Slack `WebappURL` points at `/reviews/<sid>` instead of the home page. Its test asserted the old behaviour and was updated.
- `services/reviews.js` gains `get` and `update` and stops unwrapping the axios response (`Dashboard/store.js` follows); `services/sessions.js` gains `get`.
- `formatRelativeTime` moves from `pages/Sidecars/status.js` to `utils/datetime.js`, which also gains `formatFullDate`.

Three behaviours come from the API rather than the design, and the reviewer should know they are deliberate:

- **Approve is hidden once a review is approved, Reject is not.** A retry only forwards while the status is `APPROVED`, so rejecting is the window to undo an approval before anyone consumes it. Approving twice does nothing.
- **`Approved` and `Released` are different labels.** `APPROVED` is released and waiting for the requester to run the statement again; `EXECUTED` is the retry that consumed it. Approving runs nothing.
- **Approve is disabled when the user's role is not among the review's groups.** `doIndividualReview` answers 200 and changes nothing in that case, so the button is what has to refuse it.

Known gap, out of scope: `GET /reviews` takes no filter, no pagination and no sort — it reads every review of the organization, with a subquery per row for the groups, on each load. The page filters client-side.

## 🧪 Testing

Validated end to end against a control plane with a real sidecar: `hoop start sidecar` against `hoop start control-plane`, a postgres lane with `high: require_review` and `approval_rule`, and psql as the client.

1. `DELETE FROM users;` through the lane → classified `high`, held, connection dropped with `statement held for human approval (review …)`.
[...]

PR #1837, opened 2026-09-21
fix(webapp): sign out lands on the login screen, not the role page

luanlorenzo:
## 📝 Description

Signing out of the control plane cleared the token but left the user inside the app shell, on the "Administrators and approvers only" dead end, with no way back to a signed-in state.

`ProtectedRoute` checks the token only inside a mount effect guarded by `initialized.current`. React Router reuses the instance across routes — the file's own comment says so — so a sign-out with a page open never reaches that check. `logout()` clears the user store, the gate re-renders without re-running the effect, and falls through to `if (adminOnly && !isAdmin) return <Navigate to="/" replace />`. That redirect races the menu's `navigate('/login')` and wins. `/` then renders the role landing, which sees `role: STANDARD` and shows the dead end — inside the shell, because the gate did not stop that route either.

The fix adds a render-time `isAuthenticated` guard, after `initializing` so the mount path is untouched and still saves the redirect URL for a deep link. Signing out saves none: it is not a place to come back to.

## 📣 User-facing impact

Signing out now returns you to the login screen instead of leaving you on an empty page that looks signed in.

## 🔗 Related Issue

None — found while testing #1833.

## 🚀 Type of Change

## 📋 Changes Made

- `components/ProtectedRoute.jsx`: redirect to `/login` at render time when the token is gone.

Not control-plane-specific. `ProtectedRoute` is the gate both products share, and the gateway does not revalidate on sign-out either; there `/` is the ClojureScript bundle, which handles its own session and sends the user to the login screen, so the fall-through never showed. The control plane surfaced it because `/` is the landing by role.

### Test Configuration:
- **Browser(s)**: Chrome
- **OS**: macOS

### Tests performed:

Steps, against `hoop start control-plane`:

1. Sign in as an admin and open `/sidecars`.
2. Sign out from the header menu.
3. Expected: the login screen, with no app shell. Before this change: `/` with the sidebar, the header and the "Administrators and approvers only" panel, while `localStorage` held no `jwt-token`.

`npm run lint` and `npm run build` clean for the changed file.

## ✅ Checklist

## 📄 Additional Notes

One related path I did not touch: session expiry. The 401 interceptor in `services/api.js` takes its own route out, and I did not reproduce it. Worth a look separately rather than folding it into this fix.


sandromello:
✅ Build Completed with Success, Version=1837.0.0-gdb3e494

sandromello:
✅ Build Completed with Success, Version=1837.0.0-g03d53bd

sandromello:
✅ Build Completed with Success, Version=1837.0.0-ga08316e

